PR #87, opened 2017-05-06
Duplicate tab names

### LINKED ISSUE #18: Same titles in tabs
Opening two or more folders with same folder names will result in two tabs named the same, and there is no way to determine which one is which, it can be done only through actually clicking on the tab. There should be some kind of mechanism like ie.: sublime text does for same file names.

Launchpad Details: [#LP1621637](https://bugs.launchpad.net/bugs/1621637) Adam Bieńkowski - 2016-09-08 21:23:31 +0000

jeremypw:
Fixes #18 

This PR makes tabs with the same basename distinguishable by:
* Prepending sufficient parent directories or, where the path is the same
* Appending a copy number in brackets.

It only considers individual windows; it was considered too complex for this branch to make it work across multiple Terminal windows. 

The window title shows the modified active tab label.


cassidyjames:
I still have to pull this and try it out, but from reading the comments and glancing through the code, I do have a thought:

I'm unsure that disambiguation with `(1)`, `(2)`, etc. when the tabs are the exact same path is needed or wanted. If I duplicate a tab, I already have a spatial disambiguation of the tabs: their position on the tab bar. In this case, the location is actually identical, so I'd expect the labels to be identical. As soon as one of those tabs runs a command, we replace the title with with command run, so that helps track the state of the tab as well.

The point of disambiguation is to prevent someone from thinking two _different_ directories are the same, not to make every tab globally unique.

jeremypw:
OK, I can easily pull out that feature.

cassidyjames:
Other than the `(1)` `(2)` thing, I think the behavior is an improvement. There are potentially some weird-ish edge cases, but I don't think they're blockers at all.

src/PantheonTerminalWindow.vala
```diff
@@ -788,6 +802,97 @@ namespace PantheonTerminal {
...
+            });
+        }
+
+        /** Compare every tab label with every other and resolve ambiguities **/
+        private bool check_for_tabs_with_same_name () {
+            /* Take list copies so foreach clauses can be nested safely*/
+            var terms = terminals.copy ();
+            var terms2 = terminals.copy ();
+
+            foreach (TerminalWidget t in terms) {
```
danrabbit: Can we not use variable names like `t`? I feel like this makes the code more difficult to read, especially if it ever gets more complex. `terminal_widget` would be more clear

src/PantheonTerminalWindow.vala
```diff
@@ -788,6 +802,97 @@ namespace PantheonTerminal {
...
+                }
+
+                /* Reset tab name to basename so long name only used when required */
+                t.tab_name = name;
+
+                foreach (TerminalWidget t2 in terms2) {
+                    string t2_path = t2.get_shell_location ();
+                    string t2_name = t2.window_title;
+
+                    if (t2 != t && t2_name == name) {
```
danrabbit: I don't feel like these are good variable names and it makes the code very unclear.

src/PantheonTerminalWindow.vala
```diff
@@ -788,6 +802,97 @@ namespace PantheonTerminal {
...
+                }
+            }
+
+            return true;
+        }
+
+        private string disambiguate_name (string name, string path, string conflict_path) {
+            string prefix = "";
+            string prefix2 = "";
+            string pth = path;
```
danrabbit: Why are you creating a new string `pth` instead of just using the existing string `path`?

src/PantheonTerminalWindow.vala
```diff
@@ -788,6 +802,97 @@ namespace PantheonTerminal {
...
+            string pth2 = conflict_path;
```
danrabbit: Same question here. It seems like `conflict_path` is a better variable name than `pth2`

src/PantheonTerminalWindow.vala
```diff
@@ -788,6 +802,97 @@ namespace PantheonTerminal {
...
+            return (prefix + name).replace ("//", "/");
+        }
+
+        /*** Simplified version of PF.FileUtils function, with fewer checks ***/
+        private string get_parent_path_from_path (string path) {
+            if (path.length < 2) {
+                return Path.DIR_SEPARATOR_S;
+            }
+
+            StringBuilder sb = new StringBuilder (path);
```
danrabbit: Let's use `string_builder` instead of `sb`

src/TerminalWidget.vala
```diff
@@ -53,6 +52,18 @@ namespace PantheonTerminal {
         public Granite.Widgets.Tab tab;
         public string? uri;
 
+        private string _tab_name = "";
```
danrabbit: Seems like it would be more in keeping with Gtk to call this `tab_label`

src/PantheonTerminalWindow.vala
jeremypw: Personally, I find it easier to read (& write) complex code with short variable names otherwise I cannot see the wood for the trees.  However, I can easily replace the short variables with longer names in keeping with the preferred style.

jeremypw: I meant the above to refer to short functions and local variables only.

src/PantheonTerminalWindow.vala
jeremypw: To quote ric*tz on Slack: "using a function parameter as local-variable is not a good idea and leads to such problems parameters are unowned".

src/TerminalWidget.vala
jeremypw: When duplicates were being distinguished by numbers, the tab label was different from the tab name; but now I agree there is no need to distinguish.

jeremypw:
I have changed the variable names as advised and made a slight improvement to the disambiguation function (removing an unnecessary parameter).  I noticed a bug while testing but it turned out to be present in trunk.  I'll raise an issue (if not already there)

Issue #94, opened 2017-05-18
Cannot exit from restored tab.


cgroner:
Once a terminal tab has been closed and restored, typing `exit` in the terminal does not close the tab. Additionally, closing the tab using the close 'x' in the UI causes a segmentation fault.

Steps to reproduce:

1. Open new tab.
2. Close newly opened tab using the close 'x' in the tab UI.
3. Reopen the tab by right clicking tab bar and selecting the "Undo Close Tab" entry.
4. In the reopened terminal, type `exit`.

I would expect typing 'exit' in step 4 would close the tab and it remains open. Clicking the close 'x' button in the tab UI then triggers a segfault.

pantheon-terminal version: 0.4.2

cgroner:
It appears that the [signal handler](https://github.com/elementary/terminal/blob/12d615e456037e77baab1051fe9450a439b93efc/src/PantheonTerminalWindow.vala#L572) added to the `TerminalWidget` to close the terminal tab is somehow getting disconnected after the tab is initially closed and subsequently not called again.

This also seems to be happening to the [`window_title_changed` signal handler](https://github.com/elementary/terminal/blob/12d615e456037e77baab1051fe9450a439b93efc/src/TerminalWidget.vala#L146), which is likely the cause of bug #88.

LightSunrider:
Fixed by #126

PR #95, opened 2017-05-19
Fix #88 & #94 - Disconnect signals when tab closed; reconnect on restoring

### LINKED ISSUE #88: Tab label does not update after restoring tab.
If a tab is closed and then restored, changing the working directory does not update the tab label.

jeremypw:
For reasons not immediately clear, all the terminal widget signals become disconnected from their handlers when the terminal is made restorable.  This branch explicitly handles disconnection and reconnection of signals when tab closed/restored.

This fixes two related issues.  #88 and #94


jeremypw:
Conflicted merge resolved but needs retesting to check for regressions as it was not straightforward.

jeremypw:
High priority assigned as it is associated with a crash under some circumstances.

src/TerminalWidget.vala
```diff
@@ -174,6 +137,67 @@ namespace PantheonTerminal {
...
+
+        public void disconnect_signals () {
+            settings.changed.disconnect (restore_settings);
+            child_exited.disconnect (on_child_exited);
+            button_press_event.disconnect (on_button_press);
+            button_release_event.disconnect (on_button_release);
+            selection_changed.disconnect (on_selection_changed);
+        }
+
+        private bool on_button_press (Gdk.EventButton event) {
```
donadigo: Now, I don't know if this would work, but I think, this could be done differently by just overriding the default `button_press_event` virtual signal here.

src/TerminalWidget.vala
```diff
@@ -174,6 +137,67 @@ namespace PantheonTerminal {
...
+                menu.select_first (false);
+                menu.popup (null, null, null, event.button, event.time);
+
+                return true;
+            }
+
+            return false;
+        }
+
+        private bool on_button_release (Gdk.EventButton event) {
```
donadigo: The same could be done here.

donadigo:
Those issues were resolved by a much simpler fix in https://github.com/elementary/terminal/pull/126. Closing.

Issue #86, opened 2017-05-06
A "Open in Files" option in context menu


donadigo:
It would be really convinient to have an option in the context menu to open current terminal directory directly in the preferred File Manager (rather than executing "xdg-open ." etc).

PR #101, opened 2017-05-31
Added open in files option to context menu

rickybas:
In reference to issue #86 

### Changes
* Open in Files option added to context menu
* Shortcut for Open in Files is Control+Shift+e


rickybas:
Hi @donadigo , I have added the error catching and improvements you suggested.

PR #105, opened 2017-06-15
Fix #98 - Pass F keys to TerminalWidget first.

### LINKED ISSUE #98: F-keys not work in mc
F-keys not work,  I have to mouse-click on "10 Quit" to quit from Midnight Commander

jeremypw:
See issue #98.

This fix was tested with htop but should also work with other commandline programs.


jeremypw:
Added medium priority label as it is associated with broken functionality under some circumstances.

donadigo:
I believe we shouldn't do things like this. This could lead to some unwanted behaviour. In theory returning `false` should pass the event further to the terminal widget itself.

jeremypw:
Seeking a more elegant solution. 

Just putting 
> return current_terminal.key_event (e);

at the end of the keyhandler blocks effect of F11 and may cause other regressions.

jeremypw:
Withdrawing since better solution submitted by donadigo.

jeremypw:
PR closed in favour of fix-F10-key

PR #108, opened 2017-06-18
Add screenshot to appdata

danrabbit:


jeremypw:
Neither appcenter nor synaptic shows a screenshot after installing the modified appdata and restarting.  Are they supposed to?

donadigo:
@jeremypw I don't think the AppStream data will be readily available in AppCenter when you install a new AppStream file. This data has to be recompiled and delivered by a specific package in order to be visible in package managers. Please correct me if I'm wrong.

Issue #96, opened 2017-05-19
On fresh install, the Terminal window size is very small


uncertainquark:
![screenshot from 2017-05-19 18-55-48](https://cloud.githubusercontent.com/assets/22585732/26249586/bb074dac-3c96-11e7-99d0-feaf45bfdfc2.png)

A more comfy quarter of a screen would be nice as this one is too small to not fit even small commands in one line.

I am sorry if I am nitpicking but its a small thing that can matter.

cassidyjames:
@elementary/ux I agree and have noticed this; I always size it up. What should the default be (keeping in mind in Loki, Mutter includes the shadow as part of the window size request)?

danrabbit:
I think last time we looked at this, @codygarver had a reason for keeping the default Terminal size very small

vjr:
Was it due to the possibility of low-res screens or regular screens with bad gfx configs (causing low-res again), in which scenario being able to open a terminal and work on fixup commands is ideal?

Could the size not just be computed to say half the length and width and window centered, so it's appropriate for whatever screen res shows up?

vjr:
It's currently hardcoded here: https://github.com/elementary/terminal/blob/master/org.pantheon.terminal.gschema.xml#L22

If we can set those values to -1 (like the initial x/y pos) then calculate it at startup similar to https://github.com/elementary/terminal/blob/master/src/PantheonTerminalWindow.vala#L332 then that would be cool.

Let me see if I can propose a PR here :-)

vjr:
Correct PR link is https://github.com/elementary/terminal/pull/119

PR #119, opened 2017-07-01
Pick a larger window size for first run.

vjr:
To address issue https://github.com/elementary/terminal/issues/96


src/PantheonTerminalWindow.vala
```diff
@@ -325,6 +325,14 @@ namespace PantheonTerminal {
             default_width = PantheonTerminal.saved_state.window_width;
             default_height = PantheonTerminal.saved_state.window_height;
 
+            if (default_width == -1) {
+                default_width = Gdk.Screen.width () * 2 / 3;
```
cassidyjames: Is this HiDPI-safe? I.e. if I have a 3200x1800 display with a scaling factor of 2, does `Gdk.Screen.width()` return 3200 or 1600?

cassidyjames: I'm also curious how this behaves with multiple displays.

vjr: Works on my hidpi display. For multiple screen maybe it needs work but existing move x y code might already have this issue anyway.

vjr:
@cassidyjames good point about the multi screen thing. this commit https://github.com/elementary/terminal/pull/119/commits/a0c6f57f30706f2e153972436b46bbd91286e164 should address it by changing the calls of `Gdk.Screen.width ()` (which gets the default screen dimensions) to call `get_screen ().get_height ()` since this should return the screen on which the current window resides. How does that look? Note the existing code here (for centering the window) might have multi screen issue for the same code usage: https://github.com/elementary/terminal/blob/master/src/PantheonTerminalWindow.vala#L335

cassidyjames:
@vjr that sounds better! I'll have to take a look on a HiDPI display and a few different resolutions to make sure this works well.

vjr:
OK this branch is now "ready for review" , there's a known centering problem for multi display filed as a separate issue #121 which I'll try to work on shortly, if I can set up multi display for myself.

vjr:
I was able to get a multi monitor setup by connecting my TV and the latest commit https://github.com/elementary/terminal/pull/119/commits/04dc0a7a1d149f429776953b5ee77d65e932553d should also address issue #121 so please test and let me know. It is working for me in my basic tests (opening first run, resizing and moving, then reopening) for hidpi, lodpi, single and dual monitor.

vjr:
Thank you @cassidyjames . FYI I also tested on 3 computers with HiDPI, multi monitor and 1080p and 1280x800 displays and all seem to look good to me :-) Looking forward to the code review.

Issue #45, opened 2017-04-26
No warning sign when pasting "sudo" using the middle mouse button


elementaryBot:
The warning sign shows up when pasting using Ctrl+V, but not when using the middle mouse button.

Launchpad Details: [#LP1444145](https://bugs.launchpad.net/bugs/1444145) Tomáš Bubeníček - 2015-04-14 20:58:16 +0000

PR #125, opened 2017-07-18
Unsafe paste. (#45)

LightSunrider:


danrabbit:
Can you explain what the intended change here is? Is there an issue that this is intended to resolve?

LightSunrider:
Unfortunately my knowledge is not very broad and I can make a mistake (if I have not already done that :) )

In the terminal widget (vte, gtk or the other, I'm not sure) have their own code to insert text from the clipboard by pressing the middle mouse button. Because of this, you can insert non-secure commands without warning. This can be fixed by own 'middle mouse button pressing' signal handler for paste.

I tested better and realized that my solution does not work on the restored tabs. I think that the beautiful solution will require a little refactoring (what I'm afraid and decided not to do so as not to spoil the existing code).

LightSunrider:
I have carefully read code of the widgets. Gtk.Container automatically destroys the child widget. To avoid this, the child widget should be removed.

My solution should fix #45, #88, #94, and links in restored tabs (they were not clickable)

src/PantheonTerminalWindow.vala
```diff
@@ -658,6 +670,8 @@ namespace PantheonTerminal {
             restorable_terminals.insert (term.terminal_id, term);
             tab.restore_data = term.terminal_id;
 
+            ((Gtk.Container)tab.page).remove(term);
```
donadigo: Wouldn't `term.destroy ()` work here?

src/PantheonTerminalWindow.vala
```diff
@@ -613,6 +613,18 @@ namespace PantheonTerminal {
...
 
+            t.button_press_event.connect ((e) => {
+                if (e.button == Gdk.BUTTON_MIDDLE) {
+                    if (this.search_toolbar.search_entry.has_focus) {
+                        return false;
+                    } else if (clipboard.wait_is_text_available ()) {
+                        action_paste ();
+                        return true;
+                    }
+                }
```
donadigo: Add a whitespace here.

src/PantheonTerminalWindow.vala
LightSunrider: Sorry, I don't know what you meant. If after this line add `term.destroy ()`, then all term signals will be disabled

donadigo: @LightSunrider ah yes, you're right here.

LightSunrider:
I should revert 9a1ceb83ff1bd3cba2e28cfefe659914b66bb0f2 in this branch and create new pull request with this commit in new branch?

donadigo:
@LightSunrider yeah it would be ideal.

LightSunrider:
Thanks, I hope I did it right

LightSunrider:
Can you review my changes? They are incorrect or incomplete?

donadigo:
Ups, sorry, I forgot I reviewed this before. I will review it today / tomorrow. I removed the Incomplete label.

src/Widgets/TerminalWidget.vala
```diff
@@ -130,6 +130,9 @@ namespace PantheonTerminal {
 
                     return true;
                 }
+                else if (event.button == Gdk.BUTTON_MIDDLE) {
```
donadigo: Make this `else if` on the same line as the above `}`:

```
...
} else if (...) {

}
```

src/Widgets/TerminalWidget.vala
```diff
@@ -130,6 +130,9 @@ namespace PantheonTerminal {
...
+                    return window.handle_paste_event();
```
donadigo: Add a space between the `handle_paste_event` and `()`.

src/PantheonTerminalWindow.vala
```diff
@@ -319,6 +319,17 @@ namespace PantheonTerminal {
             });
         }
 
+        public bool handle_paste_event () {
```
donadigo: Could this be called also here?
https://github.com/LightSunrider/terminal/blob/97fbeeb5949e907b988fc0f20534a1fd41d62180/src/PantheonTerminalWindow.vala#L309

This is the same code as in this method.

LightSunrider:
Thank you for your comments. I fixed it.

src/PantheonTerminalWindow.vala
```diff
@@ -306,19 +306,25 @@ namespace PantheonTerminal {
...
+                        handle_paste_event ();
                     }
                 }
 
                 return false;
             });
         }
 
+        public bool handle_paste_event () {
+            if (this.search_toolbar.search_entry.has_focus) {
```
jeremypw: Unnecessary use of ****this****.  There is no ambiguity and rest of this class does not use this form for `searchbar`.

LightSunrider:
Fixed. All right now?

PR #129, opened 2017-08-07
Rewrite Unsafe Paste Dialog as Gtk.Dialog

danrabbit:
Rewrite as a HIG-compliant Gtk.Dialog

* GObject-style construction
* Update license header
* Don't use namespace

Before:

screenshot from 2017-08-07 10 13 58

After:

screenshot from 2017-08-07 10 08 08

### DIFF
--- a/src/UnsafePasteDialog.vala
+++ b/src/UnsafePasteDialog.vala
@@ -1,62 +1,79 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/***
-    BEGIN LICENSE
-
-    Copyright (C) 2011-2015 Pantheon Terminal Developers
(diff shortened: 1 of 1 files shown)


donadigo:
@artemanufrij it looks like you either didn't checkout or haven't restarted the terminal after compiling and installing.

artemanufrij:
@donadigo you are right. _started terminal from terminal_ :confused: 
but, checkbox looks still like it has a margin
unbenannt

danrabbit:
This is going to happen on Gtk 3.18 and nothing you can really do about it because the box-shadow is considered part of the allocation. In 3.22, shadows can be outside the allocation so that's why it's aligned on Juno but not Loki

PR #136, opened 2017-08-11
rewrite SearchToolbar.vala

danrabbit:
* GObject-style construction
* Update copyright header
* Use lambdas for methods that don't get re-used
* Reduce variable scope
* Don't create variables when unnecessary 
* Don't use `this`
* Make `clear ()` a public method instead of a signal

### DIFF
--- a/src/Widgets/SearchToolbar.vala
+++ b/src/Widgets/SearchToolbar.vala
@@ -1,101 +1,91 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
-/***
-    BEGIN LICENSE
-
-    Copyright (C) 2011-2015 Pantheon Terminal Developers
-    This program is free software: you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


src/Widgets/SearchToolbar.vala
```diff
@@ -1,101 +1,86 @@
...
-            } catch (RegexError er) {
-                warning ("There was an error to compile the regex: %s", er.message);
-            }
+        public void clear () {
+            search_entry.text = "";
         }
 
         public void previous_search () {
-            this.window.current_terminal.search_find_previous ();
+            window.current_terminal.search_find_previous ();
```
matzipan: Is there a reason why the handler for search_changed, which is pretty 7 lines long was moved to an anonymous callback, but these ones which are 1 line long haven't been?

danrabbit: These methods are public, the others were private and only used once

matzipan: Missed that part.

PR #146, opened 2017-09-28
Use GLib.Action in place of deprecated Gtk.Action

danrabbit:


src/PantheonTerminalWindow.vala
```diff
@@ -49,39 +49,69 @@ namespace PantheonTerminal {
...
+            { ACTION_PREVIOUS_TAB, action_previous_tab },
+            { ACTION_QUIT, action_quit },
+            { ACTION_SEARCH, action_search },
+            { ACTION_SELECT_ALL, action_select_all },
+            { ACTION_ZOOM_IN_FONT, action_zoom_in_font },
+            { ACTION_ZOOM_IN_FONT, action_zoom_out_font }
+        };
+
+        static construct {
+            action_accelerators.set (ACTION_CLOSE_TAB, "<Control><Shift>w");
```
donadigo: You can use a simpler syntax for this:
`action_accelerators[ACTION_CLOSE_TAB] = "<Control><Shift>w";`

src/PantheonTerminalWindow.vala
```diff
@@ -49,39 +49,69 @@ namespace PantheonTerminal {
...
+        public const string ACTION_SHOW_IN_BROWSER = "action_open_in_files";
+        public const string ACTION_PASTE = "action_paste";
+        public const string ACTION_PREVIOUS_TAB = "action_previous_tab";
+        public const string ACTION_QUIT = "action_quit";
+        public const string ACTION_SEARCH = "action_search";
+        public const string ACTION_SELECT_ALL = "action_select_all";
+        public const string ACTION_ZOOM_IN_FONT = "action_zoom_in_font";
+        public const string ACTION_ZOOM_OUT_FONT = "action_zoom_out_font";
+
+        public static Gee.MultiMap<string, string> action_accelerators = new Gee.HashMultiMap<string, string> ();
```
donadigo: This could be `private`. Move construction of this MultiMap to the `static construct` block.

src/PantheonTerminalWindow.vala
```diff
@@ -49,39 +49,69 @@ namespace PantheonTerminal {
...
-                <menuitem name="Paste" action="Paste"/>
-                <menuitem name="Select All" action="Select All"/>
-                <menuitem name="Search" action="Search"/>
-                <menuitem name="Show in File Browser" action="Show in File Browser"/>
-            </popup>
-            </ui>
-        """;
+        public SimpleActionGroup actions { get; construct; }
+
+        public const string ACTION_PREFIX = "win.";
```
donadigo: Do we really need this constant here? We can just add the `win.` prefix to all those actions and do not care about adding this prefix everytime we use an action.

PR #152, opened 2017-10-04
Fix light bleed on maximized Terminal windows

danrabbit:
Fixes https://github.com/elementary/stylesheet/issues/210

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -48,6 +48,10 @@ namespace PantheonTerminal {
             .terminal-window.background {
                 background-color: transparent;
             }
+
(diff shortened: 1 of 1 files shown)


Tidrek:
I checked it out. So, it solves that light bleed problem. But at the cost of transparency. When maximized, terminal window now becomes totally opaque. Is it supposed to be so?

danrabbit:
It doesn't seem that I can use gradient or anything here. That's the tradeoff that must be made to fix the issue :man_shrugging:

Tidrek:
Well, I think everyone will agree that preserving transparency while fullscreen is more important :/
So, I guess that means that in order to fix the issue properly, more significant changes to stylesheets/layout have to be made?

PR #157, opened 2017-10-21
Port accelerators over to GLib.action

danrabbit:
Port only accelerators over to GLib.Action. Don't port anything in the context menu.

This is a portion of the work from #146 that doesn't appear to cause any regression


danrabbit:
I was having trouble keeping the menu from breaking in my other branch so I just left those ones alone for now. I’ll come back to them once I figure out how to keep it from breaking 😬

jeremypw:
In Files, the popup menu actions are handled separately by the widget (AbstractDirectoryView) not the window.  You are probably right to leave that to another branch as it needs some significant refactoring.

PR #160, opened 2017-10-26
Restore zoom level from saved state

### LINKED ISSUE #143: zoom-level should be saved/restored between sessions
Would be nice if the zoom-level of the characters would be saved after closing the terminal and restored in a new session or a new tab.

Thalexander:
This stores the last set zoom factor into SavedState and restores it upon opening a new terminal (tab or window.)

Fixes #143

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -196,9 +196,8 @@ namespace PantheonTerminal {
             GLib.Settings saved_state = new GLib.Settings ("io.elementary.terminal.saved-state");
             saved_state.bind ("zoom", this, "zoom_factor", GLib.SettingsBindFlags.DEFAULT);
 
-            Idle.add (() => {
+            realize.connect (() => {
                 zoom_factor = zoom_factor;
-                return false;
             });
         }
 
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -65,8 +65,6 @@ namespace PantheonTerminal {
         }
 
         public int default_size;
-        private double _zoom_factor = 1.0;
-        private SavedState saved_state = new SavedState ();
 
         const string SEND_PROCESS_FINISHED_BASH = "dbus-send --type=method_call --session --dest=io.elementary.terminal /io/elementary/terminal io.elementary.terminal.ProcessFinished string:$PANTHEON_TERMINAL_ID string:\"$(history 1 | cut -c 8-)\" >/dev/null 2>&1; ";
 
@@ -104,16 +102,7 @@ namespace PantheonTerminal {
             private set;
         }
 
-        public double zoom_factor {
-            get {
-                return _zoom_factor;
-            }
-
-            set {
-                _zoom_factor = value;
-                saved_state.zoom = value;
-            }
-        }
+        public double zoom_factor;
 
         public TerminalWidget (PantheonTerminalWindow parent_window) {
 
@@ -186,6 +175,9 @@ namespace PantheonTerminal {
             /* Make Links Clickable */
             this.drag_data_received.connect (drag_received);
             this.clickable (regex_strings);
+
+            GLib.Settings saved_state = new GLib.Settings ("io.elementary.terminal.saved-state");
+            saved_state.bind ("zoom", this, "zoom_factor", GLib.SettingsBindFlags.DEFAULT);
         }
 
         public void restore_settings () {
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -386,32 +386,23 @@ namespace PantheonTerminal {
         }
 
         public void increment_size () {
-            Pango.FontDescription current_font = this.get_font ();
-            if (default_size == 0) default_size = current_font.get_size ();
-            if (current_font.get_size () > 60000) return;
+            if (zoom_factor > 5.8) {
+                return;
+            }
 
             zoom_factor += 0.1;
-            current_font.set_size ((int) Math.floor (default_size * zoom_factor));
-            this.set_font (current_font);
         }
 
         public void decrement_size () {
-            Pango.FontDescription current_font = this.get_font ();
-            if (default_size == 0) default_size = current_font.get_size ();
-            if (current_font.get_size () < 2048) return;
+            if (zoom_factor < 0.2) {
+                return;
+            }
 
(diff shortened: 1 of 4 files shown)


jeremypw:
I think it is simpler to bind the zoom property of the settings to the zoom_factor property of the TerminalWidget.

Thalexander:
Would that not force all terminals to have the same zoom level?

danrabbit:
Do we think it would make sense for each tab to have its own zoom level? I think this makes sense as a global setting, personally. Go go gadget @elementary/ux

jeremypw:
If you use bidirectional binding then the zoom-factors would be the same in each terminal.  You could use a SET binding flag and load the initial zoom-factor to allow individial adjustment.  Then the saved factor would be whichever was changed last. This is what Files does at the moment.

To save individual zoom-factors is more complicated ...

jeremypw:
As the content of the terminals is only text then having differing font sizes in different terminals does not make much sense to me.

Thalexander:
Okay the zoom level is now shared between terminals.

src/Widgets/TerminalWidget.vala
```diff
@@ -103,6 +102,8 @@ namespace PantheonTerminal {
             private set;
         }
 
+        public double zoom_factor;
```
jeremypw: This needs to be a property and the setter clause needs to contain code that will change the size of the font when the factor changes (with proper checks).  The increment, decrement and set_default zoom functions need to be modified accordingly (they only need change the zoom_factor).

src/Widgets/TerminalWidget.vala
```diff
@@ -389,6 +393,17 @@ namespace PantheonTerminal {
...
+        public void load_saved_font_size () {
+            zoom_factor = saved_state.zoom;
+            Pango.FontDescription current_font = this.get_font ();
+            if (default_size == 0) {
+                default_size = current_font.get_size ();
+            }
+
+            current_font.set_size ((int) Math.floor (default_size * zoom_factor));
+            this.set_font (current_font);
+        }
```
jeremypw: This function is not needed.

jeremypw:
OK, ignore last review - I didn't realise you were still working.

Thalexander:
The Idle add to correct the font seems really dodgy, but it needs it to work as the font is not defined until after construction... Is there a better way to handle that?

src/Widgets/TerminalWidget.vala
```diff
@@ -103,6 +102,25 @@ namespace PantheonTerminal {
...
+                if (current_font != null) {
+                    if (default_size == 0) {
+                        default_size = current_font.get_size ();
+                    }
+
+                    current_font.set_size ((int) Math.floor (default_size * zoom_factor));
+                    this.set_font (current_font);
+                }
+            }
+        }
```
jeremypw: Some sanity checking of zoom_factor would be a good idea.

src/Widgets/TerminalWidget.vala
```diff
@@ -174,6 +192,14 @@ namespace PantheonTerminal {
...
             this.drag_data_received.connect (drag_received);
             this.clickable (regex_strings);
+
+            GLib.Settings saved_state = new GLib.Settings ("io.elementary.terminal.saved-state");
+            saved_state.bind ("zoom", this, "zoom_factor", GLib.SettingsBindFlags.DEFAULT);
+
+            Idle.add (() => {
+                zoom_factor = zoom_factor;
+                return false;
+            });
```
jeremypw: Maybe better to connect to the realize signal to set the font size at the correct time.

src/Widgets/TerminalWidget.vala
Thalexander: That sounds more like another issue, as that problem would have existed before these changes.

jeremypw: At the moment you are duplicating some code in the increment, decrement and restore_default zoom functions and these also contain some range checking.  If you simplify these functions to just change the zoom factor then the range checking can be consolidated in the zoom factor setter.

Thalexander: I have simplified the increment, decrement and default zoom functions, I think it makes sense to only test the zoom_factor for the range when increment or decrementing as the functions exist and there is no reason to test it otherwise. What do you think a sane range would be? Currently the limits are a maximum of 590% and a minimum of 10%; I think from my testing 40% and 250% seem better limits for what a user is likely to want although I am not sure about the behaviour on HiDPI screens.

jeremypw:
This is better, but does not protect against stupid numbers being entered directly into the setting using dconf-editor or the command-line.  It could be argued that if users want to be stupid, let them, provided the app does not crash.  However, I cannot really see the disadvantage of concentrating the range checking in one place.

Thalexander:
I've moved the range checking to the set function, I added a 0.01 offset to the range checking due to floating point numbers.

src/Widgets/TerminalWidget.vala
```diff
@@ -103,6 +102,32 @@ namespace PantheonTerminal {
             private set;
         }
 
+        public double _zoom_factor = 1.0;
```
jeremypw: Backing variables for properties should be private

src/Widgets/TerminalWidget.vala
```diff
@@ -103,6 +102,32 @@ namespace PantheonTerminal {
...
+        public double zoom_factor {
+            get {
+                return _zoom_factor;
+            }
```
jeremypw: Conventionally a blank line between get and set clauses (after closing brace).

src/Widgets/TerminalWidget.vala
```diff
@@ -103,6 +102,32 @@ namespace PantheonTerminal {
...
+            set {
+                _zoom_factor = value;
+
+                if (_zoom_factor < 0.19) {
+                    _zoom_factor = 0.2;
+                } else if (_zoom_factor > 5.01) {
+                    _zoom_factor = 5;
+                }
+
```
jeremypw: Because "zoom_factor" property is bound to a setting, it feels better to test value and assign _zoom_factor accordingly rather than assign _zoom_factor and then maybe immediately reassign it - i.e.

```
                if (value < 0.19) {
                    _zoom_factor = 0.2;
                } else if (value > 5.01) {
                    _zoom_factor = 5;
                } else {
                    _zoom_factor = value;
                }
```

(Although no adverse effects were noticed your way)

PR #165, opened 2017-11-10
Dark light settings

### LINKED ISSUE #147: Consider using themes
I think it would be nice if we could expose a way in the UI for the user to set some pre-made themes like we do in Code.

This requires some things to solve. How do we want to store theme information? Is there an available standard format that's already doing that? If not, do we want to create one from scrach?

danrabbit:
Adds a setting for Dark/Light/High Contrast style Fixes #147

* Reformatted the hex_palette array and use colors from default gsettings
* Added a setting for gtk_prefer_dark and restore it on app startup
* Set prefer_dark, foreground, and background colors from the popover

screenshot from 2017-11-10 15 53 20

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -248,6 +248,12 @@ namespace PantheonTerminal {
             color_button.append_text (_("Light"));
             color_button.append_text (_("Dark"));
 
+            if (settings.prefer_dark_style) {
+                color_button.selected = 1;
+            } else {
+                color_button.selected = 0;
+            }
+
             var style_popover_grid = new Gtk.Grid ();
             style_popover_grid.margin = 12;
             style_popover_grid.orientation = Gtk.Orientation.VERTICAL;
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -885,10 +885,6 @@ namespace PantheonTerminal {
                 new_tab (Environment.get_home_dir ());
         }
 
-        void action_about () {
-            app.show_about (this);
-        }
-
         void action_zoom_in_font () {
             current_terminal.increment_size ();
             set_zoom_default_label (current_terminal.zoom_factor);
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -53,6 +53,31 @@ namespace PantheonTerminal {
             .terminal-window.background.maximized {
                 background-color: #000;
             }
+
+            .color-button {
+                border-radius: 50%;
+                box-shadow:
+                    inset 0 1px 0 0 alpha (@inset_dark_color, 0.7),
+                    inset 0 0 0 1px alpha (@inset_dark_color, 0.3),
+                    0 1px 0 0 alpha (@bg_highlight_color, 0.3);
+            }
+
+            .color-button:focus {
+                border-color: @colorAccent;
+            }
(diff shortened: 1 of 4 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -55,27 +55,33 @@ namespace PantheonTerminal {
...
+        public const string ACTION_CLOSE_TAB = "action_close_tab";
+        public const string ACTION_FULLSCREEN = "action_fullscreen";
+        public const string ACTION_NEW_TAB = "action_new_tab";
+        public const string ACTION_NEW_WINDOW = "action_new_window";
+        public const string ACTION_NEXT_TAB = "action_next_tab";
+        public const string ACTION_PREVIOUS_TAB = "action_previous_tab";
+        public const string ACTION_ZOOM_IN_FONT = "action_zoom_in_font";
+        public const string ACTION_ZOOM_OUT_FONT = "action_zoom_out_font";
+
+        private static Gee.MultiMap<string, string> action_accelerators = new Gee.HashMultiMap<string, string> ();
```
donadigo: Can this be constructed in the `static construct`?

src/PantheonTerminalWindow.vala
```diff
@@ -305,6 +316,21 @@ namespace PantheonTerminal {
             zoom_default_button.clicked.connect (() => action_zoom_default_font ());
             zoom_out_button.clicked.connect (() => action_zoom_out_font ());
 
+            color_button.mode_changed.connect (() => {
```
donadigo: What should happen here:

The `mode_changed` signal should set `settings.prefer_dark_style`.

Then you connect to `settings.notify["prefer-dark-style"]` and change the background and foreground based on that value. (You make an `update_style` or something).

danrabbit: In the case that we add a third style (for example a high contrast black on white style), prefering the dark stylesheet does not necessarily imply what the background and foreground colors should be.

src/PantheonTerminalWindow.vala
```diff
@@ -305,6 +363,24 @@ namespace PantheonTerminal {
             zoom_default_button.clicked.connect (() => action_zoom_default_font ());
             zoom_out_button.clicked.connect (() => action_zoom_out_font ());
 
+            color_button_dark.clicked.connect (() => {
+                settings.prefer_dark_style = true;
+                settings.background = "rgba(37, 46, 50, 0.95)";
+                settings.foreground = "#94a3a5";
```
donadigo: Can we get those strings as constants somewhere?

src/PantheonTerminalWindow.vala
```diff
@@ -305,6 +363,24 @@ namespace PantheonTerminal {
...
+            });
+
+            color_button_light.clicked.connect (() => {
+                settings.prefer_dark_style = false;
+                settings.background = "rgba(253, 246, 227, 0.95)";
+                settings.foreground = "#586e75";
```
donadigo: These too.

src/PantheonTerminalWindow.vala
```diff
@@ -305,6 +363,24 @@ namespace PantheonTerminal {
...
+            });
+
+            color_button_white.clicked.connect (() => {
+                settings.prefer_dark_style = false;
+                settings.background = "#fff";
+                settings.foreground = "#333";
```
donadigo: And these also.

PR #177, opened 2017-11-22
Remove unused schema gettext domain

codygarver:


codygarver:
We don't translate settings and it was the incorrect domain anyway.

PR #180, opened 2017-12-06
Execute parameter

jeremypw:
Use same code to spawn process whether active shell or command.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -315,10 +315,6 @@ namespace PantheonTerminal {
                 spawn_in_idle (GLib.Environment.get_current_dir (), program_with_args);
             } catch (Error e) {
                 warning (e.message);
-                var notification = new Notification (_("Run program"));
-                notification.set_body (_("Unable to parse command %s \n %s").printf (program_string, e.message));
-                notification.set_icon (new ThemedIcon ("utilities-terminal"));
-                app.send_notification ("spawn-error", notification);
             }
         }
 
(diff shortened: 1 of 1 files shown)


donadigo:
Cannot compile on Juno:
```
/home/donadigo/Projects/terminal-execute-parameter/src/PantheonTerminalWindow.vala:597.40-597.61: error: Argument 1: Cannot convert from `void PantheonTerminal.PantheonTerminalWindow.update_context_menu_cb (Gtk.Clipboard, Gdk.Atom[])' to `void Gtk.ClipboardTargetsReceivedFunc (Gtk.Clipboard, Gdk.Atom[]?)'
            clipboard.request_targets (update_context_menu_cb);
```

I don't think this is really a problem with this branch.

jeremypw:
Failure to compile on Juno due to a line unchanged by this PR therefore should not block review/merge.

danrabbit:
Looks like this branch should be building on Juno again due to fixes in master :)

src/Widgets/TerminalWidget.vala
```diff
@@ -298,10 +298,27 @@ namespace PantheonTerminal {
...
+                string[]? program_with_args = null;
+                Shell.parse_argv (program_string, out program_with_args);
+
+                spawn_in_idle (GLib.Environment.get_current_dir (), program_with_args);
+            } catch (Error e) {
+                warning (e.message);
+            }
+        }
+
+        public void spawn_in_idle (string dir, string[] argv){
```
donadigo: Space before `{`.

jeremypw:
This is not connected to any particular issue and I did not notice any difference in behaviour when executing apps from the command with the -e flag.  As far as I recall, this was just to simplify the code by only having one function to launch processes.

jeremypw:
Not essential

PR #185, opened 2017-12-29
PantheonTerminalWindow.vala: make constructor properties public

danrabbit:
* make `app`, `restore_pos`, and `recreate_tabs` public properties
* Sort public and private variables
* Use `Object` instead of passing vars to `init`
* Combine `init` and `construct`

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -21,10 +21,12 @@ namespace PantheonTerminal {
 
     public class PantheonTerminalWindow : Gtk.Window {
         private Pango.FontDescription term_font;
+        private Granite.Widgets.DynamicNotebook notebook;
         private Gtk.Clipboard clipboard;
         private Gtk.Clipboard primary_selection;
         private PantheonTerminal.Widgets.SearchToolbar search_toolbar;
         private Gtk.Revealer search_revealer;
+        private Gtk.ToggleButton search_button;
         private Gtk.Button zoom_default_button;
 
         private HashTable<string, TerminalWidget> restorable_terminals;
@@ -42,15 +44,13 @@ namespace PantheonTerminal {
         public bool focus_restored_tabs { get; construct; default = true; }
         public bool recreate_tabs { get; construct; default = true; }
         public bool restore_pos { get; construct; default = true; }
+        public Gtk.UIManager ui { get; private set; }
         public PantheonTerminalApp app { get; construct; }
         public SimpleActionGroup actions { get; construct; }
+        public TerminalWidget current_terminal { get; private set; default = null; }
         
         public GLib.List <TerminalWidget> terminals = new GLib.List <TerminalWidget> ();
-        public Granite.Widgets.DynamicNotebook notebook;
         public Gtk.ActionGroup main_actions;
-        public Gtk.ToggleButton search_button;
-        public Gtk.UIManager ui;
-        public TerminalWidget current_terminal = null;
 
         public const string ACTION_PREFIX = "win.";
         public const string ACTION_CLOSE_TAB = "action_close_tab";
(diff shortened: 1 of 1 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
         private const string SOLARIZED_LIGHT_BG = "rgba(253, 246, 227, 0.95)";
         private const string SOLARIZED_LIGHT_FG = "#586e75";
 
-
-
+        public bool unsafe_ignored;
+        public bool focus_restored_tabs { get; construct; default = true; }
+        public bool recreate_tabs { get; construct; default = true; }
+        public bool restore_pos { get; construct; default = true; }
+        public PantheonTerminalApp app { get; construct; }
```
jeremypw: This duplicates Widget.application - is it advantageous to have a separate property as well as set_application (app) in construct {}?

src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
...
         public SimpleActionGroup actions { get; construct; }
+        
+        public GLib.List <TerminalWidget> terminals = new GLib.List <TerminalWidget> ();
+        public Granite.Widgets.DynamicNotebook notebook;
```
jeremypw: Can be made private.

src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
...
+        public Gtk.ActionGroup main_actions;
```
jeremypw: Can be made read-only with { get; private set; }

src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
...
+        public Gtk.ToggleButton search_button;
+        public Gtk.UIManager ui;
```
jeremypw: Can be made read-only with { get; private set; }

src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
...
+        public TerminalWidget current_terminal = null;
```
jeremypw: Can be made read-only with { get; private set; }

src/PantheonTerminalWindow.vala
```diff
@@ -53,9 +39,19 @@ namespace PantheonTerminal {
...
+        public bool focus_restored_tabs { get; construct; default = true; }
+        public bool recreate_tabs { get; construct; default = true; }
+        public bool restore_pos { get; construct; default = true; }
+        public PantheonTerminalApp app { get; construct; }
         public SimpleActionGroup actions { get; construct; }
+        
+        public GLib.List <TerminalWidget> terminals = new GLib.List <TerminalWidget> ();
+        public Granite.Widgets.DynamicNotebook notebook;
+        public Gtk.ActionGroup main_actions;
+        public Gtk.ToggleButton search_button;
```
jeremypw: Can be made private

src/PantheonTerminalWindow.vala
jeremypw: Can be made private

jeremypw: Can be made private

src/PantheonTerminalWindow.vala
danrabbit: I'd like to chase this down in a follow-up PR, but there seems to be some conflict between `Gtk.Application` and `PantheonTerminalApp`. Trying to use `Widget.Application` is giving me errors about setting shortcut accelerators and some other issues.

PR #186, opened 2017-12-30
Restore tabs when starting with a directory

### LINKED ISSUE #62: Open Terminal from Files results in lost tabs
Loki Beta-2

* Expected behavior:

Tabs opened in Terminal should be stored and re-opened after closing Terminal.

* What happens:

Tabs are lost if a new Terminal instance is launched from Files.

* Steps:

1- Open some tabs in Terminal (i.e.: Downloads/ and Documents/)
2- Close Terminal.
3- Open Files.
4- Right click and select 'Open in/Terminal' (from within any folder)
5- The pinned terminal opens, and all previously stored tabs are lost.

Launchpad Details: [#LP1612317](https://bugs.launchpad.net/bugs/1612317) Gabriel_P - 2016-08-11 15:47:30 +0000

Thalexander:
Restores tabs when opening Terminal with a directory, this tab however would not gain focus as tab restores are done at full idle. To ensure the tab correctly grabs focus the tab restore and subsequently create functions had to be modified so they had a parameter to define whether they would grab focus.

Fixes #62

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -43,6 +43,7 @@ namespace PantheonTerminal {
 
         public TerminalWidget current_terminal = null;
         private bool is_fullscreen = false;
+        public bool focus_restored_tabs { get; construct; default = true; }
         private string[] saved_tabs;
 
         private const string HIGH_CONTRAST_BG = "#fff";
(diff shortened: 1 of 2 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -108,7 +108,7 @@ namespace PantheonTerminal {
 
         public PantheonTerminalWindow.with_working_directory (PantheonTerminalApp app, string location,
                                                               bool should_recreate_tabs = true) {
-            init (app, should_recreate_tabs);
+            init (app, should_recreate_tabs, true, false);
```
danrabbit: Instead of doing this can we add a new public property? This conflicts with the changes in #185 to make this class more in GObject style

Thalexander: I have matched the style of your pull request, this does still conflict but it is a pretty minor conflict to resolve.

PR #189, opened 2018-01-01
Don't restore tabs in a new window

### LINKED ISSUE #43: "New Window" shouldn't open session
When I open a new terminal window from Plank (right click + New Window), the tabs opened in the old terminal shouldn't be opened. Instead I should get one new tab `cd`ed to `~`.

Launchpad Details: [#LP1416550](https://bugs.launchpad.net/bugs/1416550) SuperScript - 2015-01-30 21:30:19 +0000

Thalexander:
When opening a new window don't restore tabs because they would already be open.

Fixes #43 

Edit: The code repetition could be addressed after #185, but not currently without creating more conflict for that pr.


jeremypw:
Has conflicts.

jeremypw:
This behaves differently to Files - which does not restore tabs even if the new window is the first window to be opened (using Plank "New Window" option).  I am not sure which is the desired behaviour but there should be consistency.  Needs design team input on this point.

Incidentally Scratch/Core restores tabs in both cases but crashes if initially launched from Plank "New WIndow".

jeremypw:
Another issue is that if the New Window is the last one to be closed then the original saved opened tabs are lost.  Again I am not sure what the desired behaviour is.

jeremypw:
On reflection, I feel that the behaviour of this branch regarding Plank is correct.  The other issue can be left for another PR if change is desired.

PR #190, opened 2018-01-02
Restore tab order

### LINKED ISSUE #13: State only saves on clean close
The app's state is only saved when you explicitly/cleanly close it. If it terminates uncleanly (i.e. crashes, the computer dies, etc.), the state is not saved.

Possible solution: periodically save the state, perhaps after certain actions occur (i.e. after navigating to a folder, opening a tab, or toggling a UI state).

Launchpad Details: [#LP1454331](https://bugs.launchpad.net/bugs/1454331) Cassidy James Blaede - 2015-05-12 16:27:25 +0000

### LINKED ISSUE #14: Tabs in terminal change order after close/open
Steps:

1- Open several tabs in Terminal, located in different folders
2- Close Terminal
3- Open Terminal

Expected result: tabs should remain in the order I gave them before closing the app.

Result: some tabs are moved from their original (pre-closing of the app) positions.

See attached image.

Launchpad Details: [#LP1461595](https://bugs.launchpad.net/bugs/1461595) Gabriel_P - 2015-06-03 15:38:47 +0000

Thalexander:
Restores tab order by loading them in one Idle.add and saving tabs on tab changes, not just on close.

Fixes #14 
Fixes #13


src/PantheonTerminalWindow.vala
```diff
@@ -992,6 +990,8 @@ namespace PantheonTerminal {
             }
 
             name_check_timeout_id = Timeout.add (50, () => {
+                save_opened_terminals ();
```
jeremypw: I am not sure this is necessary here.  It is already being called after adding, closing and restoring a tab.  Only need to save if tab labels are changed (see below) due to disambiguation. Maybe at line 1033?

Thalexander: It will not in all cases update the saved tabs on a name change with that set up unfortunately.

jeremypw:
There may be a better solution, but if you insert a terminal message in `save_opened_terminals ()`, you will see that it is called twice every time you add or remove a terminal.  If you do not change `schedule_name_check ()`, you probably do not need to save opened terminals both in `new_tab ()` and in `on_new_tab_added ()` (indirectly) for example.

Can you give an example of a case where the state is not saved unless you make these calls everywhere you have?

jeremypw:
I would also point out that saving the tab order is not an inherent function of disambiguating the tab names and therefore should be be called there by default - only when the name is actually changed.

jeremypw:
OK, that seems reasonable - I'll approve now - thanks.

PR #193, opened 2018-01-10
Remove invasive CMake variables

danrabbit:
Kills some invasive build system stuff. This is gonna make Meson porting easier, but it's also just stuff we don't need.

Everything is always RDNN now, so we don't need the build system to manage which variables have which schemes re: exec, gettext, etc


danrabbit:
But what’s the goal of having the build system alter these files? It seems like making things complicated for no gain

codygarver:
The unnecessary complication is that this project is using 2 config files to accomplish what 1 could do, but I'd guess that's because this pre-dates the easy way being known or possible. That's not done in any of the other projects.

The goal is the same as any other variable in code, centralize repetitious info and re-use it efficiently. I think it also helps readability because just seeing "io.elementary.terminal" all over the place doesn't give the same context as `EXEC_NAME` and `GETTEXT_NAME`, because you know what that's talking about.

codygarver:
It's also making it a vapi instead of just a config.vala, which makes it overcomplicated.

PR #194, opened 2018-01-11
Port menu to glib.action

danrabbit:
I don't know why this doesn't work :(


jeremypw:
Agree with DMH - that is how the Files context menus work.

danrabbit:
Superseded by #253

PR #196, opened 2018-01-18
Restore tab focus

### LINKED ISSUE #12: Save active tab
When you have +1 tabs, if you close and open pantheon-terminal again, the active tab will always be the las from left to right. Pantheon should save the last active tab instead.

Launchpad Details: [#LP1439517](https://bugs.launchpad.net/bugs/1439517) Erasmo Marín - 2015-04-02 06:13:46 +0000

Thalexander:
Adds a new key to save to index of the focused tab so that when restoring tabs it will be focused to the correct tab (if the tabs are set to restore focus.)

Fixes #12

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -695,7 +695,7 @@ namespace PantheonTerminal {
                     }
                 }
 
-                if (focus_restored_tabs && focus == i) {
+                if (focus_restored_tabs && focus == tabs.length) {
                     var t = notebook.get_tab_by_index (notebook.n_tabs - 1);
                     notebook.current = t;
                     t.grab_focus ();
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -647,6 +647,8 @@ namespace PantheonTerminal {
             set_zoom_default_label (current_terminal.zoom_factor);
             new_tab.icon = null;
             new_tab.page.grab_focus ();
+
+            PantheonTerminal.saved_state.focused_tab = notebook.get_tab_position(new_tab);
         }
 
         private void open_tabs () {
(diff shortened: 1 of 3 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -647,6 +647,8 @@ namespace PantheonTerminal {
             set_zoom_default_label (current_terminal.zoom_factor);
             new_tab.icon = null;
             new_tab.page.grab_focus ();
+
+            PantheonTerminal.saved_state.focused_tab = notebook.get_tab_position(new_tab);
```
jeremypw: Missing space before bracket

src/PantheonTerminalWindow.vala
```diff
@@ -1046,6 +1063,7 @@ namespace PantheonTerminal {
             });
 
             PantheonTerminal.saved_state.tabs = opened_tabs;
+            PantheonTerminal.saved_state.focused_tab = notebook.get_tab_position(notebook.current);
```
jeremypw: Missing space before bracket

src/PantheonTerminalWindow.vala
```diff
@@ -676,14 +678,29 @@ namespace PantheonTerminal {
...
-                        new_tab (loc, null, focus_restored_tabs);
+                        new_tab (tabs[i], null, false);
                     }
                 }
+
+                if (focus_restored_tabs && focus == tabs.length) {
+                    var t = notebook.get_tab_by_index (notebook.n_tabs - 1);
+                    notebook.current = t;
+                    t.grab_focus ();
+                }
```
jeremypw: Can the tab be focussed here only? i.e. tabs all created unfocussed and then `var t = notebook.get_tab_by_index (focus.clamp (0, notebook.n_tabs - 1)` etc.

src/PantheonTerminalWindow.vala
```diff
@@ -676,14 +678,29 @@ namespace PantheonTerminal {
 
             PantheonTerminal.saved_state.tabs = {};
 
+            int focus = 0 <= PantheonTerminal.saved_state.focused_tab < tabs.length
+                            ? PantheonTerminal.saved_state.focused_tab : tabs.length - 1;
```
jeremypw: This is rather unclear and may not compile with older versions of valac. At least add a comment to explain the effect.

Consider using `focus = PantheonTerminal.saved_state.focused_tab.clamp (0, tabs.length - 1)`. This is clearer and means that a negative setting focusses the first tab.  As this situation is likely to be a rare corner case the exact behaviour is not that important.

src/PantheonTerminalWindow.vala
```diff
@@ -676,14 +678,24 @@ namespace PantheonTerminal {
 
             PantheonTerminal.saved_state.tabs = {};
 
+            int focus = PantheonTerminal.saved_state.focused_tab.clamp(0, tabs.length - 1);
             Idle.add_full (GLib.Priority.LOW, () => {
+                focus += notebook.n_tabs;
```
jeremypw: I am not sure about this.  Normally if `notebook.n_tabs` is greater than 0 then terminal is being launched with a specified working directory in which case focus_restored_tabs is false and `focus` is not used anyway.  Is there a case when this not so?  Not that it does any harm.

Thalexander: No there is no case when this happens, I have included this so if the behaviour changes it will still work.

jeremypw:
OK, fair enough.

PR #200, opened 2018-01-20
Allow not ensuring tab existence for new window with coord

### LINKED ISSUE #199: Extra tab created when new window created by dragging tab
If a tab is dragged out of an existing window, a new window is created (as expected).  However, the new window contains an extra tab (usually "Home") as well as the one dragged which is not expected.  Moreover, the extra tab is focused not the one dragged.

Thalexander:
When creating a new window with coord it always attempted to create a tab if none existed (from #189) as no tab existing breaks the window, this adds a parameter to not ensure this, for if a tab will be created later.

Fixes #199

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -68,8 +68,8 @@ namespace PantheonTerminal {
             }
         }
 
-        public PantheonTerminalWindow new_window_with_coords (int x, int y, bool should_recreate_tabs=true) {
-            var window = new PantheonTerminalWindow.with_coords (this, x, y, should_recreate_tabs);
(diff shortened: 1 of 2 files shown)


src/Application.vala
```diff
@@ -68,8 +68,8 @@ namespace PantheonTerminal {
             }
         }
 
-        public PantheonTerminalWindow new_window_with_coords (int x, int y, bool should_recreate_tabs=true) {
-            var window = new PantheonTerminalWindow.with_coords (this, x, y, should_recreate_tabs);
+        public PantheonTerminalWindow new_window_with_coords (int x, int y, bool should_recreate_tabs=true, bool ensure_tab=true) {
```
jeremypw: Should be spaces around "=".  Better default to "false".

PR #204, opened 2018-01-22
Grab focus to terminal widget when restoring

### LINKED ISSUE #203: Terminal not in focus
After terminal open, I can't write commands, because shell area not in focus. I need to click on it.

Version - 0.4.3+r1047

Thalexander:
The tab was being focused not the terminal widget.

Fixes #203


Thalexander:
I have done so, although I am noticing some unusual cursor behaviour with this. I am not sure if that is because of this commit or if it just makes it more visible.

jeremypw:
This effect can be fixed by grabbing focus in an Idle loop rather inside the signal handler.

PR #220, opened 2018-03-11
Meson build

danrabbit:


danrabbit:
Since I have an approve from Corentin, I'm going to merge to avoid more translation conflicts. I also have translators waiting on updated translation files that this branch provides.

I'll propose a branch to fix packaging shortly

PR #230, opened 2018-04-05
Update some deprecated posix signals

mdh34:
This only builds with vala 0.40


tintou:
Hint: something like this:
```if meson.get_compiler('vala').version().version_compare('>=0.40')
    error('Missing required dependency: valac >= 0.40')
endif```

Issue #233, opened 2018-04-27
Opening a second window should not reopen tabs from the first window


jlnr:
When I have a single Terminal window with a few tabs, and I open a second window (e.g. middle-click on Plank icon), the new window contains the same tabs as the first window.

Expected behavior, coming from web browsers: The new window only contains a single "default state" tab (home folder).

PR #234, opened 2018-04-29
Do not restore tabs when opened with commandline -w parameter

jeremypw:
Partly addresses issue #233.

The constructor `PantheonTerminalWindow.with_working_directory () now does not restore tabs nor does it save tabs when exiting.

This fixes the linked issue with respect to opening from Files.

Opening from Plank will be another PR.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -173,7 +173,7 @@ namespace PantheonTerminal {
                 window.add_tab_with_working_directory (working_directory);
                 window.present ();
             } else
-                new PantheonTerminalWindow.with_working_directory (this, working_directory, true);
+                new PantheonTerminalWindow.with_working_directory (this, working_directory);
         }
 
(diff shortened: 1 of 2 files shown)


jlnr:
I don't think `-w` windows should behave differently from other terminal windows wrt. saving tabs. If I open my only terminal window via Files, I would still expect it to store its tabs. Likewise, if I open a second terminal window via Plank and close the first one, I would expect the second window to save its tabs when I turn off the computer. I think that is how it works in browsers too.

danrabbit:
I think I agree with @jlnr that the best solution might be to assume a single-window workflow and restore tabs unless there is already another window open

@cassidyjames thoughts?

jeremypw:
I guess the smart thing to do in that case would be to check whether any of the restored tabs match the requested path and if so not to open another tab at that uri?  Otherwise #245 would not be fixed.

PR #247, opened 2018-05-20
Explicitly declare menuitems as translatable

### LINKED ISSUE #246: Context menu entries is untranslatable
For example, in russian - https://github.com/elementary/terminal/blob/master/po/ru.po

danrabbit:
Fixes #246


Djaler:
This doesn't work for me

PR #253, opened 2018-06-05
Fix menu translation

### LINKED ISSUE #246: Context menu entries is untranslatable

leonardo-lemos:
Fix Menu action translations and some menu deprecations. This PR fixes #246.

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -155,7 +155,6 @@ namespace PantheonTerminal {
             }
 
             main_actions = new GLib.SimpleActionGroup ();
-            //main_actions.set_translation_domain ("pantheon-terminal");
             main_actions.add_action_entries (main_entries, this);
 
             clipboard = Gtk.Clipboard.get (Gdk.Atom.intern ("CLIPBOARD", false));
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -44,13 +44,13 @@ namespace PantheonTerminal {
         public bool focus_restored_tabs { get; construct; default = true; }
         public bool recreate_tabs { get; construct; default = true; }
         public bool restore_pos { get; construct; default = true; }
-        public Gtk.UIManager ui { get; private set; }
+        public Gtk.Menu menu { get; private set; }
         public PantheonTerminalApp app { get; construct; }
         public SimpleActionGroup actions { get; construct; }
         public TerminalWidget current_terminal { get; private set; default = null; }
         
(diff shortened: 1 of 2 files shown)


Djaler:
Now it works good, except "Search". Looks like it even not exists in pot and po. So, can you also update them?

leonardo-lemos:
@Djaler Yes, I can, but not in this PR.

src/PantheonTerminalWindow.vala
```diff
@@ -166,30 +154,37 @@ namespace PantheonTerminal {
...
-            }
+            var copy_menuitem = new Gtk.MenuItem.with_label (_("Copy"));
+            var paste_menuitem = new Gtk.MenuItem.with_label (_("Paste"));
+            var select_all_menuitem = new Gtk.MenuItem.with_label (_("Select All"));
+            var search_menuitem = new Gtk.MenuItem.with_label (_("Search"));
+            var show_in_file_browser_menuitem = new Gtk.MenuItem.with_label (_("Show in File Browser"));
 
-            Gtk.AccelGroup accel_group = ui.get_accel_group ();
-            add_accel_group (accel_group);
+            copy_menuitem.activate.connect (action_copy);
```
danrabbit: Instead of doing this, we should set `action_name` on the menuitems. This will handle sensitivity

src/PantheonTerminalWindow.vala
```diff
@@ -139,6 +137,11 @@ namespace PantheonTerminal {
             action_accelerators[ACTION_ZOOM_IN_FONT] = "<Control>KP_Add";
             action_accelerators[ACTION_ZOOM_OUT_FONT] = "<Control>minus";
             action_accelerators[ACTION_ZOOM_OUT_FONT] = "<Control>KP_Subtract";
+            action_accelerators[ACTION_COPY] = "<Control><Shift>c";
+            action_accelerators[ACTION_PASTE] = "Control><Shift>v";
+            action_accelerators[ACTION_SEARCH] = "<Control><Shift>f";
+            action_accelerators[ACTION_SELECT_ALL] = "<<Control><Shift>a";
```
ryonakano: I think the line 143 should be `action_accelerators[ACTION_SELECT_ALL] = "<Control><Shift>a";`.
Because after this update was installed to my environment, pressing `Shift + a` selects all in the terminal and pressing `Ctrl + Shift + a` do nothing.

ryonakano: I'm sorry, it fixed just now #257

PR #254, opened 2018-06-06
Refactor TerminalWindow actions and menu

jeremypw:
* Replaces deprecated code (GtkActionEntry).
* Add some additional accelerators
* Modify search actions to be more like io.elementary.code

One side-effect of doing it this way is that the popup shows the accelerators.  Although this aids discovery I am not sure whether that is approved design.  If necessary this can be fixed by removing the action accelerators and handling these in the key_press_event closure.


jeremypw:
This seems to have been superceded recently.

PR #255, opened 2018-06-06
Add search next and previous to actions

danrabbit:
Ports Jeremy's changes from #254

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -65,6 +65,8 @@ namespace PantheonTerminal {
         public const string ACTION_COPY = "action_copy";
         public const string ACTION_PASTE = "action_paste";
         public const string ACTION_SEARCH = "action_search";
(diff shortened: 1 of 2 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -157,6 +161,9 @@ namespace PantheonTerminal {
                 app.set_accels_for_action (ACTION_PREFIX + action, action_accelerators[action].to_array ());
             }
 
+            app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_NEXT, {"<Control>g", "<Control>Down"});
+            app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_PREVIOUS, {"<Control><Shift>g", "<Control>Up"});
```
jeremypw: Why not extend the action_accelerators multimap?

danrabbit: Just seemed like a lot more work :grimacing:

PR #256, opened 2018-06-06
Add Icon Translation Comment

leonardo-lemos:


ryonakano:
I'm not sure this is correct solution, but I can confirm that this works by adding `preset: 'glib'` in `po/extra/meson.build`.

```
i18n.gettext('extra',
    args: ['--directory='+meson.source_root(), '--from-code=UTF-8'],
    install: false,
    preset: 'glib'
)
```

I'm sorry if I am wrong.

PR #259, opened 2018-06-08
Buttons lose focus when finished

### LINKED ISSUE #241: Settings gear icon button shouldn't retain keyboard focus
- Click the settings gear
- Change settings...
- Click settings gear again
- Start typing the next shell command

Expected result:
- Text input goes to current tab

Actual result:
- Nothing, because the keyboard focus is still on the button

jeremypw:
Fixes #241 

Alternative solution which does not prevent focusing of buttons e.g. via keyboard, but loses focus after action completed.


jeremypw:
Fixed conflicts with master.

PR #262, opened 2018-06-09
Copy last output action

### LINKED ISSUE #209: Add easier way to copy output from previous command
Right now you have to highlight from the bottom and scroll up (often many pages) to copy the full output of a command.

There should be a right click menu item that would take the full output of the last command and copy to clipboard.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55240649-add-easier-way-to-copy-output-from-previous-command?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #209

This adds a "copy last output" action linked to the context menu and a hotkey (current `<Alt>c` but that can be easily changed).

At the moment, the command line is also included but facility to exclude this is built in for possible future use.

There are some limitations at present;  

* It is assumed the command is on one line - otherwise only the last line of the command is copied.
* There is no user feedback as to whether any output was copied
* The view does not scroll to the top of the output.
*  The output is not marked selected (does not seem to be a simple way of doing this).


cassidyjames:
This seems to work as expected UX-wise, and is an awesome feature! I think it's fine that there's not a confirmation that it was copied, as copying doesn't usually show any sort of confirmation. Same with scrolling to the top and not highlighting the selection: I think this behavior is fine.

Assuming the command was on one line should _probably_ be fixed before this is merged, though.

src/Widgets/TerminalWidget.vala
```diff
@@ -343,6 +353,10 @@ namespace PantheonTerminal {
         }
 
         private void clickable (string[] str) {
+//            if (!get_realized ()) {
+//                return;
+//            }
+
```
tintou: Do not keep commented code

src/PantheonTerminalWindow.vala
```diff
@@ -63,6 +63,7 @@ namespace PantheonTerminal {
         public const string ACTION_ZOOM_IN_FONT = "action_zoom_in_font";
         public const string ACTION_ZOOM_OUT_FONT = "action_zoom_out_font";
         public const string ACTION_COPY = "action_copy";
+        public const string ACTION_COPY_LAST_OUTPUT = "ACTION_COPY_LAST_OUTPUT";
```
tintou: Keep it lowercase for consistency

tintou:
```suggestion
        public const string ACTION_COPY_LAST_OUTPUT = "action_copy_last_output";
```

jeremypw:
Oops - I forgot the request to fix the multiline command limitation.  I'll have a look at this asap. As I recall it was not straight forward.

PR #263, opened 2018-06-09
Do not pass Esc characters to terminal.

### LINKED ISSUE #249: Incorrect keys behaviour after Esc
Steps to reproduce:
1. Open Terminal
2. Press Esc
3. Press `a`

Expected: `a` typed
Actual: like "Enter" key pressed

Another example:
Press `p` after Esc -> works like an `Up` key

jeremypw:
Fixes #249 

Is there any reason to pass typed Esc characters to the terminal if no foreground process?  What about modified Esc characters?

Is an audible signal required for ignored key (audibility subject to gnome desktop sound setting)?


davidmhewitt:
Apologies, that should have been something I could have tested myself. I can confirm that this pretty much breaks Vim (making it even harder to exit :stuck_out_tongue: ) and presumably anything else that needs the escape key.

jeremypw:
OK, thanks - I should have thought of that - I thought it was too easy!.  I'll see if it can be fixed.

Djaler:
I don't think audible feedback is necessary in this case

jeremypw:
@elementary/ux  What is the ruling on whether an audible feedback is needed when a key press is ignored?

jeremypw:
I guess that in any case it should honor the "event sounds" setting.

jeremypw:
Withdrawing as no consensus as to whether this is correct approach after nearly a year.

Issue #215, opened 2018-03-04
Change/disable keyboard shortcuts


Janiczek:
I hit the "reset zoom to 100%" keyboard shortcut quite often (accidentally). It would be great to be able to change or remove it.

jeremypw:
I am not sure this is a specifically Terminal issue.  Other elementary apps do not provide a way of changing shortcuts and there needs to be a common solution (if one is implemented).

Janiczek:
The left panel of Keyboard > Shortcuts switchboard could be a good place for applications to plug themselves into? :)

![global_keyboard_shortcuts](https://user-images.githubusercontent.com/149425/41217616-b9e7b800-6d58-11e8-874b-6c5a6ea1223e.png)

jeremypw:
A lot of in-application shortcuts are hard-coded accelerators so there needs to be some infra-structure that applications use instead of hard-coding before they can be configured at system level.

PR #264, opened 2018-06-09
Scroll to last command

### LINKED ISSUE #238: Scroll to nearest command execution
Sometimes when executing a command the output is so long that scrolling through it to the top also takes long time. There should be kind of shortcut or way to scroll to the line where last command was executed.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57869943-scroll-to-nearest-command-execution?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #238 

This is based on #262 which should be reviewed and merged first.

The accelerator chosen was `<Alt>Up` but this can be changed if necessary.

Limitations:
* Will only scroll to the last command - further pressing of the accelerator will not find previous commands - a history of command positions would have to be implemented for that.
* The cursor position remains where it is - not moved to the command as the desired behaviour is not clear and moving the cursor would mean "copy last output" could not be used following this action.


cassidyjames:
I'm not sure about the specific shortcut, but I like the idea and direction here. I think the listed limitations are fine.

leonardo-lemos:
@jeremypw Can you resolve conflicts, please?

jeremypw:
@leonardo-lemos Conflicts resolved.

calderov:
Is there a way to disable the <Alt>Up shortcut? It conflicts with the Micro editor default shortcut to move lines around, so we can assume other apps may be impacted too. It would be nice if this feature could be turned off or at least the shortcut could be changed.

jeremypw:
@calderov Would you mind opening a request issue regarding this?  It is kind of related to #215 but it might be possible to special case this.

m93a:
This hardcoded shortcut really bothers me. I have Alt+Up bound to "cd .." in my setup, but there is currently no way to make it work in io.elementary.terminal. I might be forced to change to a different terminal emulator :cc

jeremypw:
I found this on StackOverFlow:

> VTE based terminals, such as e.g. GNOME Terminal, GNOME Console, Xfce4 Terminal, Terminator, Tilix, Guake and more, will support this feature with the forthcoming GNOME 46 (VTE 0.76) release in March 2024.

>The hotkey is hardcoded for the time being, it's Ctrl+Shift+Left and Ctrl+Shift+Right.

>Under the hood, VTE uses the quasi-standard OSC 133 shell integration escape sequences to see where the prompts are, like several other terminals already do.

So it may be possible to discontinue implementing this shortcut ourselves in future. Meanwhile it would be possible to change it to Gnome Terminal's choice  (although there is always a possibility that someone will have customised those).

PR #266, opened 2018-06-10
Do not process keystrokes when there is a foreground process.

### LINKED ISSUE #212: Ctrl+V vim shortcut, make it possible to switch off Ctrl+V for pasting text
Hi, I use Vim as my terminal text editor. One of the vim features — shortcut for visual selection. But it is conflict with elementary-terminal which paste text from buffer. Is it possible to provide users an option for turning it off? Generally the shortcut for pasting text in terminal emulator: Ctrl+Shift+V.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55660001-ctrl-v-vim-shortcut-make-it-possible-to-switch-off-ctrl-v-for-pasting-text?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #212 

This PR causes the Files window to ignore all key presses if there is a foreground process in the current terminal. So, for example, Ctrl+V behaves as expected when vim is running.

Are there any keystrokes that take priority over the foreground process?


Shnatsel:
Interesting. Does this mean that Ctrl+C stops the foreground process if there is one, and copies if there is none?

How does this interact with Ctrl+C used to clear the current line in bash?

voldyman:
But it will break the workflow for anyone who doesn’t use vim.

The mental model went from

* use ctrl-c/v top copy-paste

To

* ctrl-c/v only works in the shell

Which people will translate into ctrl-c/v works sometimes

jeremypw:
I suspected this might be too simplistic.  Maybe if TerminalWindow processes keystrokes *after* the foreground process? Then if the foreground process does not consume it will work normally.  A new setting should be avoided if possible.

jeremypw:
It seems reasonable to me that when a foreground process is running the user expects keystrokes to interact with that, not the shell, and behave accordingly.  Whether Ctrl+C copies or not will depend on what the process if and whether there is anything selected.  In the Terminal (without a foreground process), Ctrl+C will copy if there is a selection, otherwise it will be passed the shell and (in the case of bash) clear the line. Again, it seems reasonable that if there is a selection, the user expects keystrokes to  interact with that.

jeremypw:
Any actual examples of where a foreground process is running in the shell (not a launched gui) and the user would expect keystrokes to be handled by Terminal would be useful.

If Terminal processes keystrokes *after* the shell then various behaviours such as <Alt>+[1-8] to change tab no longer work, (because the shell processes them) so I do not think that is a viable solution.

If it helps, I can make this PR apply *only* to <Ctrl>+v (as mentioned in the issue description).

Shnatsel:
>  In the Terminal (without a foreground process), Ctrl+C will copy if there is a selection, otherwise it will be passed the shell and (in the case of bash) clear the line (and end any foreground process)

Sadly, this behavior is still immensely confusing because it's easy to select some text in a scrolling terminal and forget about it. No selected text is visible (because not all of shell log fits on the screen), but Ctrl+C still inexplicably refuses to clear the line.

Shnatsel:
> If it helps, I can make this PR apply only to +v (as mentioned in the issue description).

Actually no, it doesn't. There are cases when I want to paste into foreground process - mostly text input promts (SSH key generation and lots of others), text editors such as nano, etc.

jeremypw:
Note that this PR does not suppress <Ctrl>+v - it is passed onto the foreground process.  There if the foreground process supports pasting with this key combination it should still work.  It is questionable whether Terminal should interfere with the UI of other processes although I can see the consistency argument.  If the foreground process/shell swallows <Ctrl>+v it will not be possible for Terminal to act on it afterwards anyway.

Is Vim the only example of a foreground process that uses <Ctrl>+v for something other than paste?  If so I doubt it is justified exposing the natural-paste setting in the UI.

I suppose, theoretically Terminal could check whether the foreground process is Vim and pass on <Ctrl>-v only in that case - but that is unlikely to be approved.

jeremypw:
I have confirmed that `nano` at least swallows `<Ctrl>+v` so I am coming to the conclusion there is no solution that will please everyone.

jeremypw:
There does not seem to be any clear conclusion regarding this PR so I will close it for now.  Feel free to reopen - I will not delete the branch for a while.

Shnatsel:
I feel the only proper solution is to copy text on selecting it (and clearly communicating that the text was copied), dropping the selection after copying, and not special-casing Ctrl+C key combination.

fabiopicchi:
> Note that this PR does not suppress` <Ctrl>+v` - it is passed onto the foreground process. There if the foreground process supports pasting with this key combination it should still work. It is questionable whether Terminal should interfere with the UI of other processes although I can see the consistency argument. If the foreground process/shell swallows `<Ctrl>+v` it will not be possible for Terminal to act on it afterwards anyway.
> 
> Is Vim the only example of a foreground process that uses`<Ctrl>+v` for something other than paste? If so I doubt it is justified exposing the natural-paste setting in the UI.
> 
> I suppose, theoretically Terminal could check whether the foreground process is Vim and pass on `<Ctrl>-v` only in that case - but that is unlikely to be approved.

I feel that this is the best solution even though `nano` swallows `<Ctrl>+v`. `nano`, `vim` and terminals have been around for years and they exist before the `<Ctrl>+v' convention for pasting. I understand that people not familiar with terminals would enjoy the ability to intuitively copy and paste text but I would argue that they would definitely not use `vim` or even `nano`.

Furthermore, terminals in Windows definitely do not use the `<Ctrl>+c` `<Ctrl>+v' shortcuts for copying and pasting. I think Cygwin uses `<Shift> + <Insert>` and who know what cmd uses (I still right click the application menu and go to Edit > Paste).

Finally, the gnome terminal uses `<Ctrl>+<Shift>+v' for pasting and it works nicely overall. Never came across a situation where it was a problem. You can even use it inside vim instead of configuring it to read data from the clipboard and using the cumbersome `"+p`.

PR #275, opened 2018-06-22
Fix API break with Vte.feed_child

danrabbit:


src/Widgets/TerminalWidget.vala
```diff
@@ -406,16 +406,16 @@ namespace PantheonTerminal {
                         }
                     }
 
-                    string uris_s = string.joinv ("", uris);
-                    this.feed_child (uris_s, uris_s.length);
+                    var uris_s = (char []) string.joinv ("", uris);
```
tintou: don't case here

src/Widgets/TerminalWidget.vala
```diff
@@ -406,16 +406,16 @@ namespace PantheonTerminal {
...
+                    this.feed_child (uris_s);
```
tintou: then use `this.feed_child (uris_s.data);` instead

src/Widgets/TerminalWidget.vala
```diff
@@ -406,16 +406,16 @@ namespace PantheonTerminal {
...
 
                     break;
                 case DropTargets.STRING:
                 case DropTargets.TEXT:
-                    var data = selection_data.get_text ();
+                    var data = (char []) selection_data.get_text ();
```
tintou: don't case here

src/Widgets/TerminalWidget.vala
```diff
@@ -406,16 +406,16 @@ namespace PantheonTerminal {
...
 
                     if (data != null) {
-                        this.feed_child (data, data.length);
+                        this.feed_child (data);
```
tintou: then use `this.feed_child (uris_s.data);` instead

PR #287, opened 2018-08-14
Use Consistent Keybindings

nuclearfall:
The keybinding for opening a new tab in Epiphany, Code, and Files are all simply <Control>t. It seems that most of the keybindings associated with VTE are unnecessarily <Control><Shift>. Applications should have consistent keybindings for the same features when possible. 

I propose changing keybindings for New Window, New Tab, Close Tab, Select All, and Find to the hotkeys used by other applications: Ctrl-N, Ctrl-T, Ctrl-W, Ctrl-A, and Ctrl-F respectively. These seem fairly well standardized across all other elementary applications that use these functions.

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -129,10 +129,10 @@ namespace PantheonTerminal {
         }
 
         static construct {
-            action_accelerators[ACTION_CLOSE_TAB] = "<Control><Shift>w";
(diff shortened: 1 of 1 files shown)


danrabbit:
The reason that Terminal has a different shortcut scheme is that these common shortcuts interfere with BASH and BASH programs. See [this cheatsheet](https://github.com/fliptheweb/bash-shortcuts-cheat-sheet)

Issue #290, opened 2018-08-16
feed_child break compilation again


muriloventuroso:
In Elementary Juno I had the following error when compiling:
```../src/Widgets/TerminalWidget.vala:410.38-410.54: error: Argument 1: Cannot convert from `char[]' to `unowned string'
                    this.feed_child (uris_s.to_utf8 ());
                                     ^^^^^^^^^^^^^^^^^
../src/Widgets/TerminalWidget.vala:418.42-418.56: error: Argument 1: Cannot convert from `char[]' to `unowned string'
                        this.feed_child (data.to_utf8 ());
                                         ^^^^^^^^^^^^^^^```
This seems to be related to the version of libvte. It was working, but after a system upgrade the package apparently downgraded.
It has also been reported here: https://bugs.launchpad.net/ubuntu/+source/ubuntu-release-upgrader/+bug/1780501

**Comment by cassidyjames** [comment]
@codygarver mentioned this in Slack, but I'll post here so it's public:

>FYI the API change that 0.5.1 addressed got reverted in Bionic's vte. I added a packaging patch to revert the commit 8a87b49 so the build works again.
>
>So if you try to build Terminal from source on Bionic and it fails, it's due to 8a87b49.
>
>Ubuntu reverted it because Bionic isn't supposed to get an API break like that, it's still present in upstream vte

I think the recommended solution is an IFDEF in the code, so that it can also still compile correctly on Fedora and the sort.

**Comment by cassidyjames** [comment]
@codygarver what would you recommend we switch on for an IFDEF?

**Comment by ricotz** [comment]
This bionic-vte meson option should be false by default

**Comment by danrabbit** [comment]
Gonna drop this from the beta2 milestone because it doesn't affect Juno

**Comment by cnd** [comment]
same problem here, can't compile
```
TerminalWidget.vala:411.38-411.43: error: Argument 1: Cannot convert from `unowned string' to `unowned char[]?'
                    this.feed_child (uris_s, uris_s.length);
                                     ^^^^^^
```

davidmhewitt:
Going to close as I haven't heard anything about this in years. CI is passing.

PR #291, opened 2018-08-19
Reduce headerbar height by disabling subtitle

MrVoltz:
I was struggling with this for a while, finally found the cause today.

Quote from docs (https://developer.gnome.org/gtk3/stable/GtkHeaderBar.html#gtk-header-bar-set-subtitle):
> Note that GtkHeaderBar by default reserves room for the subtitle, even if none is currently set. If this is not desired, set the “has-subtitle” property to FALSE.

Subtitle isn't used anywhere in Terminal, so this shouldn't break anything.


cassidyjames:
This looks fine to me. Thanks! The CI is failing due to #290, but this builds fine otherwise.

PR #292, opened 2018-08-20
IFDEF Ubuntu VTE patches

### LINKED ISSUE #290: feed_child break compilation again

cassidyjames:
Probably needs better names, but this at least gets it building on elementary OS (since that variable isn't defined anywhere). Fixes #290.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -407,15 +407,22 @@ namespace PantheonTerminal {
                     }
 
                     var uris_s = string.joinv ("", uris);
+#if NON_UBUNTU_PATCHED_VTE
                     this.feed_child (uris_s.to_utf8 ());
-
+#else
+                    this.feed_child (uris_s, uris_s.length);
+#endif
                     break;
                 case DropTargets.STRING:
                 case DropTargets.TEXT:
                     var data = selection_data.get_text ();
 
                     if (data != null) {
+#if NON_UBUNTU_PATCHED_VTE
                         this.feed_child (data.to_utf8 ());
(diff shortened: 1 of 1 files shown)


codygarver:
Superseded by 914d4b0

danrabbit:
@codygarver can we make sure we’re using pull requests and not pushing things to master please. Having code reviews is important and we need to make sure we’re leading by example

PR #299, opened 2018-10-08
294 reverse phrasing

ls42:
This takes care of

* Rephrasing to a more positive version
* Tick CheckButton by default
* Add translation string to po files

Tested it, worked for me :smiley:

### DIFF
--- a/src/Dialogs/UnsafePasteDialog.vala
+++ b/src/Dialogs/UnsafePasteDialog.vala
@@ -44,10 +44,10 @@ public class PantheonTerminal.UnsafePasteDialog : Gtk.Dialog {
         );
         secondary_label.xalign = 0;
 
-        var do_not_show_check = new Gtk.CheckButton.with_label (_("Do not show this dialog again"));
-        do_not_show_check.margin_bottom = 12;
-        do_not_show_check.margin_top = 12;
-        settings.schema.bind ("unsafe-paste-alert", do_not_show_check, "active", SettingsBindFlags.DEFAULT | SettingsBindFlags.INVERT_BOOLEAN);
+        var show_protection_warnings = new Gtk.CheckButton.with_label (_("Show paste protection warnings"));
+        show_protection_warnings.margin_bottom = 12;
+        show_protection_warnings.margin_top = 12;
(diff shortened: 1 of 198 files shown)


src/Dialogs/UnsafePasteDialog.vala
```diff
@@ -70,6 +70,8 @@ public class PantheonTerminal.UnsafePasteDialog : Gtk.Dialog {
         add_action_widget (cancel_button, 1);
         add_action_widget (ignore_button, 0);
 
+        show_protection_warnings.set_active (true);
```
cassidyjames: If this i bound to the setting, does it need to be explicitly marked as active?

cassidyjames: I tested locally and it appears to still properly check itself when bound to the setting, so I believe this should be removed.

ls42: I added it so the dialog still has the same default behaviour: Showing the dialog if the CheckButton is untouched by the user. The old phrasing requires the user to act to mute the dialog. The new phrasing should still require the user to act.

I think this is required to fulfill the requirement "and have it checked by default." that your made in the issues description.

Sorry in advance if maybe I don't get what your mean.

cassidyjames: @ls42 No worries, it's kind of confusing! But by using a binding to the setting, the check will automatically check itself if that underlying setting is active. The setting is already active by default (controlled by the default GSchema), and it must be active to show the dialog in the first place, so this line is just unnecessary.

ls42: Alright! Very interesting. I remove the line and it's still working.

ls42:
I removed the changes to the po files. I'm not sure why they made the CI fail, though.

cassidyjames:
@ls42 it looks like CI is failing in master right now, so it's not your branch that's doing it.

ls42:
>  Thanks for the contribution

You're very welcome. I hope I can get involved even more, I really enjoy using elementaryOS and hope it really takes off (number-of-happy-users-wise)!

Issue #244, opened 2018-05-14
ZSH integration is not updated to the new D-bus API for completion notifications


codygarver:
More details in issue #214

PR #305, opened 2018-10-19
Fix zsh process completion notification integration

blaknite:
The zsh process completion notification script does not work with the latest version of the app due to the exit code not being passed. This PR fixes this issue in the integration script.

For compatibility the integration script now appends the pre-command function to the existing list.

This resolves issue #244


blaknite:
I don't believe the failed build is related to my changes. Could I get some assistance with that?

PR #307, opened 2018-10-19
Remove <Control>g and <Control><Shift>g keybindings

### LINKED ISSUE #306: Make <control>-g optional keybinding
Currently `<Control>g` is bound to `ACTION_SEARCH_NEXT`. When I use emacs inside the terminal ctrl-g is used to cancel any action. This conflict with the terminal's bindings. Since there are two key bindings for `ACTION_SEARCH_NEXT` how about having the `<Control>g` to be either optional or removed.

satran:
`<Control>g` is used in Emacs to cancel out actions. Since `ACTION_SEARCH_NEXT` and `ACTION_SEARCH_PREVIOUS` have other keybindings `<Control>Up` and `<Control>down` the `<Control>g` could be removed without giving up on keybindings.

I'm not entirely sure if removing it is the correct approach. I'm not sure why `<Control>g` was used in the first place but I could add an alternate keybinding if it is necessary.

fixes #306

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -144,9 +144,7 @@ namespace PantheonTerminal {
             action_accelerators[ACTION_COPY] = "<Control><Shift>c";
             action_accelerators[ACTION_PASTE] = "<Control><Shift>v";
             action_accelerators[ACTION_SEARCH] = "<Control><Shift>f";
-            action_accelerators[ACTION_SEARCH_NEXT] = "<Control>g";
             action_accelerators[ACTION_SEARCH_NEXT] = "<Control>Down";
-            action_accelerators[ACTION_SEARCH_PREVIOUS] = "<Control><Shift>g";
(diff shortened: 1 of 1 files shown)


alexander-95:
Since `<ctrl>+g` and `<ctrl>+<shift>+g` are used some of the other built in apps I would guess that simply removing these from the terminal would be undesirable. I've also been experimenting with this and I've found an alternative solution: to only assign keybindings to `ACTION_SEARCH_NEXT` and `ACTION_SEARCH_PREVIOUS` when the search bar is active. This gives us the best of both worlds, being able to use  `<ctrl>+g` and `<ctrl>+<shift>+g` when using terminal's search feature and being able to use `<ctrl>+g` for emacs. Take a look at the change here:
https://github.com/alexander-95/terminal/commit/98356c310037017f30978de30086f61819d6247a

alexander-95:
Yeah, I'll make a pull request either tonight or tomorrow morning.

PR #309, opened 2018-10-21
Fix ctrl g

### LINKED ISSUE #306: Make <control>-g optional keybinding

alexander-95:
This change is to allow the use of `<ctrl>+g` in emacs. The problem is that instead, it gets grabbed by the terminal for `ACTION_SEARCH_NEXT`. My fix is to make the shortcuts for `ACTION_SEARCH_NEXT` and `ACTION_SEARCH_PREVIOUS` conditional on whether the search bar is active. This allows use of `<ctrl>+g` for the terminal's search feature and for emacs when the search bar is not active.

More discussion can be found on github here https://github.com/elementary/terminal/pull/307 and on stack exchange here https://elementaryos.stackexchange.com/questions/16462/controlg-inside-terminal-is-not-functioning-correctly/16538#16538

fixes #306

### DIFF
--- a/po/en_GB.po
+++ b/po/en_GB.po
@@ -8,8 +8,8 @@ msgstr ""
 "Project-Id-Version: pantheon-terminal\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-10-14 11:18-0700\n"
-"PO-Revision-Date: 2018-04-29 14:52+0000\n"
-"Last-Translator: Ciaran Ainsworth <cda@whistlingkappa.com>\n"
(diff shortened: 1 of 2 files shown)


alexander-95:
Now that the build is passing, it would be nice to get this looked at for the sake of fellow emacs users.

Issue #314, opened 2018-10-29
Terminal quits when called with `-h` or `--help` flags.


psetq:
When I attempt to launch the terminal with the -h or --help flags the terminal quits.

For example, I launch the terminal with `SUPR-T`, then in the terminal application I type 

    io.elementary.terminal --help

The terminal application where I typed that last command closes.

The `--version` flag for the terminal app doesn't seem to return a version.

    ~$ io.elementary.terminal --version
    [DEBUG 09:47:22.006980] Settings.vala:163: Loading settings from schema 'io.elementary.terminal.saved-state'
    [DEBUG 09:47:22.007614] Settings.vala:163: Loading settings from schema 'io.elementary.terminal.settings'

I'm using elementary OS 5.0 Juno.

Thalexander:
The help issue is a duplicate of #205.

arshubham:
Also duplicate for https://github.com/elementary/terminal/issues/282

Issue #282, opened 2018-07-12
Trying to display help info causes the app crash


ryonakano:
![peek 2018-07-12 19-14](https://user-images.githubusercontent.com/26003928/42627664-a1d6891c-8608-11e8-848d-405fd27f74db.gif)

Executing `io.elementary.terminal --help` immediately crashes the app.

On Juno Beta1

arshubham:
I installed gnome-terminal to test this out and I get the following message if pantheon-terminal is open and then it crashes.
`GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying` 

if pantheon-terminal is not open the help info is displayed normally.

Issue #205, opened 2018-01-23
Help parameter will close open terminal instances


Thalexander:
When running io.elementary.terminal -h any open windows will close. Other parameters such as version work as expected and do not affect open windows.

jeremypw:
Also happens with e.g. Files and Code so not a specific Terminal problem.

jeremypw:
Does not happen with Mate Terminal

jeremypw:
If automatic --help support is enabled (see g_option_context_set_help_enabled()), and the argv array contains one of the recognized help options, this function will produce help output to stdout and call exit (0).
Since Terminal is a unique instance, it closes along with any windows open.

danrabbit:
This doesn't seem to be the case anymore. Please reopen if you can reproduce :)

PR #315, opened 2018-10-30
Add option "-h" to show application help

psetq:
Proposed fix for issues #205 , #282 , and #314 . 

Implements the "-h" & "--help" option flags to show application help overriding the default exit behaviour to keep existing terminals open.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -25,6 +25,9 @@ public class PantheonTerminal.TerminalApp : Gtk.Application {
     [CCode (array_length = false, array_null_terminated = true)]
     private static string[]? command_e = null;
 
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -25,6 +25,9 @@ public class PantheonTerminal.TerminalApp : Gtk.Application {
     [CCode (array_length = false, array_null_terminated = true)]
     private static string[]? command_e = null;
 
+    /* option_help will be true if help flag was given. */
```
donadigo:
```suggestion
    // option_help will be true if help flag was given.
```

psetq: Thanks for the suggestion.

PR #318, opened 2018-11-02
Adds accels to tooltip markup

danrabbit:
The way the search toggle is handled here isn't as nice as the way it's done in Code, so sometimes the tooltip isn't updated there. I think fixing that mechanism is probably outside the scope of this branch though

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -251,20 +251,33 @@ namespace PantheonTerminal {
 
             search_button = new Gtk.ToggleButton ();
             search_button.image = new Gtk.Image.from_icon_name ("edit-find-symbolic", Gtk.IconSize.SMALL_TOOLBAR);
-            search_button.tooltip_text = _("Find…");
             search_button.valign = Gtk.Align.CENTER;
+            search_button.tooltip_markup = Granite.markup_accel_tooltip (
+                {"<Ctrl><Shift>f"},
+                _("Find…")
+            );
+
 
             var zoom_out_button = new Gtk.Button.from_icon_name ("zoom-out-symbolic", Gtk.IconSize.MENU);
-            zoom_out_button.tooltip_text = _("Zoom Out");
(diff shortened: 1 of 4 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -251,20 +251,33 @@ namespace PantheonTerminal {
...
+                application.get_accels_for_action (zoom_default_button.action_name),
+                _("Default zoom level")
+            );
 
             var zoom_in_button = new Gtk.Button.from_icon_name ("zoom-in-symbolic", Gtk.IconSize.MENU);
-            zoom_in_button.tooltip_text = _("Zoom In");
             zoom_in_button.action_name = ACTION_PREFIX + ACTION_ZOOM_IN_FONT;
+            zoom_in_button.tooltip_markup = Granite.markup_accel_tooltip (
+                application.get_accels_for_action (zoom_in_button.action_name),
+                _("Zoom In")
```
cassidyjames: So we were bad about this in Code, but I guess this is supposed to be sentence case? https://elementary.io/docs/human-interface-guidelines#text-buttons-tooltips

src/PantheonTerminalWindow.vala
```diff
@@ -251,20 +251,33 @@ namespace PantheonTerminal {
...
+                _("Find…")
+            );
+
 
             var zoom_out_button = new Gtk.Button.from_icon_name ("zoom-out-symbolic", Gtk.IconSize.MENU);
-            zoom_out_button.tooltip_text = _("Zoom Out");
             zoom_out_button.action_name = ACTION_PREFIX + ACTION_ZOOM_OUT_FONT;
+            zoom_out_button.tooltip_markup = Granite.markup_accel_tooltip (
+                application.get_accels_for_action (zoom_out_button.action_name),
+                _("Zoom Out")
```
cassidyjames: So we were bad about this in Code, but I guess this is supposed to be sentence case? https://elementary.io/docs/human-interface-guidelines#text-buttons-tooltips

PR #328, opened 2018-11-23
PantheonTerminalWindow: Make search action a boolean

### LINKED ISSUE #327: Find button tooltip doesn't changed
Find button tooltip changes from `Find…` to `Hide find bar` only if shortcut was used, but no if I just click on this button

danrabbit:
Fixes #327 

Makes the search action a boolean and binds the search button to the search action instead of connecting to the toggle signal.

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -1090,10 +1090,25 @@ namespace PantheonTerminal {
                 search_toolbar.clear ();
                 current_terminal.grab_focus ();
             }
-            app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_NEXT,
-                                       action_accelerators[ACTION_SEARCH_NEXT].to_array ());
-            app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_PREVIOUS,
-                                       action_accelerators[ACTION_SEARCH_PREVIOUS].to_array ());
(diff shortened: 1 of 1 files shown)


src/PantheonTerminalWindow.vala
```diff
@@ -1086,13 +1079,16 @@ namespace PantheonTerminal {
...
                 action_accelerators.remove_all(ACTION_SEARCH_NEXT);
                 action_accelerators.remove_all(ACTION_SEARCH_PREVIOUS);
                 search_button.tooltip_markup = Granite.markup_accel_tooltip (
                     {"<Ctrl><Shift>f"},
                     _("Find…")
                 );
+                search_toolbar.clear ();
+                current_terminal.grab_focus ();
             }
             app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_NEXT,
```
jeremypw: This results in a critical terminal warning when the accelerator collection is empty. It seems a null array is returned by to_array () instead of an empty one?  This may be an upstream bug.  A work around is:

```
var accels = action_accelerators[ACTION_SEARCH_NEXT].is_empty ? new string [] {} : action_accelerators[ACTION_SEARCH_NEXT].to_array ();
            app.set_accels_for_action (ACTION_PREFIX + ACTION_SEARCH_NEXT,
                                       accels);

```

PR #330, opened 2018-12-01
Add open in browser in context menu

### LINKED ISSUE #3: Show "Open Link" option in context menu when right click a url
when a url in entered in the terminal it is highlighted, when the user right clicks on the url there should be an option in the context menu to open the url.

Launchpad Details: [#LP1156710](https://bugs.launchpad.net/bugs/1156710) Akshay Shekher - 2013-03-18 17:17:02 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44562132-show-open-link-option-in-context-menu-when-right-click-a-url?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mohelm97:
Well, it's something :D
Fixes #3

### DIFF
--- a/src/PantheonTerminalWindow.vala
+++ b/src/PantheonTerminalWindow.vala
@@ -71,6 +71,7 @@ namespace PantheonTerminal {
         public const string ACTION_SELECT_ALL = "action_select_all";
         public const string ACTION_OPEN_IN_FILES = "action_open_in_files";
         public const string ACTION_SCROLL_TO_LAST_COMMAND = "action_scroll_to_last_command";
+        public const string ACTION_OPEN_IN_BROWSER = "action_open_in_browser";
 
         private static Gee.MultiMap<string, string> action_accelerators = new Gee.HashMultiMap<string, string> ();
 
@@ -92,7 +93,8 @@ namespace PantheonTerminal {
             { ACTION_SEARCH_PREVIOUS, action_search_previous },
             { ACTION_SELECT_ALL, action_select_all },
             { ACTION_OPEN_IN_FILES, action_open_in_files },
-            { ACTION_SCROLL_TO_LAST_COMMAND, action_scroll_to_last_command }
+            { ACTION_SCROLL_TO_LAST_COMMAND, action_scroll_to_last_command },
+            { ACTION_OPEN_IN_BROWSER, action_open_in_browser}
         };
 
         public PantheonTerminalWindow (TerminalApp app, bool recreate_tabs = true) {
@@ -185,6 +187,9 @@ namespace PantheonTerminal {
 
             primary_selection = Gtk.Clipboard.get (Gdk.Atom.intern ("PRIMARY", false));
 
+            var open_in_browser_menuitem = new Gtk.MenuItem.with_label (_("Open in Browser"));
+            open_in_browser_menuitem.set_action_name (ACTION_PREFIX + ACTION_OPEN_IN_BROWSER);
+
             var copy_menuitem = new Gtk.MenuItem.with_label (_("Copy"));
             copy_menuitem.set_action_name (ACTION_PREFIX + ACTION_COPY);
 
@@ -204,6 +209,7 @@ namespace PantheonTerminal {
             show_in_file_browser_menuitem.set_action_name (ACTION_PREFIX + ACTION_OPEN_IN_FILES);
 
             menu = new Gtk.Menu ();
+            menu.append (open_in_browser_menuitem);
             menu.append (copy_menuitem);
             menu.append (copy_last_output_menuitem);
             menu.append (paste_menuitem);
@@ -1023,6 +1029,20 @@ namespace PantheonTerminal {
 
             } catch (ConvertError e) {
                 warning (e.message);
+            }
+        }
+
+        void action_open_in_browser () {
+            if (current_terminal.uri != null) {
+                var uri = current_terminal.uri;
+
+                if (uri != null) {
+                    try {
+                        Gtk.show_uri (null, uri, Gtk.get_current_event_time ());
+                    } catch (GLib.Error error) {
+                        warning ("Could Not Open link");
+                    }
+                }
             }
         }
 
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -46,6 +46,7 @@ namespace PantheonTerminal {
                 this.app = value.app;
                 this.menu = value.menu;
                 this.menu.show_all ();
+                this.menu.get_children ().nth_data (0).set_visible (false);
             }
         }
 
@@ -129,10 +130,17 @@ namespace PantheonTerminal {
             /* Connect to necessary signals */
             button_press_event.connect ((event) => {
                 if (event.button ==  Gdk.BUTTON_SECONDARY) {
+                    // Reset the state of buttons when we lunch the context menu again, and we have a uri
+                    if (uri != null) {
+                        window.get_simple_action (PantheonTerminalWindow.ACTION_COPY).set_enabled (false);
+                        this.menu.get_children ().nth_data (0).set_visible (false);
+                    }
+
                     uri = get_link (event);
 
                     if (uri != null) {
                         window.get_simple_action (PantheonTerminalWindow.ACTION_COPY).set_enabled (true);
+                        this.menu.get_children ().nth_data (0).set_visible (true);
                     }
 
                     menu.select_first (false);


src/PantheonTerminalWindow.vala
```diff
@@ -1014,6 +1020,20 @@ namespace PantheonTerminal {
             }
         }
 
+        void action_open_in_browser () {
+            if (current_terminal.uri != null) {
```
jeremypw: Do you mean `if (current_terminal != null) {`? Otherwise this is redundant since you check for null uri later.

src/PantheonTerminalWindow.vala
```diff
@@ -71,6 +71,7 @@ namespace PantheonTerminal {
         public const string ACTION_SELECT_ALL = "action_select_all";
         public const string ACTION_OPEN_IN_FILES = "action_open_in_files";
         public const string ACTION_SCROLL_TO_LAST_COMMAND = "action_scroll_to_last_command";
+        public const string ACTION_OPEN_IN_BROWSER = "action_open_in_browser";
```
jeremypw: Should not use underscores in action names (I know the existing ones need changing) although only affects certain functions.

src/PantheonTerminalWindow.vala
```diff
@@ -1014,6 +1020,20 @@ namespace PantheonTerminal {
             }
         }
 
+        void action_open_in_browser () {
+            if (current_terminal.uri != null) {
+                var uri = current_terminal.uri;
+
+                if (uri != null) {
+                    try {
+                        Gtk.show_uri (null, uri, Gtk.get_current_event_time ());
```
jeremypw: Gtk.show_uri is deprecated.  Use Gtk.show_uri_on_window

src/Widgets/TerminalWidget.vala
```diff
@@ -129,10 +130,17 @@ namespace PantheonTerminal {
             /* Connect to necessary signals */
             button_press_event.connect ((event) => {
                 if (event.button ==  Gdk.BUTTON_SECONDARY) {
+                    // Reset the state of buttons when we lunch the context menu again, and we have a uri
```
jeremypw: Spelling:  "lunch" -> "launch"

src/Widgets/TerminalWidget.vala
```diff
@@ -46,6 +46,7 @@ namespace PantheonTerminal {
                 this.app = value.app;
                 this.menu = value.menu;
                 this.menu.show_all ();
+                this.menu.get_children ().nth_data (0).set_visible (false);
```
jeremypw: This relies on a certain menu order.  I would just disable the linked action and not hide the option.

src/Widgets/TerminalWidget.vala
```diff
@@ -129,10 +130,17 @@ namespace PantheonTerminal {
             /* Connect to necessary signals */
             button_press_event.connect ((event) => {
                 if (event.button ==  Gdk.BUTTON_SECONDARY) {
+                    // Reset the state of buttons when we lunch the context menu again, and we have a uri
+                    if (uri != null) {
+                        window.get_simple_action (PantheonTerminalWindow.ACTION_COPY).set_enabled (false);
+                        this.menu.get_children ().nth_data (0).set_visible (false);
+                    }
```
jeremypw: Why do you need this? You reset uri and the action in the following code.

Mohelm97: Yeah so this is an old not current behaviour, if you make the context menu on top of uri you will got the copy enabled, but if you close it and open it somewhere else no one changes the copy state so it will stay enabled, but It should not be enabled as there's nothing selected and we are not on top of uri

src/Widgets/TerminalWidget.vala
Mohelm97: I hide it because I thought the I would be annoying to have an almost always disabled menu item, so it's just the first one if you have a link, or it won't appear

Mohelm97: I don't know they only idea I came up with is storing this menu item but it's not a good idea too :(

src/Widgets/TerminalWidget.vala
jeremypw: Sorry yes, you are right. The first clause could be true and the second false (or vice versa) so both are needed.

src/Widgets/TerminalWidget.vala
jeremypw: In Files this problem is solved by building the context menu on the fly each time it is required since the context varies a lot. I'd have to study the code in more detail to see whether that is appropriate here though.

Mohelm97: I think editing the "Show in File Browser" is a better idea, I will just wait for the UX team to comment on this as I won't be able to edit anything today :D

jeremypw:
@elementary/ux Needs an opinion on whether to hide or disable the menuitem. This does not seem to be covered in the HIG.  I think the general wisdom is to disable rather than hide in most cases?

cassidyjames:
@jeremypw I am not 100% sure of the context, but I'd say: if the context menu is for something you're clicking on, only showing the menu item when you're clicking on that thing makes sense.

jeremypw:
@cassidyjames I agree it should only appear in a selection context, not when secondary-clicking on the background say.  So it only appear when the selection (or hovered item) is a web address (i.e it currently gets underlined)? That should be doable.

jeremypw:
@Mohelm97 : See UX team ruling on this ^  Will you be able to progress this now? :-)

jeremypw:
I am going to try and progress this PR myself now as it has been left a while.

jeremypw:
I have combined the two "open in browser" actions since they use the same command ultimately (`Gtk.show_uri_on_window`) so there is no explicit control how it is shown and also because it looked odd that when clicking on a uri two different open in browser options were displayed.  Initially the item label changed depending on whether it was expected to use a web browser or a file browser but this was not possible in any elegant way when Granite.AccelLabels are used since their labels are construct only.  Another factor is it seems excessive to have two different keyboard shortcuts and two actions leading to the same code.  Need design team comment on this though.

src/MainWindow.vala
```diff
@@ -186,11 +186,18 @@ namespace Terminal {
...
-            update_context_menu ();
             clipboard.owner_change.connect (update_context_menu);
 
             primary_selection = Gtk.Clipboard.get (Gdk.Atom.intern ("PRIMARY", false));
 
+            var open_in_browser_menuitem = new Gtk.MenuItem ();
+            open_in_browser_menuitem.set_action_name (ACTION_PREFIX + ACTION_OPEN_IN_BROWSER);
+            open_in_browser_menuitem.add (
+                new Granite.AccelLabel.from_action_name (
+                    _("Open in Browser"), open_in_browser_menuitem.action_name
```
danrabbit: I see how this is technically just a URI, but I don't think it's clear that the location will open in Files when this menuitem is triggered the majority of the time. This label feels like optimizing for the less-likely case that we're opening a web URL

I think this might be better off waiting for a version of Granite where we can easily update this label based on the URI scheme

jeremypw: OK, seems like a plan.

jeremypw:
Blocked until Granite.AccelLabel supports variable text.

jeremypw:
Variable label now supported in Granite master but leaving blocked until released.

src/Widgets/TerminalWidget.vala
cassidyjames: No, the "Show in File Browser" item relates to the whole view and not what's being clicked on specifically; that's why it's separated below the text items. An "Open in Browser" item should probably be grouped along with the text-related items up above. I'm not super picky about whether it's insensitive or hidden when it's irrelevant; I feel like for context menus when you're clicking on a specific item it makes sense to only show items relevant to that item, i.e. not showing the "Open in Browser" item at all unless a link is being clicked. But if it greatly simplifies things to always show it and only toggle sensitivity, I think we'd want it to be at the bottom of the text-relted itms, at least.

@danrabbit thoughts?

jeremypw: There is precedent for the same menuitem applying to either an item or the background depending on the context - e.g. the "Open in" menuitem in Files.

jeremypw:
New release of Granite - unblocking.

jeremypw:
Now need to update the menuitem label according to the target as per requested change.

jeremypw:
The latest changes allows opening the current selection in a browser without it needing an explicit scheme.  So you can type e.g. `\home`, select it and press `<Shift><Ctrl>E` to open that directory in the file manager. The selection is not fully validated as a path as far as setting the action enabled is concerned (it must contain a `Path.DIR_SEPARATOR_S`) but invalid paths just do nothing (apart from generate a terminal warning).

src/MainWindow.vala
```diff
@@ -186,11 +188,18 @@ namespace Terminal {
...
             clipboard.owner_change.connect (update_context_menu);
 
             primary_selection = Gtk.Clipboard.get (Gdk.Atom.intern ("PRIMARY", false));
 
+            var open_in_browser_menuitem = new Gtk.MenuItem ();
+            open_in_browser_menuitem.set_action_name (ACTION_PREFIX + ACTION_OPEN_IN_BROWSER);
+            open_in_browser_menuitem_label = (new Granite.AccelLabel.from_action_name (
+                                                      "", open_in_browser_menuitem.action_name
+                                                  )
+                                             );
```
danrabbit, line 200: This is kind of a mess. Should probably be
```suggestion
            open_in_browser_menuitem_label = new Granite.AccelLabel.from_action_name (
                "", open_in_browser_menuitem.action_name
            );
```
```suggestion
            open_in_browser_menuitem_label = (new Granite.AccelLabel.from_action_name (
                                                      "", open_in_browser_menuitem.action_name
                                                  )
                                             );
```

src/MainWindow.vala
```diff
@@ -215,23 +224,14 @@ namespace Terminal {
...
-            show_in_file_browser_menuitem.set_action_name (ACTION_PREFIX + ACTION_OPEN_IN_FILES);
-            show_in_file_browser_menuitem.add (
-                new Granite.AccelLabel.from_action_name (
-                    _("Show in File Browser"),
-                    show_in_file_browser_menuitem.action_name
-                )
-            );
-
             menu = new Gtk.Menu ();
+            menu.append (open_in_browser_menuitem);
```
danrabbit, line 228: If this is moved (Which is fine, this is probably more consistent with menuitem placement elsewhere), we probably need to add a separator beneath it

src/MainWindow.vala
```diff
@@ -839,15 +840,44 @@ namespace Terminal {
             }
         }
 
-        private void update_context_menu () {
-            clipboard.request_targets (update_context_menu_cb);
```
danrabbit, line 843: It doesn't seem intentional to remove this. If it is, then `update_context_menu_cb` should also be removed since it is now never used

src/MainWindow.vala
```diff
@@ -839,15 +840,44 @@ namespace Terminal {
...
+        private void update_menu_label (string? uri) {
+            var scheme = uri != null ? Uri.parse_scheme (uri) : "";
+            get_simple_action (ACTION_OPEN_IN_BROWSER).set_enabled (true);
+
+            if (scheme == null || scheme == "") {
+                get_simple_action (ACTION_OPEN_IN_BROWSER).set_enabled (uri.contains (Path.DIR_SEPARATOR_S));
+            }
+
+            if (scheme == null || scheme == "" || OPEN_IN_FILE_MANAGER_SCHEMES.contains (scheme)) {
+                open_in_browser_menuitem_label.label = _("Open in File Manager");
```
danrabbit, line 869: Should probably avoid breaking translations here

```suggestion
                open_in_browser_menuitem_label.label = _("Show in File Browser");
```

jeremypw:
@danrabbit  Thanks for the review - I haven't looked at this in a while but "Open in Web Browser" should show with schemes such as "http" and "https:". Most other remote schemes do open in the file manager though - I'll check it out and implement your other suggestions.

jeremypw:
There will be a delay until I can get a system set up that will actually build the latest Terminal :-(

src/MainWindow.vala
jeremypw: OK, done.  It was moved to match the position of the "Open In" item in Files.

jeremypw:
Should be OK - careless mistake.

Are there any other schemes, apart from http and https that should be opened in Web Browser rather than Files?

danrabbit:
I can confirm that "Open in Web Browser" now shows correctly and works as expected, but it seems that "Show in File Browser" is now always insensitive.

I guess in an ideal world we should probably be using `https://valadoc.org/gio-2.0/GLib.AppInfo.get_default_for_uri_scheme.html` and `https://valadoc.org/gio-2.0/GLib.AppInfo.launch_default_for_uri.html` instead of trying to figure out what scheme does what manually

jeremypw:
@danrabbit Thanks for the review.  "Show in File Browser" should be sensitive if either
* there is a selection and that selection is a valid path or
* there is a link with a scheme file:/// under the cursor.

Good idea about `GLib.AppInf.get_default_for_uri_scheme()`.  I'll see what I can do.

jeremypw:
Using `AppInfo.get_default_for_uri_scheme ()` works but this has raised a number of problems (some of which were already present because this PR uses `Gtk.show_uri_on_window ()` which uses the default app - not just web or file browser).

* If you want the choice to be limited to "Web Browser" or "File Browser" there seems to be no good way to determine, in all locales, whether the returned app is a Web Browser.
* Some schemes such as "ftp" that we probably want to open in Files will by default open in the Web Browser.
* Some schemes such as "mailto://" that a user might want to open in a web browser if it is a webmail address will open in Mail.
* Some returned apps e.g. that for ssh:// are neither Web Browser nor File Browser.

If you want to allow other apps and have their name in the menu:
* Not all apps have a translated name or generic name in all locales and the app name or id may not be meaningful

If we want to stick to a simple Web/File browser choice then we will have to stop using `Gtk.show_uri_on_window ()` and use e.g. `AppInfo.launch_uris ()`

danrabbit:
In master, if there is no selection you can still show the pwd in the file browser which is extremely useful.

It's probably not within the scope of Terminal to decide the appropriate app that should be associated with a URI scheme. If we want Files to handle ftp:// schemes, we should probably set that in the default-settings package.

I think the way to account for users wanting to open something other than the default is to provide a full "Open In" submenu that would list every app that handles that URI scheme. If we follow the same design scheme as Files, I think this could be a separate pull request since we still show the default at the top level. There's an issue report for this feature: #422 

Can you explain what you mean about app names? I think we should just be able to do something like `"Open in %s".printf (appinfo.get_display_name ())` right? I believe this is what we do in other "Open In" menus

jeremypw:
@danrabbit The latest commits should implement opening the current link, selection or working directory in the appropriate default application.  As advised I have used appinfo.get_display_name () for the label.  When the "Show in " menu option is not enabled the description falls back to "Show in Default Application" as this is what it will do when it is enabled.

For simplicity the content type is guessed from the filename (if necessary).  I am not sure how complex we want to get with this - replicating how Files does it (i.e looking at the actual file attributes offering a way of changing the default), but that can be left for another PR if required.

src/MainWindow.vala
```diff
@@ -847,15 +848,50 @@ namespace Terminal {
...
+                var ctype = ContentType.guess (uri, null, out uncertain);
+                if (!uncertain) {
+                    appinfo = AppInfo.get_default_for_type (ctype, true);
+                }
+
+                if (appinfo == null) {
+                    /* Fallback to File Browser */
+                    appinfo = AppInfo.get_default_for_type ("inode/directory", true);
+                }
+            }
```
danrabbit, line 881: Can you explain what case this covers? I'm unsure what this is intended to do

src/MainWindow.vala
```diff
@@ -1182,21 +1218,53 @@ namespace Terminal {
             current_terminal.select_all ();
         }
 
-        private void action_open_in_files () {
-            try {
-                string uri = Filename.to_uri (current_terminal.get_shell_location ());
+        void action_open_in_browser () {
```
danrabbit, line 1221:
```suggestion
        private void action_open_in_browser () {
```

src/MainWindow.vala
jeremypw: This covers everything that 
* does not have an explicit scheme and/or
* Does not have a recognisable extension (i.e. the content type cannot be guessed)

For example, the pwd is opened through this fallback.  This is a simple, low code alternative to creating a GLib.File and trying to read the file attributes (which is what Files does).  I am not sure of what use cases people have and therefore how sophisticated it is worth making this.

src/MainWindow.vala
```diff
@@ -41,6 +42,7 @@ namespace Terminal {
         private const string DARK_FG = "#a5a5a5";
         private const string SOLARIZED_LIGHT_BG = "rgba(253, 246, 227, 0.95)";
         private const string SOLARIZED_LIGHT_FG = "#586e75";
+        private const string OPEN_IN_FILE_MANAGER_SCHEMES = "file|trash|network|recent|afp|dav|davs|ftp|sftp|smb|mtp";
```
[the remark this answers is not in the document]
jeremypw: Removed

src/MainWindow.vala
```diff
@@ -531,6 +531,7 @@ namespace Terminal {
                 current_terminal.grab_focus ();
             });
 
+
```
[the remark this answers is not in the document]
jeremypw: Oops

src/MainWindow.vala
jeremypw: Fixed

src/MainWindow.vala
jeremypw: After latest commits, no longer guess content.  Also interpret "~" and ".." in path. Needs testing with unusual characters in filename.

jeremypw:
Closing as the linked issue has been closed by another PR

Issue #322, opened 2018-11-11
What about a move tab shortcut?


Blast-City:
Hi.  

Right know you can select a tab by pressing Ctrl+Shift+Left/Right.  

What if there was also a shortcut to move tabs?

Ctrl+Shift+Up/Down , Ctrl+Alt+Left/Right or Ctrl+Shift+Alt+Left/Right.

Blast-City:
Thanks.

jeremypw:
Needs UX team decision on what shortcut to use (and whether another shortcut is desirable).  Ideally other core apps using a notebook should implement the same shortcut.

janxkoci:
Actually most apps I know (Files, Nautilus, Thunar, Sublime text, web browsers like Epiphany, Firefox, Chrome, all terminal emulators with tabs, like gnome terminal, tilix etc, the list goes on) use these shortcuts:

- **Ctrl+PgUp / Ctrl+PgDn** --> switch to previous / next tab
- **Ctrl+Shift+PgUp / Ctrl+Shift+PgDn** --> move current tab left / right

I like this, because I have PgUp & PgDn keys right above my arrow keys, so I can use right Ctrl to easily navigate tabs in _most_ apps with right hand, just moving quickly from a touchpad. The _only_ app I regularly use where this doesn't work is - guess what - the elementary Terminal :confused: (also Code, but I rarely use it for anything).

PR #333, opened 2018-12-03
Add moving current tab shortcuts

Mohelm97:
Adds #322


Mohelm97:
I think it would be better to add it over the DynamicNotebook, But quite not sure, Like why they added the shortcut for travelling between tabs here rather than the DynamicNotebook?

jeremypw:
On reflection, I think there are problems implementing actions within Granite itself so the PR's approach is probably OK.

jeremypw:
@Mohelm97 Would you like to fix the conflict so that this PR can be progressed, thanks :-)

Mohelm97:
@jeremypw Cool, I will do it soon :D

Mohelm97:
@jeremypw It's ready now :D

PR #337, opened 2018-12-04
UnsafePasteDialog: Use GraniteMessageDialog

danrabbit:
Subclass Granite.MessageDialog instead of Gtk.Dialog


danrabbit:
@cassidyjames oops, fixed

PR #349, opened 2018-12-21
Honor privacy setting

### LINKED ISSUE #348: Terminal does not respect privacy setting when saving & restoring tabs
See https://github.com/elementary/code/issues/473

elementary apps should be consistent in this respect.

jeremypw:
Fixes #348 

* Do not save or restore tabs when global history setting is off.

Existing settings are not cleared.

Should not be released until Granite version containing new history_is_enabled () is released


src/Application.vala
```diff
@@ -17,7 +17,8 @@
 * Boston, MA 02110-1301 USA
 */
 
-public class PantheonTerminal.TerminalApp : Gtk.Application {
+namespace PantheonTerminal {
+public class TerminalApp : Gtk.Application {
```
Philip-Scott: Why did you move the Namespace out of the class name? I think we are using the previous method in the rest of the apps :)

src/Settings.vala
```diff
@@ -78,4 +78,12 @@ namespace PantheonTerminal {
             base ("io.elementary.terminal.settings");
         }
     }
+
+    public class PrivacySettings : Granite.Services.Settings {
+        public bool remember_recent_files { get; set; }
+
+        public PrivacySettings () {
+            base ("org.gnome.desktop.privacy");
```
Philip-Scott: This will crash if the setting schema is not installed (Maybe in a different OS). See main review comment :)

Philip-Scott:
I'm also -1 on removing the `remove-tabs` setting.... Seems like it's something that would be useful and not behind the Privacy settings. Maybe keep the same behavior that was before, and use the privacy setting as an override :)

Philip-Scott:
Maybe using this could replace it :) https://github.com/elementary/granite/pull/257

src/Application.vala
jeremypw: Not sure - I guess there was a reason at the time ...

src/Settings.vala
jeremypw: Good point

jeremypw:
@Philip-Scott :  Thanks for the helpful review :-) It would make sense to use the Granite function so I'll block this till that is merged.  I take your point that our own "remember-tabs" setting could still be useful to stop this behaviour even if the privacy setting allowed it.

jeremypw:
Need to wait for latest granite to be used by Travis for this to pass CI.

src/MainWindow.vala
```diff
@@ -1231,18 +1245,20 @@ namespace PantheonTerminal {
...
-                    return;
-                }
-
-                var location = term.get_shell_location ();
-                if (location != null && location != "") {
-                    opened_tabs += location;
-                }
-            });
+            if (Granite.Services.System.history_is_enabled () &&
+//                PantheonTerminal.settings.remember_tabs) {
```
cassidyjames: Left a comment in here?

PR #357, opened 2019-02-04
Set pointer_autohide property to true

### LINKED ISSUE #355: Auto-hide the mouse cursor when is over the window area and a key is pressed.
I'm sorry if this issue is a duplicate.

On Ubuntu terminal and on VLC there is this feature that auto-hides the mouse when the mouse is over the window area and a key is pressed.

What if elementary Terminal had a similar function?

If the mouse cursor was over the window it would auto-hide when a key was pressed.

This works particularly well when the terminal is in fullscreen mode.

cassidyjames:
Fixes #355


Blast-City:
Hi.

Terminal shortcuts seems to make the cursor show again.

PR #369, opened 2019-04-12
Zoom with Ctrl+Equal

### LINKED ISSUE #368: Zoom in on Ctrl+Equal as well
Many applications (including files and third party apps) use `ctrl` + `=` in my experience. This scheme also has the benefit of using the same modifiers for zoom out and zoom in, so it is simpler to toggle the zoom level.

cassidyjames:
~Trying to fix https://github.com/elementary/terminal/issues/368. This also (usually) segfaults, just like https://github.com/elementary/code/pull/661~

Fixes #368

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -165,7 +165,10 @@ namespace PantheonTerminal {
             set_application (app);
 
             foreach (var action in action_accelerators.get_keys ()) {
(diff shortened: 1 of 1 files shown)


cassidyjames:
According to @davidmhewitt:

>`set_accels_for_action` expects a null terminated array and `MultiMap.to_array ()` doesn't give you a null terminated array, so the bigger question is why did it work at all before.
>
>But, every time we use this pattern, we should be appending null to the array.

PR #375, opened 2019-05-19
Use GSettings for Saved State

### LINKED ISSUE #15: Window states are saved inconsistently
Scratch saves its' window state in gsettings keys window-state, window-height, window-width, window-x, window-y. 

* Switchboard uses position instead of window-x, window-y. 
* Terminal uses opening-x, opening-y instead of window-x, window-y. 
* Pantheon Files uses maximized (boolean) instead of window-state. 

I propose using the keys used by Scratch consistently across all apps.

Launchpad Details: [#LP1475825](https://bugs.launchpad.net/bugs/1475825) quassy - 2015-07-18 02:18:55 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44562187-window-states-are-saved-inconsistently?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #377: Fullscreen state is not saved if window was firstly maximized.
1- Open terminal.
2- Maximize the window (Super+Up).
3- Press F11 to make the window fullscreen.
4- Close Terminal and then open it.

**Result:** The Fullscreen state is not saved. This doesn't happen if the window wasn't previously maximized.

Sorry if this a deliberate behaviour.

danrabbit:
While we're here, modernize saved state settings to use fewer keys

Fixes #15 
Fixes #377

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -698,11 +698,11 @@ namespace PantheonTerminal {
                 if (get_window () == null)
                     return false;
 
-                /* Save window state */
-                if (is_maximized) {
-                    PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 1);
-                } else if ((get_window ().get_state () & Gdk.WindowState.FULLSCREEN) != 0) {
+                /* Check for fullscreen first: https://github.com/elementary/terminal/issues/377 */
+                if ((get_window ().get_state () & Gdk.WindowState.FULLSCREEN) != 0) {
                     PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 2);
+                } else if (is_maximized) {
+                    PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 1);
                 } else {
                     PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 0);
 
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -591,7 +591,7 @@ namespace PantheonTerminal {
 
             if (restore_pos) {
                 int window_x, window_y;
-                PantheonTerminal.TerminalApp.saved_state.get ("window-position", "(ii)", out window_x, out window_y);
+                PantheonTerminal.TerminalApp.saved_state.get ("window-position", "(ii)", out rect.x, out rect.y);
 
                 if (window_x != -1 ||  window_y != -1) {
                     move (window_x, window_y);
@@ -791,7 +791,7 @@ namespace PantheonTerminal {
             PantheonTerminal.TerminalApp.saved_state.set_strv ("tabs", {});
 
             int focus = PantheonTerminal.TerminalApp.saved_state.get_int ("focused-tab");
-            focus.clamp (0, tabs.length - 1);
+            focus = focus.clamp (0, tabs.length - 1);
 
             Idle.add_full (GLib.Priority.LOW, () => {
                 focus += notebook.n_tabs;
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -552,28 +552,27 @@ namespace PantheonTerminal {
 
         private void restore_saved_state (bool restore_pos = true) {
             saved_tabs = saved_state.tabs;
-            default_width = PantheonTerminal.saved_state.window_width;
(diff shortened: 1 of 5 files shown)


ryonakano:
I'm not sure whether this should be done in this PR, and this is a correct solution, but checking the window is fullscreen or not first will also fix #377.

```diff
diff --git a/src/MainWindow.vala b/src/MainWindow.vala
index 65f7bbb..e8f8949 100644
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -699,10 +699,10 @@ namespace PantheonTerminal {
                     return false;
 
                 /* Save window state */
-                if (is_maximized) {
-                    PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 1);
-                } else if ((get_window ().get_state () & Gdk.WindowState.FULLSCREEN) != 0) {
+                if ((get_window ().get_state () & Gdk.WindowState.FULLSCREEN) != 0) {
                     PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 2);
+                } else if (is_maximized) {
+                    PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 1);
                 } else {
                     PantheonTerminal.TerminalApp.saved_state.set_enum ("window-state", 0);
 ```

**Comment by danrabbit** [comment]
@ryonakano thanks for the tip! Fixed

File: src/MainWindow.vala [hunk 1]
```diff
@@ -787,9 +788,11 @@ namespace PantheonTerminal {
                 }
             }
 
-            PantheonTerminal.saved_state.tabs = {};
+            PantheonTerminal.TerminalApp.saved_state.set_strv ("tabs", {});
+
+            int focus = PantheonTerminal.TerminalApp.saved_state.get_int ("focused-tab");
+            focus.clamp (0, tabs.length - 1);
```
donadigo: `clamp` returns you a new int here, so currently, you are clamping and then immediately discarding the result.
```suggestion
            focus = focus.clamp (0, tabs.length - 1);
```

src/MainWindow.vala
```diff
@@ -570,35 +573,34 @@ namespace PantheonTerminal {
...
-            if (default_height == -1) {
+                default_width = geometry.width * 2 / 3;
                 default_height = geometry.height * 3 / 4;
             }
 
             if (restore_pos) {
-                int x = saved_state.opening_x;
-                int y = saved_state.opening_y;
+                int window_x, window_y;
+                PantheonTerminal.TerminalApp.saved_state.get ("window-position", "(ii)", out window_x, out window_y);
```
donadigo: You could use the already available `rect` instead of defining new window x and y.
```suggestion
                PantheonTerminal.TerminalApp.saved_state.get ("window-position", "(ii)", out rect.x, out rect.y);
```

src/MainWindow.vala
```diff
@@ -33,6 +32,10 @@ namespace PantheonTerminal {
         private bool is_fullscreen = false;
         private string[] saved_tabs;
 
+        private const int NORMAL = 1;
+        private const int MAXIMIZED = 1;
```
donadigo: It looks like something went wrong here. Shouldn't `NORMAL` be `0`?

src/MainWindow.vala
```diff
@@ -570,35 +573,33 @@ namespace PantheonTerminal {
...
+                if (rect.x != -1 ||  rect.y != -1) {
+                    move (rect.x, rect.y);
                 }
             }
 
-            if (PantheonTerminal.saved_state.window_state == PantheonTerminalWindowState.MAXIMIZED) {
+            if (PantheonTerminal.TerminalApp.saved_state.get_enum ("window-state") == MainWindow.MAXIMIZED) {
                 maximize ();
-            } else if (PantheonTerminal.saved_state.window_state == PantheonTerminalWindowState.FULLSCREEN) {
+            } else if (PantheonTerminal.TerminalApp.saved_state.get_enum ("window-state") == MainWindow.FULLSCREEN) {
```
donadigo: You could get the window state once here by caching it.

Issue #55, opened 2017-04-26
Right click key does not work in terminal


elementaryBot:
The right click key (on the keyboard not the mouse) does not work in a terminal window. It should be equivalent to the mouse right click button.

Launchpad Details: [#LP1557932](https://bugs.launchpad.net/bugs/1557932) Bob Meyers - 2016-03-16 07:47:54 +0000

cassidyjames:
I can confirm the [Menu key](https://help.gnome.org/users/gnome-help/stable/keyboard-key-menu.html.en) does not show the menu as expected.

cassidyjames:
The key does work in GNOME Terminal.

ryonakano:
Seem like this is fixed in #382 but remains open, so closing.

PR #382, opened 2019-06-25
Pressing the menu keyboard button opens right-click menu

jeremyvaartjes:
Fix for issue #55

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -485,7 +485,12 @@ namespace PantheonTerminal {
                     case Gdk.Key.Down:
                         current_terminal.remember_command_start_position ();
                         break;
-
+                    case Gdk.Key.Menu:
+                        if (!search_toolbar.search_entry.has_focus && !style_popover.visible) {
+                            menu.popup_at_widget(current_terminal, Gdk.Gravity.NORTH_EAST,
+                                            Gdk.Gravity.NORTH_EAST, e);
+                        }
+                        break;
                     default:
                         if ((e.state & Gtk.accelerator_get_default_mod_mask ()) == 0) {
                             current_terminal.remember_command_start_position ();
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -486,10 +486,23 @@ namespace PantheonTerminal {
                         current_terminal.remember_command_start_position ();
                         break;
                     case Gdk.Key.Menu:
-                        if (!search_toolbar.search_entry.has_focus && !style_popover.visible) {
-                            menu.popup_at_widget(current_terminal, Gdk.Gravity.NORTH_EAST,
-                                            Gdk.Gravity.NORTH_EAST, e);
-                        }
+                        /* Popup context menu below cursor position */
+                        long col, row;
+                        current_terminal.get_cursor_position (out col, out row);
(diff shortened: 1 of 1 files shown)


jeremypw:
This works as expected from the code - thanks.

There is a question over where the menu should pop up;

* At one corner (as in this PR)
* At the pointer (as in e.g. Mate Terminal or Files)
* At the cursor (as in Code)
* Somewhere else

Popping up at the cursor seems most logical for a text based app, although a little harder to achieve.

I'll flag this for comment from @elementary/ux

cassidyjames:
I agree text-cursor seems the most ideal, but I'm also not that opposed to on the mouse pointer or corner of the app.

jeremypw:
Further investigation indicates that this is only a problem if you popup the meny at scale_factor = 1 and then change the scale_factor to 2 using the interactive debugger without restarting the program.  So it may not be a serious problem in real life.

jeremypw:
Here is a way to pop the menu up at the cursor:
```
    /* Popup context menu below cursor position */
    long col, row;
    current_terminal.get_cursor_position (out col, out row);
    var cell_width = current_terminal.get_char_width ();
    var cell_height = current_terminal.get_char_height ();
    var rect_window = current_terminal.get_window ();

    Gdk.Rectangle rect = {(int)(col * cell_width),
                          (int)(row * cell_height),
                          (int)cell_width,
                          (int)cell_height};

    menu.popup_at_rect (rect_window, rect, Gdk.Gravity.SOUTH_WEST, Gdk.Gravity.NORTH_WEST, e);
```
```

**Comment by jeremyvaartjes** [comment]
I made the change you have suggested. I was originally worried that putting it on the terminal carrot would make the menu not work when you scroll the window, but it just causes the terminal to scroll to the carrot which is good. One issue I did pick up is that the positioning stops working properly when the "clear" command is used, the menu is just placed at the bottom of the window. Probably not a big issue though since it is still tied to the window.

**Comment by jeremypw** [comment]
The positioning is also off if the Menu key is pressed while a child process is running (e.g. man).  I am not sure whether the context menu should be shown at all in this situation (although right-click still shows it).

**Comment by jeremypw** [comment]
I think the problem with the `clear` command may be an upstream issue with vte-2.91 library.  The row number returned by `get_cursor_position ()` is not updated even thought the cursor clearly moves.  Even after typing some more characters the result is not updated.  If this is the case, then the best we can do is make sure the menu is not displayed outside the terminal window.

**Comment by jeremypw** [comment]
You can fix the `clear` issue like this:
```
        var cell_width = current_terminal.get_char_width ();
        var cell_height = current_terminal.get_char_height ();
        var rect_window = current_terminal.get_window ();
        var vadj_val = current_terminal.get_vadjustment ().get_value ();

        Gdk.Rectangle rect = {(int)(col * cell_width),
                              (int)((row - vadj_val) * cell_height),
                              (int)cell_width,
                              (int)cell_height};

        menu.popup_at_rect (rect_window,
                            rect,
                            Gdk.Gravity.SOUTH_WEST,
                            Gdk.Gravity.NORTH_WEST,
                            e);
```

jeremypw:
This also fixes the child process issue except that if the cursor is at the bottom of the screen then the menu does not flip or slide to keep within the window.  However that is also an issue in master so need not be fixed here.

PR #384, opened 2019-07-01
Change PantheonTerminal.TerminalApp to Terminal.Application

danrabbit:
Shorter and saner namespacing

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -17,7 +17,7 @@
 * Boston, MA 02110-1301 USA
 */
 
-public class PantheonTerminal.TerminalApp : Gtk.Application {
+public class Terminal.Application : Gtk.Application {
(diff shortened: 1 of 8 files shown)


src/Application.vala
```diff
@@ -44,7 +44,7 @@ public class PantheonTerminal.TerminalApp : Gtk.Application {
         Intl.setlocale (LocaleCategory.ALL, "");
     }
 
-    public TerminalApp () {
+    public Application () {
         Granite.Services.Logger.initialize ("PantheonTerminal");
```
jlnr: Should this be "Terminal" now? It looks like this arg is silently discarded anyway: https://github.com/elementary/granite/blob/07ff49fabb2d3617d2717b751ae5eaded60add0f/lib/Services/Logger.vala#L95

PR #386, opened 2019-07-05
Move cursor to clicked position (as near as possible)

### LINKED ISSUE #383: Key + mouse click to position cursor?
What if you could press <kbd>Ctrl</kbd> (or other modifier) and click to move the cursor in the terminal to under the mouse (within the current line).
I've recently heard about that this is possible on the OS X Terminal: https://medium.com/sweetmeat/point-and-click-your-cursor-to-any-spot-in-a-terminal-line-on-os-x-452e9cd533ff

jeremypw:
Fixes #383 

Possible solution to linked issue.  Seems a bit hacky - suggestions welcome.


andreasfelix:
Hi, why not use the `button_press_event`? I think this would be more in line standard editors.

src/Widgets/TerminalWidget.vala
```diff
@@ -149,13 +149,42 @@ namespace PantheonTerminal {
...
+
+                            /* Synthesise a cursor press - is there a better way? */
+                            Gdk.EventKey key_event = (Gdk.EventKey)(new Gdk.Event (Gdk.EventType.KEY_PRESS));
+                            key_event.send_event = 1;
+                            key_event.window = (Gdk.Window)(this.get_window ().ref ()); /* Need to add a ref else crash on second key press - vapi error? */
+                            key_event.keyval = (p_row < c_row || (p_row == c_row && p_col < c_col)) ? Gdk.Key.Left : Gdk.Key.Right;
+                            key_event.is_modifier = 0;
+
+                            int last_row = c_row;
+                            int last_col = c_col;
```
andreasfelix: both variables are unused

jeremypw:
@andreasfelix Thanks for the review!  I think I used `button-release-event` because I was concerned about interfering with drag-select or drag-n-drop.  Would it solve any problem to use `button-press-event`?  Your other points are good ones - thanks - will address.

andreasfelix:
> Would it solve any problem to use button-press-event?

No, I just think it is the standard way to move the cursor.

jeremypw:
Not sure why CI is failing for branch - the tests are timing out but this has happened sometimes in master too.

jeremypw:
Seems terminal is not a drag source anyway so I have moved the logic to button-press-event. I have not found an issue yet.

andreasfelix:
- What is the rational behind `!get_has_selection ()` in `else if (event.button == Gdk.BUTTON_PRIMARY && !get_has_selection () && !has_foreground_process ())`? This makes it necessary to click twice to move the cursor if text is selected.
- Selecting text above the current line now moves the cursor to the first possible position. This could be ruled out with:
  ```vala
  if (p_row < remembered_command_start_row) {
      return false;
  }
  ```
- Your commit 2dfd129  which turns off the blinking does not fix **3.** anymore.

  **3.** could be fixed by setting the cursor color to the background color before the first `key_event` and set it back to the normal color after the last `key_event`:

  ```vala
  set_color_cursor (background_color);
  Idle.add (() => { /* wait for button press event to be processed */
      /* Cursor will move as close as possible to pointer */
      for (int i = 0; i < n_events; i++) {
          key_event.time = (uint32)(get_monotonic_time ());
          key_press_event (key_event);
      }

      Timeout.add (100, () => {
          set_color_cursor (cursor_color);
          return false;
      });

      return false;
  });
  ```
  There is probably a better solution than using `Timeout`, but one has to make sure that the `key_events` are already processed!

jeremypw:
I tried making the cursor invisible by setting it to background color before but it causes a similar issue when you click just after the cursor blinks on - it immediately disappears again.  Maybe slightly less jarring but not ideal.  Turning blinking off seemed the better solution but none I tried were ideal.

Also if the timeout expires just before the cursor is due to blink off then again the cursor will flash on and off.  I am not sure this issue is worth the extra code complexity tbh.

jeremypw:
I tried the color change method again and it works better your way - I didn't have the timeout when I did it before.  So I have now used it.  I have also taken on board your other suggestions - thanks!

andreasfelix:
I would have some other minor suggestions. I have already implemented them in https://github.com/andreasfelix/terminal/commit/6fc6025d72b6f073dbd982b036f6bd9c97c2d23b. Feel free to merge them if you agree.

- The last commit makes it not possible to move the cursor for multi line commands. This could be solved by only moving relative to column if the clicked_row is below the last possible row.
- I think it would be better if `get_cell_clicked` would also return `long` types. Otherwise the values have to be casted back and forth if one passes them to another function of `Vte.Terminal`
- I would keep "Turn off blink till the button_release", as this is also common for text editors. (Check the markdown editor of Github.)
- I would rename `c_row/c_col` to `current_row/current_col` and `p_row/p_col` to `clicked_row/clicked_col`
- I would move
  ```vala
  Gdk.RGBA background_color;	
  Gdk.RGBA cursor_color;
  ```
  to the top and initialize them at declaration.
- I would check `event.button == Gdk.BUTTON_PRIMARY` before `event.button ==  Gdk.BUTTON_SECONDARY`

jeremypw:
Sounds good to me - I have merged your changes.

jeremypw:
Somtimes the blink gets permanently turned off - not sure why - investigating.

andreasfelix:
- Would maybe better to use 
  ```vala
  Gdk.threads_add_idle_full (Priority.LOW, () => {...});
  ```

  instead of

  ```vala
  Timeout.add (100, () => {...});
  ```
  as this would not require the "hard-coded" 100ms value?

- Also when clicking the "white space" the cursor should not only move relative to the current col but also move to the last possible row. Otherwise this could lead to unwanted behavior:

  move_relative

  This could be implemented by synthesizing a `Key.End` press and then moving the cursor relative 
  within the last line. As this is more complicated and would require an additional await 
  of the `Key.End` press event, maybe it is easier/better to just ignore clicks on the "white space" with:

  ```vala
  if (clicked_symbol == 0U) {
    return false;
  }
  ```

jeremypw:
I tried synthesising Gdk.Key.End and it works on the first click but there does not seem any (simple, elegant) way of keeping track of the end of the commandline so there were problems on subsequent clicks.  So I have ignored clicks on space as advised.  Also replaced hard-coded Timeout - thanks.

src/Widgets/TerminalWidget.vala
```diff
@@ -149,6 +149,8 @@ namespace Terminal {
                                                          null, null).get_char ();
                     if (clicked_symbol != 0U) {
                         delta_cells += (clicked_row - current_row) * get_column_count ();
+                    } else {
```
andreasfelix: Could be simplified to:

```vala                                                    
if (clicked_symbol == 0U) {
    return Gdk.EVENT_STOP;
}

delta_cells = clicked_col - current_col + (clicked_row - current_row) * get_column_count ();
```

jeremypw: OK, done.

andreasfelix: I think there is no reason to increment. You could simplify further:
‘‘‘
long delta_cells = clicked_col - current_col + (clicked_row - current_row) * get_column_count ();
‘‘‘

jeremypw:
After further testing I had to revert trying to be clever when clicking on empty space. It causes problems if the command is edited in the middle and then clicked at the end to continue for example - the cursor cannot be returned to the end (by clicking).  I think it  would be better to accept the original "brute force" approach where the cursor moves as far as possible if clicked outside the command as the lesser of two evils.  A refinement can always be submitted later if found.

src/Widgets/TerminalWidget.vala
```diff
@@ -129,7 +132,49 @@ namespace Terminal {
...
-                if (event.button ==  Gdk.BUTTON_SECONDARY) {
+                remembered_cursor_blink_mode = cursor_blink_mode;
+                cursor_blink_mode = Vte.CursorBlinkMode.OFF;
+
+                if (event.button == Gdk.BUTTON_PRIMARY && !has_foreground_process ()) {
+                    long current_col, current_row, clicked_row, clicked_col;
+                    get_cursor_position (out current_col, out current_row);
+                    get_cell_clicked (event, out clicked_row, out clicked_col);
+                    if (clicked_row < remembered_command_start_row) {
+                        return Gdk.EVENT_STOP;
```
andreasfelix: I think it should be `Gdk.EVENT_PROPAGATE`

src/Widgets/TerminalWidget.vala
```diff
@@ -139,13 +184,15 @@ namespace Terminal {
                     menu.select_first (false);
                     menu.popup_at_pointer (event);
 
-                    return true;
+                    return Gdk.EVENT_PROPAGATE;
                 }
 
-                return false;
+                return Gdk.EVENT_STOP;
```
andreasfelix: I think it should be `Gdk.EVENT_PROPAGATE`

src/Widgets/TerminalWidget.vala
```diff
@@ -139,13 +184,15 @@ namespace Terminal {
                     menu.select_first (false);
                     menu.popup_at_pointer (event);
 
-                    return true;
+                    return Gdk.EVENT_PROPAGATE;
```
andreasfelix: I think it should be `Gdk.EVENT_STOP`

andreasfelix:
I also noticed that the condition `clicked_row < remembered_command_start_row` is not sufficient to rule out clicks on old output. This is because the `remembered_command_start_row` contains the position of the row in regard to all rows, while the `clicked_row` value is in regard to all visible rows.

jeremypw:
Not sure if its in the guidance yet, but dan and others now recommend using the named boolean constants to end event and source closures for readability/clarity. There is no technical advantage as far as I know,

jeremypw:
I was aware of possible problems caused by scrolling/resizing but wanted to get a basic solution working first.  I am beginning to think that the simplest solution to making terminal behave more like a text view is to have a hidden text view behind it and process all the text there!

jeremypw:
I think I have found a simpler solution for the scrolling/resizing problem.

src/Widgets/TerminalWidget.vala
```diff
@@ -489,6 +530,11 @@ namespace Terminal {
             return get_text_range (start_row, 0, output_end_row - 1, 1000, null, null) + "\n";
         }
 
+        private void get_cell_clicked (Gdk.EventButton event, out long row, out long col) {
+            row = (long)(event.y / get_char_height ());
```
andreasfelix: Just a minor suggestion: As the functions of the `Vte.Terminal` also use "absolute" row-values, would it maybe better to return "absolute" values in the `get_cell_clicked_function`?
```vala
row = (long)(event.y / get_char_height () + get_vadjustment ().get_value ());
```
Or is there a reason to prefer "relative" values?

Or will there

jeremypw: OK, I have now used absolute values consistently.

jeremypw:
Note:  This does not currently work when the commandline is a url as the code launching that url takes priority.  I am not minded to address that in the PR though.

src/MainWindow.vala
```diff
@@ -463,9 +463,8 @@ namespace Terminal {
                             return true;
                         } else if (!current_terminal.has_foreground_process ()) {
                             /* Ignore returns being sent to a foreground process */
-                            current_terminal.remember_position ();
-                            get_simple_action (ACTION_SCROLL_TO_LAST_COMMAND).set_enabled (true);
                             current_terminal.remember_command_end_position ();
```
andreasfelix: Remember cursor_command_start_ position when pressing enter.

```vala
Timeout.add (100, () => {
    current_terminal.remember_command_start_position ();
    return false;
})
```

andreasfelix: Logic breaks down if command needs more then 100ms seconds to return. `remembered_command_start_row` should be remembered as soon as command returns.

andreasfelix:
Currently `remembered_command_start_row` is only remembered after the next key press event. This leads to the problem that clicks on output, which should be ignored by 

```vala
if (clicked_row < remembered_command_start_row) {
  return Gdk.EVENT_PROPAGATE;
}
```
are registered, if in between the last <kbd>return</kbd> press and following *mouse-click* was no other key press.


One solution would be to remember the `remembered_command_start_row` as soon as the command returns. One naive approach to use a timeout when the <kbd>Return</kbd> was pressed. (See code comment) Unfortunatly this logic breaks down if the command needs longer than 100ms to return. A better solution is needed!

src/Widgets/TerminalWidget.vala
```diff
@@ -128,7 +131,45 @@ namespace Terminal {
...
             button_press_event.connect ((event) => {
-                if (event.button == Gdk.BUTTON_SECONDARY) {
+                remembered_cursor_blink_mode = cursor_blink_mode;
+                cursor_blink_mode = Vte.CursorBlinkMode.OFF;
+
+                if (event.button == Gdk.BUTTON_PRIMARY && !has_foreground_process ()) {
+                    long current_col, current_row, clicked_row, clicked_col;
+                    get_cursor_position (out current_col, out current_row);
+                    get_clicked_cell_position (event, out clicked_row, out clicked_col);
+
```
andreasfelix: just add 
```vala
```
here to fix the scrolling problem

src/MainWindow.vala
andreasfelix: Nevermind I found a better solution see comment below.

andreasfelix:
As you now update remembered positions when the window is resized, shouldn't it be save to remove the `!resized &&`?

```vala
public bool has_output () {
  return !resized && get_last_output ().length > 0;
}
```

src/Widgets/TerminalWidget.vala
```diff
@@ -138,13 +179,20 @@ namespace Terminal {
...
 
-                    return true;
+                    return Gdk.EVENT_STOP;
                 }
 
-                return false;
+                return Gdk.EVENT_PROPAGATE;
+            });
+
+            key_release_event.connect (() => {
```
andreasfelix: Wouldn't it be better to use the `cursor_moved` signal here? This should cover all cases.
```vala
cursor_moved.connect (remember_position ());
```

jeremypw: I don't think so.  We only want to remember the cursor position after the last key press, not when the cursor is moved for other reasons (e.g. due to output from a process).

jeremypw:
@andreasfelix Any chance of another look at this? I have incorporated one of your suggestions.  The one about `resized` I have not included, as this PR does not currently touch that code and as far as I can tell it does not fix any issue caused by this PR.  If that is not the case please comment further.

andreasfelix:
If I remember correctly, the reason I suggest to remove the `resized` check is, because currently there is an issue that the last output cannot reliably copied if the window was resized. To my understanding this was fixed by your commit 5337b6b, which should make this check unnecessary.

Another thing I noticed: Clicking on the prompt string should not move cursor to the beginning of the command. I often double-click the prompt string to copy the current working directory and this would unintentionally move the cursor to the beginning.

I did not test this, but should be solved by checking for the current column if  `clicked_row` equals ` remembered_command_start_row`:
```vala
if (
    clicked_row < remembered_command_start_row
    || (clicked_row == remembered_command_start_row 
    && clicked_col < remembered_command_start_col)
) {
    return Gdk.EVENT_PROPAGATE;
}
```

Also, I think it should also be made sure that this feature works correctly for different prompt strings (the PS1 variable). I, for example, have a line break after the current working directory to prevent the line wrapping of longer commands. I made a simple graphic to show what I mean:

terminal svg

It might be a good idea to make this feature toggle-able and of be default, so more people can test it...

I will use the branch this branch for 1-2 and then try to give some more feedback.

andreasfelix:
Just an idea: I found this [page](http://www.tldp.org/HOWTO/Bash-Prompt-HOWTO/x361.html) where it says, that one also can use these escape sequences to move the cursor. 

```vala
var n_events = (int) delta_cells.abs ();
for (int i = 0; i < n_events; i++) {
    feed_child (delta_cells > 0U ? "\033[C" : "\033[D", 3);
}
```

I am not sure if this is actually less hacky than emulating a keypress. At least it is some kind of an interface ...

andreasfelix:
I noticed a logical error in `remember_command_start_position`. The position is only remembered if the last key press was the <kbd>return key</kbd>. But there are other possiblities to advance the prompt: For example when triggering the tab-completition by pressing the <kbd>Tab</kbd> key twice or when pressing <kbd>Ctrl+C</kbd> or <kbd>Ctrl+D</kbd>.

This leads to the problem, that the `remembered_start_row` is not always up-to-date and therefore not all mouse clicks above the current command prompt are ignored. A more reliable way to save the command_start_col/row is needed ...

andreasfelix:
@jeremypw why did you close this?

jeremypw:
Spring-cleaning old PRs with no recent progress/interest from reviewers.

PR #387, opened 2019-07-08
MainWindow: Add accels to menuitem labels

danrabbit:
Screenshot from 2019-07-08 08 32 19@2x

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -192,23 +192,29 @@ namespace PantheonTerminal {
 
             primary_selection = Gtk.Clipboard.get (Gdk.Atom.intern ("PRIMARY", false));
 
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -1317,5 +1323,34 @@ namespace PantheonTerminal {
         public GLib.SimpleAction? get_simple_action (string action) {
             return actions.lookup_action (action) as GLib.SimpleAction;
         }
+
+        private class AccelMenuLabel : Gtk.Grid {
```
jeremypw: I wonder how much of this class could be abstracted into Granite?  However, that is for the future.

danrabbit: I have a PR for that :) https://github.com/elementary/granite/pull/300

jeremypw:
I see there is a PR for a Granite accelerator label ...

PR #390, opened 2019-07-24
Use palette colors for default bg/fg

### LINKED ISSUE #373: Dark style is low constrast
The default color scheme for the terminal is low contrast. I'm not sure if it's the somewhat small font size or the dull background color, but some combo of factors is making things a bit more difficult to read in the terminal than other apps. It is most evident in sunlight.

Here I put elementary and gnome terminal apps side-by-side to make the difference in contrast clear. 

Screenshot from 2019-05-14 19-32-38

cassidyjames:
Better matches https://github.com/elementary/stylesheet/pull/510, plus better contrast.

## Before

image

image

## After

image

image
 
Fixes #373

### DIFF
--- a/data/io.elementary.terminal.gschema.xml
+++ b/data/io.elementary.terminal.gschema.xml
@@ -91,7 +91,7 @@
     </key>
 
     <key name="foreground" type="s">
-      <default>"#94a3a5"</default>
+      <default>"#d4d4d4"</default>
       <summary>Color of the text.</summary>
       <description>
           The color of the text of the terminal.
@@ -103,7 +103,7 @@
       </description>
     </key>
     <key name="background" type="s">
-      <default>"rgba(37, 46, 50, 0.95)"</default>
+      <default>"rgba(51, 51, 51, 0.95)"</default>
       <summary>Color of the background.</summary>
       <description>
           The color of the background of the terminal.
--- a/data/Application.css
+++ b/data/Application.css
@@ -47,8 +47,8 @@
 }
 
 .color-dark radio {
-    background: #252E32;
-    border-color: #151B1C;
+    background: mix(@BLACK_500, @BLACK_300, 0.5);
+    border-color: @BLACK_700;
     color: #fff;
 }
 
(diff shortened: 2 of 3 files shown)


fabiozaramella:
This makes the text much more readable.
Has a decision been reached about this?

cassidyjames:
Now that the neutral GTK style has been shipping for a while, it'd be great to get the Terminal to match. Right now it's off and it feels weird imho.

data/io.elementary.terminal.gschema.xml
```diff
@@ -91,7 +91,7 @@
     </key>
 
     <key name="foreground" type="s">
-      <default>"#94a3a5"</default>
```
jlnr: This color is still being referenced in `io.elementary.terminal.gschema.xml`, as part of "palette", and in the method `TerminalWidget.restore_settings`. Should these be replaced as well, or should the palette be left untouched by this PR? (Not sure why there is both a palette, and default fg/bg colors.)

data/io.elementary.terminal.gschema.xml
```diff
@@ -103,7 +103,7 @@
       </description>
     </key>
     <key name="background" type="s">
-      <default>"rgba(37, 46, 50, 0.95)"</default>
```
jlnr: The `.color-dark radio` in Application.css still uses this color (with 100% opacity), should probably be changed to the gray color as well?

cassidyjames: @jlnr good catch, addressed.

data/io.elementary.terminal.gschema.xml
cassidyjames: @jlnr hm, I'm not sure. I believe the palette is used by the VTE itself for things like curses. I'm not sure if it should be updated.

jlnr: Fair enough. If it looks odd in real-world use cases, we can open another PR to fix things.

cassidyjames:
> LGTM, does this need more UX discussion?

Since we've merged the stylesheet changes and this makes Terminal match, I think we're good. :smile:

PR #391, opened 2019-07-28
Use Solarized Dark for dark UI

cassidyjames:
As an alternative to #390 if we wanted Terminal to be more explicitly Solarized Dark, like the Code syntax highlighting. Going for something like this:

image (7)

For some reason loading the stylesheet isn't working yet.

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -47,8 +47,8 @@
 }
 
 .color-dark radio {
-    background: #252E32;
-    border-color: #151B1C;
+    background: #002b36; /* Solarized Dark Base03 */
+    border-color: #000;
     color: #fff;
 }
 
--- a/data/dark.css
+++ b/data/dark.css
@@ -1,0 +1,20 @@
+/*
+* Copyright (c) 2019 elementary, Inc. (https://elementary.io)
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License version 3, as published by the Free Software Foundation.
+*
+* This program is distributed in the hope that it will be useful,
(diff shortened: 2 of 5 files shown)


fabiozaramella:
I don't know...in my opinion the neutral dark style of your previous branch is better for the eyes, and not only it is coherent with dark style across the OS but it also better matches Terminal's icon.
This one looks too blue-greenish.

cassidyjames:
@fabiozaramella we use Solarized Dark in Code for syntax highlighting, and we were chatting about aligning the developer tools to use solarized for their dark styles. The neutral dark is a good default for the stylesheet since it affect content-creation and editing apps like Photos and video editors, but that's less of a big deal with developer tools.

I imagine we'd also update the Terminal icon to match, like we did with the Code icon and Solarized Light.

fabiozaramella:
@cassidyjames Well, if Solarized (both light and dark) is going to be the default style for developer tools I'm perfectly ok with this.  

My only concern is coherence across different apps. For example, this will require to update not only Terminal icon but also dark UI for Code (at the moment only the text view uses the solarized dark color scheme) and... if Terminal and Code then get a Solarized dark UI their light counterparts, with default grey decorations, will look odd, wouldn't they?

cassidyjames:
@fabiozaramella It’s _almost_ like this is a draft PR to experiment with things! 😜️ Yes, there are a lot of implications and we’ll need to decide what the scope is. I’m just pushing this up here in a draft PR instead of keeping it to myself on my machine.

danrabbit:
This way too blue, I don't like it

cassidyjames:
Closing as it was an experiment but I don't see it getting merged

PR #397, opened 2019-08-23
Implement gnome-terminal style -x (--) flag.

### LINKED ISSUE #395: Can't use options with io.elementary.terminal
For example : 
` io.elementary.terminal -h `
 produces the following output 
>  [DEBUG 21:57:54.738715] Settings.vala:163: Loading settings from schema 'io.elementary.terminal.settings' Usage:   io.elementary.terminal [OPTION…]  

Why do I need this to work?  

I have to change "Terminal to launch console program" from ***xterm -T $TITLE -e*** to elementary's OS (Juno) ***io.elementary.terminal -e*** in CodeBlocks.
How do I fix this?

jeremypw:
Fixes #395 (hopefully)

Implements a flag (-x or --) that causes the remainder of the line to be treated as a command in its own right, to be passed to a new terminal, without requiring quotes.  This is intended to match the behaviour of gnome-terminal.  It allows more complex commands to be executed (including their own flags).  The existing `-e` flag only allows a single argument per command so complex commands need quoting.

It may be advisable to define strings that cause the command building to end - advice required.


src/Application.vala
```diff
@@ -135,9 +134,29 @@ public class Terminal.Application : Gtk.Application {
...
         string[] args = command_line.get_arguments ();
+        string commandline = "";
+        string[] arg_opt = {};
+        string[] arg_cmd = {};
+        bool build_cmdline = false;
+
+        /* Everything after "--" or "-x is to be  treated as a single command to be executed (maybe with its own options)
+         * so it is not passed to the parser. It will be passed as is to a new tab/shell. */
+        foreach (string s in args) {
+            if (s == "--" || s == "-x") {
```
Philip-Scott: You should check here that `build_cmdline` is also false, as it currently is a second -x here would not get parsed as a new tab/shell arg :)

src/Application.vala
```diff
@@ -209,7 +239,8 @@ public class Terminal.Application : Gtk.Application {
     }
 
     private const OptionEntry[] entries = {
-        { "execute", 'e', 0, OptionArg.STRING_ARRAY, ref command_e, N_("Run a program in terminal"), "COMMAND" },
+        /* -e flag is used for running single string commands. May be more than one -e flag in cmdline */
+        { "execute", 'e', 0, OptionArg.STRING_ARRAY, ref command_e, N_("Run a program in terminal"), "PROGRAM_NAME" },
```
Philip-Scott: You should add a description of `-x, --` here as well, right? :)

src/Application.vala
jeremypw: Well spotted - thanks!  Now fixed.

src/Application.vala
jeremypw: Done

Philip-Scott: I actually meant to add another entry for -x, that way the new argument will show up when you type `--help` :)

jeremypw: Oh, I see. This causes some complications as you then have to support a long form option (I note that gnome-terminal does not document the -x or -- flag in the --help output).  However, I have implemented this now.

jeremypw:
@Philip-Scott OK, thanks for the review - I'll not merge for now.

jeremypw:
@Philip-Scott Any chance of a re-review? :-)

src/Application.vala
```diff
@@ -129,9 +129,33 @@ public class Terminal.Application : Gtk.Application {
...
 
         string[] args = command_line.get_arguments ();
+        string commandline = "";
+        string[] arg_opt = {};
+        string[] arg_cmd = {};
+        bool build_cmdline = false;
+
+        /* Everything after "--" or "-x or "--commandline=" is to be  treated as a single command to be executed (maybe with its own options)
+         * so it is not passed to the parser. It will be passed as is to a new tab/shell. */
+        foreach (string s in args) {
```
tintou:
```suggestion
        foreach (unowned string s in args) {
```

src/Application.vala
```diff
@@ -140,19 +164,27 @@ public class Terminal.Application : Gtk.Application {
...
-            start_terminal_with_working_directory (working_directory);
-
         } else {
-            new_window ();
+            if (command_e != null) {
+                run_commands (command_e, working_directory);
+            } else if (commandline.length > 0) {
+                run_command_line (commandline, working_directory);
+            } else if (command_x != null) {
+                string warning = "Usage: --commandline=[COMMANDLINE] without spaces around '='\n\n";
```
tintou:
```suggestion
                const string WARNING = "Usage: --commandline=[COMMANDLINE] without spaces around '='\n\n";
```

PR #401, opened 2019-09-03
Remove unnecessary handling of middle-click.

jeremypw:
There seems to be no reason for terminal to handle middle-click if it is intended that the mouse-touchpad setting plug should be honored.

After removing this code, middle-click will paste the current selection (either inside Terminal or in another app eg.g Code) only if the mouse-touchpad setting "Middle Click Paste" is `true`.  This then matches the behaviour of Code.

Note that this depends on xsettings override being supported which is true for elementary but may not be true for another OS.  If this is a problem, it may be desirable to fallback to middle-click-paste?


cassidyjames:
I do think it's desirable to let the OS handle this instead of handling it inside apps manually. :+1:

PR #403, opened 2019-09-09
Add solarized black and pink themes

redian23:
Better matches elementary/stylesheet#510, plus better contrast and add button on "Style"

## Before 
Only 3 color buttons
Screenshot from 2019-09-09 21-55-58

## After
For Solirezed Dark & Black themes added new pelette "Dracula" for more readable text. 
Screenshot from 2019-09-11 19-48-54
Screenshot from 2019-09-11 20-59-04
Screenshot from 2019-09-10 20-11-02

Thank you @cassidyjames for colors  
https://github.com/elementary/terminal/pull/390#issue-300914485


cassidyjames:
Rather than adding several new color schemes, I think we just want to update the default to match the new UI color, as in https://github.com/elementary/terminal/pull/390

redian23:
> Rather than adding several new color schemes, I think we just want to update the default to match the new UI color, as in #390
You are can add my variant palette or add multicolor settings?. 
Just many people want to new themes.

PR #404, opened 2019-09-10
Ignore hyperlink clicks just after focus in event.

### LINKED ISSUE #104: Focusing the window opens URL links
Terminal recognizes URLs and turns them into clickable links. Add a setting to disable this feature. I often click on a terminal window to give it focus and inadvertently open a link.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46221406-focusing-the-window-opens-url-links?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #104 

Clicks on hyperlinks are ignored for a short period after focusing in so that button clicks that caused the focus in do not result in inadvertant launching of URLs.  This solution does not change the existing UX.  An alternative (maybe safer) would be to require a keyboard modifier as well as a click to launch URLs.


jeremypw:
@cassidyjames It was only working on fast clicks - if button release occurred more than 20ms after press then the hyperlink was followed.  I have pushed an improved solution that works on any length click and also works if the focus is changed using the keyboard (e.g. `<Alt><Tab>`)

PR #414, opened 2019-10-13
Added New Color Scheme Options

TheMartes:
Hey! So after a few weeks of using the pantheon terminal as my full-time terminal, I decided to add new themes into it.

**Reason?**
Well, I'm using Vim quite a lot, and I'm using a theme called Dracula. The problem is, that when I'm running tmux alongside the themes of the terminal, tmux, and vim doesn't match. I've also added a Paper Color Light/Dark as another 2 options because those are also widely popular Vim themes.

*This is just a temporary solution, but a nice addition to Pantheon Terminal. I will also start working on a new feature, where a user should be able to install other themes, which will have support for Base16 colors, Themes Could be imported from `.config` files etc. Also, the font size and font type will be configurable. The motivation behind it is to make pantheon terminal more customizable, and personal.*

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -66,3 +66,18 @@ vte-terminal {
     background: #fff;
     color: #000;
 }
+
+.color-dracula radio {
+    background: #2C2E38;
+    color: #fff;
+}
+
+.color-papercolor radio {
+    background: #EEEEEE;
+    color: #3E3F47;
+}
+
+.color-papercolor-dark radio {
+    background: #121212;
+    color: #C6C6C6;
+}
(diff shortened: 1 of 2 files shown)


jeremypw:
@TheMartes  Best discuss this with the design team before spending too much time on coding ....

danrabbit:
Hey thanks for this PR, but I'm going to close it in favor of the discussion occuring on https://github.com/elementary/terminal/issues/418

I think we probably don't want to expand the default selection of color schemes and instead would prefer to design a nice color scheme editor

PR #415, opened 2019-10-15
Granite.Services.Settings → GLib.Settings

### LINKED ISSUE #413: Replace Granite.Services.Settings with GLib.Settings

danrabbit:
Fixes #413


src/MainWindow.vala
```diff
@@ -701,10 +701,11 @@ namespace Terminal {
         }
 
         private void on_new_tab_requested () {
-            if (settings.follow_last_tab)
+            if (Application.settings.get_boolean ("follow_last_tab")) {
```
jeremypw: Should be "follow-last-tab" (otherwise crashes on adding new tab)

PR #416, opened 2019-10-15
Application: Remove new_window_with_coords

danrabbit:
My guess is that someone put this here because they didn't know about `GLib.Application.get_default ()`

It's never used anywhere else


src/MainWindow.vala
```diff
@@ -684,7 +684,14 @@ namespace Terminal {
 
         private void on_tab_moved (Granite.Widgets.Tab tab, int x, int y) {
             Idle.add (() => {
-                var new_window = app.new_window_with_coords (x, y, false);
+                var new_window = new MainWindow.with_coords (
+                    (Terminal.Application) GLib.Application.get_default (),
```
Exalm: Can't you still use `app` here though?

danrabbit: No because it's another class and `app` is an instance member

danrabbit: Oops nevermind, I'm dumb. I was looking at something else. Fixed

PR #417, opened 2019-10-15
Add 5 new themes and 3 text palette 

redian23:
# Last update 
## Preview

Screenshot from 2019-10-15 17-32-10
Screenshot from 2019-10-15 17-32-15

# Preview themes
#### High Contrast
High Contrast
#### White
White
### Solized dark
Solized dark
### Solized black
Pink
### Pink
Silver
### Monokai Dark
Monokai Dark
### Dracula
Dracula


stsdc:
The idea of theming is good. I think that Terminal should have the ability to apply user defined styles. But fresh, after install Terminal is OK with 3 initial themes.

redian23:
"But fresh, after install Terminal is OK with 3 initial themes."
What do you mean??

stsdc:
@redian23 I just think that Terminal doesn't need more themes.

ryonakano:
> I just think that Terminal doesn't need more themes. But needs mechanism that allows to add custom themes e.g. using dotfiles

We've already had a ticket for that, see #418.

Issue #418, opened 2019-10-15
User-provided color scheme


cassidyjames:
We've been getting a handful of PRs to add one-off color schemes, but these are highly subjective and personal. Instead, we've talked about adding a way for a user to set their own color scheme.

danrabbit:
## Prior Art
Tilix:
![Screenshot from 2019-10-15 10 53 28@2x](https://user-images.githubusercontent.com/7277719/66857570-6a2e3480-ef3c-11e9-98c3-edc007085b2a.png)

Xfce Terminal:
![Screenshot from 2019-10-15 11 07 06@2x](https://user-images.githubusercontent.com/7277719/66857571-6ac6cb00-ef3c-11e9-81f9-a441992215bd.png)

GNOME Terminal:
![Screenshot from 2019-10-15 11 07 37@2x](https://user-images.githubusercontent.com/7277719/66857573-6ac6cb00-ef3c-11e9-9a60-65d98e0ef2ae.png)

danrabbit:
It doesn't seem like any of the prior art is particularly contextual and it's hard to see how it will relate to the end result. Off the top of my head I feel like a good solution:

- [ ] Doesn't require hunting around in a preferences dialog. We can add a `( + )` add the end of the list of color options in the popover we have currently. This could then throw a dialog that specifically and only manages the stored color schemes.
- [ ] Includes a preview so we can see what the resulting color scheme actually looks like
- [ ] Labels the color selectors (of which it looks like only Tilix tries to do this)
- [ ] If we're really smart, we can do some color contrast checking and warn about color schemes that would be illegible. At the very least we should do this between the background and foreground color and make sure it passes the base A level

dougabugg:
Hi, any update on when we will be able to modify the colors on the terminal? Looks like this issue is 5 months old...

I just installed elementaryOS, but I am really bugged that the default terminal has such a poor color scheme. This is an example of what I'm seeing (or can't see, my eyes can't read the directory name):

![Screenshot from 2020-04-28 16 02 36](https://user-images.githubusercontent.com/18294318/80537488-c81aea80-8969-11ea-82a7-d135acc267d6.png)

I want to customize the Blue text on Green background, any solutions in the mean time?

PR #421, opened 2019-10-21
Add themes spoiler in Style menu

Redian-West:
I rendered constants for adding topics to a separate file. I think it will be easier to add new topics and edit existing ones.

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -25,11 +25,6 @@
     background-color: #000;
 }
 
-vte-terminal {
-    padding-left: 4px;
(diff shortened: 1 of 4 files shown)


jeremypw:
Needs input from the UX team.  Some screenshots would help.

danrabbit:
Hey thanks for this PR, but I'm going to close it in favor of the discussion occuring on #418

I think we probably don't want to expand the default selection of color schemes and instead would prefer to design a nice color scheme editor

Issue #283, opened 2018-07-15
Font changed only after terminal restart


Djaler:
Because of this "Conceal text" in Screenshot Tool doesn't work for Terminal content

PR #423, opened 2019-11-09
MainWindow: Apply font changes on the fly

Exalm:
Fixes https://github.com/elementary/terminal/issues/283

### DIFF
--- a/data/io.elementary.terminal.appdata.xml.in
+++ b/data/io.elementary.terminal.appdata.xml.in
@@ -45,6 +45,7 @@
         <p>Other updates:</p>
         <ul>
           <li>Include sudo password prompt when copying last output</li>
+          <li>Apply font changes without restarting</li>
           <li>Translation updates</li>
         </ul>
       </description>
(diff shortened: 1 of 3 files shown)


Exalm:
While the MR fixes it if you don't change hidden settings, it still won't work if you select a custom font. Though unlike Code, Terminal doesn't implement zooming via that preference. So, I wonder if this font should be special-cased to be applied even if a custom font is used.

jeremypw:
@danrabbit Do we need to have a separate font setting for Terminal, especially as it is not exposed in the UI?

Exalm:
Rebased, also moved system gsettings creation into `Application` to match the existing ones.

Exalm:
Oh wait, actually pushed the correct one now.

danrabbit:
@jeremypw I'm not sure I 100% understand the question, but the system monospace font is also used in Code, for example, so I don't think a Terminal-specific setting makes sense necessarily

Exalm:
Wel, Code has a setting. And it's even exposed in the UI and is also used for zooming. Terminal has a setting too, but it's only in gsettings.

jeremypw:
@danrabbit OK, so the setting `io.elementary.terminal.settings.font` could be removed and the system monospace font always used?

Exalm:
Added a line to appdata changelog.

PR #429, opened 2019-11-30
Fix POTFILES

ryonakano:


po/POTFILES
```diff
@@ -1,6 +1,4 @@
 src/Application.vala
-src/DBus.vala
```
tintou: This files exists in the repository

ryonakano: Yes, but it does not seem to contain any translatable strings

tintou: Better keep it then so that we get translations if for some reason there is a translatable string here in the future :)

PR #431, opened 2019-12-06
Strip clipboard when pasting

### LINKED ISSUE #420: Strip leading/trailing whitespace from pastes
This would fix the issue where copying a command with a newline at the end automatically executes it. It's less a security thing (a malicious attacker can always provide a specially malformed command to copy/paste), but more an annoyance thing.

cassidyjames:
Fixes #420 for the normal clipboard. Does not affect the primary clipboard, but neither does the paste protection.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -998,18 +998,18 @@ namespace Terminal {
         }
 
         private void on_get_text (Gtk.Clipboard board, string? intext) {
-            /* if unsafe paste alert is enabled, show dialog */
+            if (intext == null) {
+                return;
+            }
+
+            if (!intext.validate ()) {
+                debug ("Dropping invalid UTF-8 paste");
+                return;
+            }
+
+            var text = intext.strip ();
+
             if (Application.settings.get_boolean ("unsafe-paste-alert") && !unsafe_ignored ) {
-
-                if (intext == null) {
-                    return;
-                }
-                if (!intext.validate ()) {
-                    warning ("Dropping invalid UTF-8 paste");
-                    return;
-                }
-                var text = intext.strip ();
-
                 if ((text.index_of ("sudo") > -1) && (text.index_of ("\n") != 0)) {
                     var d = new UnsafePasteDialog (this);
                     if (d.run () == 1) {
@@ -1025,6 +1025,9 @@ namespace Terminal {
             if (board == primary_selection) {
                 current_terminal.paste_primary ();
             } else {
+                // Override clipboard with stripped version
+                Gtk.Clipboard.get_default (this.get_display ()).set_text (text, -1);
+
                 current_terminal.paste_clipboard ();
             }
         }


src/MainWindow.vala
```diff
@@ -998,18 +998,18 @@ namespace Terminal {
...
-                    warning ("Dropping invalid UTF-8 paste");
-                    return;
-                }
-                var text = intext.strip ();
+            if (!intext.validate ()) {
+                debug ("Dropping invalid UTF-8 paste");
+                return;
+            }
+
+            var text = intext.strip ();
```
tintou:
```suggestion
            unowned string text = intext._strip ();
```
That way no copy of the input string (which can potentially be huge :man_shrugging:)

tintou:
I haven't tested this branch yet but I have concerns that it might break the copy paste when space actually matters (let's say editing something with nano) so I don't think that it's worth the cost if that's true.

jeremypw:
Would it not be better to specifically remove trailing CR/LF as this was the main problem mentioned?

src/MainWindow.vala
```diff
@@ -998,18 +998,18 @@ namespace Terminal {
...
-                    warning ("Dropping invalid UTF-8 paste");
-                    return;
-                }
-                var text = intext.strip ();
+            if (!intext.validate ()) {
+                debug ("Dropping invalid UTF-8 paste");
+                return;
+            }
+
+            unowned string text = intext._strip ();
```
jeremypw: Why not just remove any trailing new line?

cassidyjames: Leading newlines could also unintentionally execute commands, but if it feels safer to only trim newlines and only trailing ones, then this could be adapted to do that as well.

jeremypw: We really need to know if we are pasting into the commandline or not.  If we are pasting into e.g. vim or an interactive interpreter (?) then we do not want to strip.

cassidyjames: @jeremypw that makes sense to me, but I'm unsure if/how we can do that.

jeremypw: @cassidyjames  I have pushed a commit that only processes the pasted text if there is no foreground process. This fixes pasting into terminal editors like nano or vim.  Also, I noticed that just stripping the string is insufficient to prevent command execution when pasting into the shell as it does not remove embedded newlines.  If you really want to prevent this, you have to make sure you only paste one line (without the newline).  My commit also does this. The UnsafePasteDialog (and setting) then becomes redundant as it is only triggered if there is a newline.

cassidyjames: @jeremypw yeah I was wondering about embedded newlines as well… but I wonder if this would be too aggressive/different from what people expect. I wonder if we could combine the checking with `sudo` and checking for newlines under the unsafe paste protection? I'll have to think about this a bit. But my initial thought is maybe:

- Treat newlines similar to how we treat `sudo` for paste protection, i.e.
  - Warn with an unsafe paste dialog (with modified copy) if a pasted command contains a newline
  - Turning unsafe paste protection off also turns off any special handling with newlines
- Maybe offer a way to paste one line without newlines in the dialog, i.e. as an alternate action? If that seems useful, we could also maybe strip `sudo` out from the sudo version, but idk if that's getting too complicated now.

jeremypw: Maybe better to take an incremental approach and deal with single line pastes first (probably 99% of cases?) and think about how to deal with embedded newlines at leisure.

jeremypw:
I would suggest at least reverting this to "Draft" as there are some possible problems.

PR #432, opened 2019-12-06
MainWindow: Natural Copy/Paste in menu

cassidyjames:
Inspired a bit by Ephemeral and discussions in #ux.

Screenshot from 2019-12-06 16 56 28@2x

Screenshot from 2019-12-06 16 56 33@2x


src/MainWindow.vala
```diff
@@ -354,32 +356,77 @@ namespace Terminal {
...
+            color_grid.add (color_button_dark);
+
+            var natural_copy_paste_label = new Gtk.Label (_("Natural Copy/Paste"));
+            natural_copy_paste_label.halign = Gtk.Align.START;
+            natural_copy_paste_label.hexpand = true;
+
+            var natural_copy_paste_switch = new Gtk.Switch ();
+            natural_copy_paste_switch.halign = Gtk.Align.END;
+
+            var natural_copy_paste_description = new Gtk.Label (_("<small>Shortcuts don’t require Shift, but may interfere with some interactive CLI apps</small>"));
```
danrabbit: I think you probably don't want to include the markup in the translatable string

src/MainWindow.vala
```diff
@@ -354,32 +356,77 @@ namespace Terminal {
...
+            color_grid.add (color_button_white);
+            color_grid.add (color_button_light);
+            color_grid.add (color_button_dark);
+
+            var natural_copy_paste_label = new Gtk.Label (_("Natural Copy/Paste"));
+            natural_copy_paste_label.halign = Gtk.Align.START;
+            natural_copy_paste_label.hexpand = true;
+
+            var natural_copy_paste_switch = new Gtk.Switch ();
+            natural_copy_paste_switch.halign = Gtk.Align.END;
```
danrabbit: I don't think this is necessary if you have the label expanding

src/MainWindow.vala
```diff
@@ -354,32 +356,77 @@ namespace Terminal {
...
+            natural_copy_paste_revealer.reveal_child = Application.settings.get_boolean ("natural-copy-paste");
+
+            var natural_copy_paste_grid = new Gtk.Grid ();
+            natural_copy_paste_grid.margin_start = natural_copy_paste_grid.margin_end = 12;
+
+            natural_copy_paste_grid.attach (natural_copy_paste_label, 0, 0);
+            natural_copy_paste_grid.attach (natural_copy_paste_switch, 1, 0);
+            natural_copy_paste_grid.attach (natural_copy_paste_revealer, 0, 1, 2);
+
+            var natural_copy_paste_button = new Gtk.Button ();
```
danrabbit: Instead of a regular button with style classes I think what would be better is using Gtk.ModelButton and then removing  its child like here: https://github.com/elementary/wingpanel-indicator-session/blob/master/src/Indicator.vala#L93

Right now there's some weird focus style stuff and that would fix it. You also wouldn't need to then add any margin to the grid

cassidyjames:
From @danrabbit's branch's additions:

image (1)

cassidyjames:
Shorter copy:

screenshot

PR #433, opened 2019-12-10
Use Modelbutton

danrabbit:
Uses a ModelButton instead of a Button to fix some styling issue. Screenshots:

Screenshot from 2019-12-09 18 05 41@2x
Screenshot from 2019-12-09 18 05 53@2x

I also messed with layout and alignment because I'm like that


cassidyjames:
So the thing I don't love (and @katiemkblaede brought up in impromptu user testing) is that the switch moves out from under the cursor when clicking it, which feels weird.

Attaching this was also makes the menu feel awkwardly wide to fit the label to me, so I think stacking below might work better in this context. For Night Light the descriptive text is only one line so it works to share space with the main label, but this is pretty different from that.

PR #435, opened 2019-12-13
Ctrl scroll zoom

### LINKED ISSUE #430: Zoom with Ctrl+Scroll
Like browsers and Code

jeremypw:
Fixes #430

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -185,11 +185,9 @@ namespace Terminal {
                             return Gdk.EVENT_STOP;
 
                         case Gdk.ScrollDirection.SMOOTH:
-                            double delta_x, delta_y;
-                            event.get_scroll_deltas (out delta_x, out delta_y);
                             /* try to emulate a normal scrolling event by summing deltas.
                              * step size of 0.5 chosen to match sensitivity */
-                            total_delta_y += delta_y;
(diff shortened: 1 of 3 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -171,6 +173,42 @@ namespace Terminal {
...
+                            increment_size ();
+                            return Gdk.EVENT_STOP;
+
+                        case Gdk.ScrollDirection.DOWN:
+                            decrement_size ();
+                            return Gdk.EVENT_STOP;
+
+                        case Gdk.ScrollDirection.SMOOTH:
+                            double delta_x, delta_y;
+                            event.get_scroll_deltas (out delta_x, out delta_y);
```
danrabbit: If you're only interesting in `delta_y` I think you can just do 
```suggestion
                            var delta_y = event.delta_y;
```

danrabbit:
Wow summing delta is way more elegant than the solution I came up with elsewhere. Definitely gonna steal that!

One thing is that it doesn't look like this updates the zoom level in the menu :)

jeremypw:
Oops - forgot about updating the menu - thanks.  This was more or less cut and pasted from Files code.

src/Widgets/TerminalWidget.vala
jeremypw: Thanks - done

Issue #440, opened 2020-01-04
Super+T dosen't open new Terminal anymore


The-Idle-Sloth:
Hello,

Since the update this morning, my shortcut to open a new terminal doesn't work anymore : the first I open works but once I have 1 teminal openned it doesn't open more terminal.

Regards,
Gregoire

The-Idle-Sloth:
Looks like it opens a new tab, not a new window. How do I correct it ?

jeremypw:
The behaviour of Terminal launched without any flags was changed to fix other issues reported.  To fix this regression I guess we would have to add another flag `-n` (New Window) which would force another window to be opened.  Would also have to change system shortcuts to use this flag (a custom shortcut could use it in the mean time).

queeup:
With `tab-bar-behavior 'Never Show Tabs'` and Super+T also ignore to open new window. It was working before.

The-Idle-Sloth:
Thanks for answer, but -n doesn't work for me, it doesn't open any new windows or tab...
I've tried it on command line and in custom shortcut menu but doesn't work for both.

Whatmore, may be update the -h option (and add --help option) so we may know how to fixe little problems like that without posting here ;-)

jeremypw:
@The-Idle-Sloth You need to wait until #447 has been merged (and released if you are on the stable version of Hera) before the `-n` flag will work.

endorama:
Hello, may I ask when you expect to release this? 

I use multiple workspace and this bug has the side effect of creating a tab in the terminal window opened regardless of the workspace I'm currently in (so ie. if I'm on workspace 3 but terminal is in workspace 1 I have to manually switch to workspace 1). Is very frustrating :) I can live with it, but I'd love an ETA for this to reach Hera. Thanks!

Kudos to the awesome work and this terminal anyway, I'm overall very happy with it!

jeremypw:
Unfortunately there has just been a release and normally there is a space of two or three months between releases unless there is a serious issue. If you are on the stable repository you could switch to the daily repository to get fixes as soon as merged but you would have to be aware that this would be more prone to have (temporary) regressions so is not suitable for production use.  There was a regression caused by the recent fix which needs #456 to be merged.

kr428:
[...]

PR #446, opened 2020-01-05
-n option to open a new window instead of a new tab

robn:
Still reopens tabs, so definitely the wrong way to do it, but gets my
workflow back well enough until a real fix.

#440


robn:
FYI, I am not expecting this to be merged, but I'm using it until #440 is fixed, and maybe someone else will find it useful.

jeremypw:
When this is installed you can create a custom shortcut which uses the new flag and reassigns `Super + T` to get the behaviour you want.

robn:
Closed by mistaked, reopened as #447.

PR #447, opened 2020-01-06
-n option to open a new window instead of a new tab

### LINKED ISSUE #440: Super+T dosen't open new Terminal anymore

robn:
Fixes #440 

[Reopening #446, which I closed by mistake]

Adds `-n` option to force a new window to open.


robn:
From #446:

> The get_last_window () function's name no longer reflects its actual action - it does not always return the last window, even if there is one. As it is a private function I think it would be safe enough to rename it and maybe pass the option as a parameter in case the function needs to be used to get the last window anyway. You could then create the new window in this function and return it instead of the callers having to test the result and then create a new window.

I started looking into this and it looks like there's a lot wrong with how windows are selected/created in response to options. I'll stick a pin in the PR and send some cleanups and through that sort all this out before coming back to this one.

jeremypw:
@robn  Please do not let the scope of the PR increase unnecessarily as it will delay approval and merging.

robn:
Alright, then I won't bother refactoring at all, and will just make sure restored tabs are handled properly. The rest I'll save for other PRs.

robn:
Ok, that's the simplest implementation of `-n` I know how to do. Have at it!

PR #448, opened 2020-01-07
Application: process help options immediately

robn:
Previously, help output would either be displayed in an existing help terminal, or just not appear at all. Neither of these are what we want; we want help to be displayed wherever the program was invoked from.

So instead, we look for the help args in main, and if present we don't even try to get running terminal instances or new windows involved, instead just showing the help directly.

I would have preferred to do all the commandline parsing in one place, either in main and then pass the results to the main thread, or in the command_line entry point and pass the output back to the caller, but it is not obvious how to do either of these things, so I did not.


jeremypw:
@robn  Please could you resolve the conflict with master, thanks.

robn:
Rebased onto master/5.5.1.

jeremypw:
@robn  Please could you give a test case that produces different results in master in this PR?

jeremypw:
Closing as master seems to already work correctly and no counter-examples given.  Please feel free to reopen.

PR #461, opened 2020-01-11
Fix issues with -e and -x options

### LINKED ISSUE #460: Options -e and -x open empty terminal
Executing a command with `io.elementary.terminal -n -e "uname -a"` or `io.elementary.terminal -n -x "uname -a"` just opens an empty terminal window.
The command is just used as an example here. I can not execute any command with either the -e or the -x flag.
This behavior extists since Terminal 5.5.1 or 5.5.0.

jeremypw:
Fixes #460 

* Do not add fallback tab in MainWindow creator when handling -e and -x flags
* Move empty directory handling to MainWindow.new_tab ()

Fixes regression caused by recent commits related to new windows and spurious duplicate tabs on window creation.

The creation of a new window was adding a tab which then stopped another tab being added in which to execute the command when there was no specified working directory.  

Another parameter was added to the MainWindow creation method which can suppress adding a tab on creation without restoration of tabs when the caller will add a tab anyway.

When used without a specified working directory, the -e and -x commands are now executed in the Enviroment working directory.


jeremypw:
Really need a complete review, overhaul of code related to opening new windows to see whether it can be simplified while still dealing with all different circumstances. Or at least made clearer.

jakubklos77:
Unforunately not even this fix makes it good again. There are multiple issues.

1. The app tries to match an existing tab for the given workdir. If it finds it it simply switches to that tab instead of creating a new tab. Suppose the already existing tab is running some command (top or whatever under /tmp). You start as 
`io.pantheon.terminal -e ls -w /tmp`
The app will switch to the tab with the "top" command running and nothing happens. 
This is bad a needs to be fixed

Either, validate if the tab is running an ongoing command and if it is then create a new tab.
If not, then switch to the tab and run the command.

2. Suppose you want to open the same path in multiple tabs. Currently this is not possible.
One tab is running "/tmp"
You start this
`io.pantheon.terminal -w /tmp`
The existing tab is used

So I guess we new a new argument to force a new tab (just like we have a new window).
This would at least make it possible to preserve some old compatibility

It was all somewhat working in the previous version.
What do you think?

jeremypw:
@jakubklos77 Thanks for review!  This PR is only meant to fix issues when using the `-e` and `-x` flags raised in #460 (and not cause any regressions).  

(1) I do agree that a tab that is running a process should be omitted from duplicate checking but I don't think Terminal ever did that so perhaps should be put in a separate PR.

(2) Regarding forcing opening a duplicate tab from the commandline, again this might be useful to someone but is a bit of a corner case unless a process is running in which it would be addressed by fixing point (1). The behaviour of the master branch is also to not open duplicate tabs so this should be the subject of a separate issue/branch.

When you say it "was all somewhat working" what version are you comparing this PR with?

jakubklos77:
@jeremypw Thank you for your response. Yes, I kind of thought it would have to be new PRs. 

I was not referring to any specific version, but the behavior. In some older version (1 month and more) I don't think any existing tabs were matched and always a new one was created.

Anyway, I am new to the code and made some changes and it works ok. I can make those PR. How hard and how long does it take to get those PRs to get merged to the master? (I don't have much experience so I am asking). Also, do you have any idea how to check if a process is running in the terminal widget? If you don't, I will analyze the code :)

Thank you again

jeremypw:
@jakubklos77 Good to hear you are starting to contribute code!  The time it can take for PRs to get reviewed, approved and merged varies enormously from hours to more than a year (if ever).  It depends on how many changes are made, how critical they are and how easy to understand/test/review.  That is why it is a good idea to make as few substantive changes as possible in each PR (something I do not always achieve!) and explain/comment them well.

jakubklos77:
@jeremypw Understood, thank you.

jakubklos77:
@jeremypw Made a small fix 
https://github.com/elementary/terminal/pull/476

jeremypw:
Timed out.

PR #467, opened 2020-01-21
Don't access network during appstream validate

danrabbit:
It looks like newer appstream is trying to validate URLs and Launchpad is failing to build because of it.

If we want to use appstream-cli we should probably do that as part of CI and not in the repo like this


tintou:
You can use `--no-net` argument to prevent it to reach the internet and keep that validation

danrabbit:
@tintou looks like this is failing even with "--no-net". I think we should move this to the actual CI anyways so it doesn't need to be here

PR #468, opened 2020-01-22
Ctrl+Click to open URLs

### LINKED ISSUE #469: Ctrl + Click to open URLs
## Prerequisites

## Feature
Sometimes there are lots of URLs inside window. I tend to mistakenly click them when trying to select some text to copy.

It's quite common to hold Ctrl/Cmd and click to open URLs.

So I created PR #468 to address this issue.

I will create another PR against `elementary-tweaks` for configuration.

eGust:
Fixes #469 

1. added boolean option `control-click-open-url` by default `false`
2. when enabled will require holding Control and click to open URLs


src/Widgets/TerminalWidget.vala
```diff
@@ -155,7 +155,11 @@ namespace Terminal {
             button_release_event.connect ((event) => {
 
                 if (event.button == Gdk.BUTTON_PRIMARY) {
-                    if (allow_hyperlink) {
+                    var required_control = Application.settings.get_boolean ("control-click-open-url");
+                    var openable = required_control
+                        ? (event.state & Gdk.ModifierType.CONTROL_MASK) > 0
```
jeremypw: This is `true` even if other modifiers are held down in addition to `Ctrl`.  Not sure is that matters in this case.

eGust: several lines below the code for scrolling is
`if ((event.state & Gdk.ModifierType.CONTROL_MASK) > 0)`

I don't mind to open links with holding Ctrl + Shift, as currently no matter what modifiers it still opens up.

jeremypw: Yes, I am not too concerned, although Files code does check that only `<Ctrl>` is held down.  I guess it can be changed if an issue arises.

src/Widgets/TerminalWidget.vala
```diff
@@ -166,7 +170,7 @@ namespace Terminal {
                             }
                         }
                     } else {
-                        allow_hyperlink = true;
+                        allow_hyperlink = !required_control;
```
jeremypw, line 173: I dont think these two booleans should be linked they have different purposes.

jeremypw:
@eGust  In order to shorten the review list, I am converting this to draft as there are conflicts, outstanding issues and no recent activity.  Is this something you still have time to work on?

jeremypw:
Closing due to age and lack of response

PR #476, opened 2020-01-27
-e execute skips active tabs search

jakubklos77:
With the new changes from 2df52a9c334be97e702793bc6938c50f1e668251 it is now impossible to start the terminal with a command of already open tab:
```
io.elementary.terminal -e ls -w /tmp
```
If there is already a tab with the path of /tmp then such tab is activated and no command is run. Hence we have a new bug.

This commit makes it possible to run such commands in a working directory which has already a tab open.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -288,16 +288,19 @@ namespace Terminal {
                 location = directory;
             }
 
-            var f1 = File.new_for_commandline_arg (location);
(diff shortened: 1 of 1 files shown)


jakubklos77:
Ideally, there should be also a way to force a new tab creation even for existing tabs of the desired working directoy (just like we have for a new window). I will create a new PR for that.

PR #477, opened 2020-01-27
Get pwd from ApplicationCommandLine if possible when no -w flag given

jeremypw:
In master, if an instance of Terminal was already running, then the environment of that instance was being used, not the environment of the command.

It seems more reasonable, if a working directory was not specified, that Terminal should use the pwd of the command line, e.g. when opening a new window or executing a command.

Also provides a route to a solution for https://github.com/elementary/granite/issues/438 and https://github.com/elementary/terminal/issues/640


src/Application.vala
```diff
@@ -190,6 +221,8 @@ public class Terminal.Application : Gtk.Application {
             }
         }
 
+
+
```
danrabbit, line 225: I guess this was not intentional :)

src/Application.vala
```diff
@@ -172,14 +174,43 @@ public class Terminal.Application : Gtk.Application {
             return 0;
         }
 
+        /* If working directory not specified get from any command supplied with a full path */
```
danrabbit, line 177: Is there a reason this isn't inside the conditional below? It seems like if `option_help` or `option_version` are true, then this wouldn't be relevant would it?

danrabbit (commented):
Left a question inline :)

What's the best way to test this?

jeremypw:
On way of testing this is to create this script in a directory other than your home directory:
```
#!/bin/bash

echo "Hello World!"
echo My current working directory is: $PWD
sleep 5s
```

Navigate to this script in Files and choose "Open in ... Terminal" from the context menu.

With master the script runs with the home directory as the PWD.  With this PR the script runs with the directory where the script is located as the PWD.

src/Application.vala
jeremypw: I usually take a comment to refer to the code following it but in this case it could be inside the conditional - I'll move it.

danrabbit: Sorry I meant the whole code block, not just the comment. It seems like we don't need to get the current working directory if the user is just printing the help or version right?

jeremypw: Oh, I see. Well, it would make the following code block longer and more difficult to follow.  It is clearer (imo) to make sure that the working directory is always set to something as a fallback before processing the flags (even though in some cases the fallback is not used).  The whole thing could be pasted after the second `else` (line 202) but all that does is add an extra indent. Another alternative is to repeat the test for null working directory four times.

jeremypw: I'll rethink it.

jeremypw:
I believe the message `[GLib] setenv()/putenv() are not thread-safe and should not be used after threads are created` that is printed before and after any output from the command is due to an upstream bug.

PR #478, opened 2020-01-28
Support for new "-t" option (create new tab)

jakubklos77:
With the new behavior of the terminal trying to match existing tab's path it is impossible to open a new tab with such existing path (working directory option -w).

- Suppose you have the terminal app open with the tab of location /tmp (running some command or simply waiting)
- Now you want to open another instance of the same path using:
```
io.elementary.terminal -w /tmp
```
This will simply activate the existing tab and nothing else will happen. Hence we cannot achieve the desired resuilt.

A new "-t" option to force a tab creation is the right solution (just lke we have for a new window).
This PR introduces the new option.

Newly, you can open the terminal with:
```
io.elementary.terminal -t -w /tmp
```

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -35,6 +35,9 @@ public class Terminal.Application : Gtk.Application {
     // option_new_window will be true if the new-window flag was given.
     private static bool option_new_window = false;
 
+    // option_new_tab will be true if the new-tab flag was given.
+    private static bool option_new_tab = false;
+
     public int minimum_width;
     public int minimum_height;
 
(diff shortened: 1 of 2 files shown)


jakubklos77:
> This does not add a new tab if there is no explicit working directory.
> I think you need to change MainWindow.vala line 281 to
> 
> ```
>             if (directory == null || directory == "" || create_new_tab) {
> ```

You are right but the solutionn should be different. See my commit 
00350b8. I think this is the right way
```
if (directory == null || directory == "") {
                if (notebook.tabs.first () == null || command!=null || create_new_tab) { //Ensure at least one tab
                    new_tab ("", command);
                }
```
General conditon for empty directory remains but the new_tab conditon is extended for
- existing command - this way it will be executed as well
- create_new_tab

PR #486, opened 2020-03-04
Add "Kill process" to context menu

### LINKED ISSUE #271: "Kill process" in the context menu
I think it would be nice if you could forcefully kill the current running process in a tab. Sometimes the process doesn't respond to any signals like Ctrl+C and you end up closing the tab altogether.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59923646-kill-process-in-the-context-menu?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

franciol:
Fixes #271

### DIFF
--- a/src/Dialogs/ForegroundProcessDialog.vala
+++ b/src/Dialogs/ForegroundProcessDialog.vala
@@ -44,6 +44,20 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
         );
     }
 
+    public ForegroundProcessDialog.kill_process (MainWindow parent) {
+        Object (
+            transient_for: parent,
+            primary_text: _("Are you sure you kill this process?"),
+            secondary_text:
+                _("There is an active process on this terminal.") + " " +
+                _("Are you sure you want to kill it?"),
+            buttons: Gtk.ButtonsType.CANCEL,
+            button_label: _("Kill Process")
+        );
+    }
+
+
+
     construct {
         image_icon = new ThemedIcon ("dialog-warning");
 
--- a/src/Dialogs/ForegroundProcessDialog.vala
+++ b/src/Dialogs/ForegroundProcessDialog.vala
@@ -47,7 +47,7 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
     public ForegroundProcessDialog.kill_process (MainWindow parent) {
         Object (
             transient_for: parent,
-            primary_text: _("Are you sure you kill this process?"),
+            primary_text: _("Are you sure you want to kill this process?"),
             secondary_text:
                 _("There is an active process on this terminal.") + " " +
                 _("Are you sure you want to kill it?"),
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1315,14 +1315,14 @@ namespace Terminal {
 
 
         private void action_terminate () {
-            if(current_terminal.has_foreground_process()){
+            if (current_terminal.has_foreground_process ()) {
                 var d = new ForegroundProcessDialog.kill_process (this);
 
                 if (d.run () == Gtk.ResponseType.ACCEPT) {
                     current_terminal.kill_fg ();
-                    d.destroy();
-                } 
-                else{
+                    d.destroy ();
+                }
+                else {
                     d.destroy ();
                 }
             }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -71,6 +71,7 @@ namespace Terminal {
         public const string ACTION_SEARCH = "action-search";
         public const string ACTION_SEARCH_NEXT = "action-search-next";
         public const string ACTION_SEARCH_PREVIOUS = "action-search-previous";
+        public const string ACTION_TERMINATE = "action-terminate";
         public const string ACTION_SELECT_ALL = "action-select-all";
         public const string ACTION_OPEN_IN_FILES = "action-open-in-files";
         public const string ACTION_SCROLL_TO_LAST_COMMAND = "action-scroll-to-last-command";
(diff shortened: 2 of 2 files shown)


src/Dialogs/ForegroundProcessDialog.vala
```diff
@@ -44,6 +44,20 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
         );
     }
 
+    public ForegroundProcessDialog.kill_process (MainWindow parent) {
+        Object (
+            transient_for: parent,
+            primary_text: _("Are you sure you kill this process?"),
```
jeremypw: Missing words: Should be e.g. "Are you sure you want to kill this process?"

src/MainWindow.vala
```diff
@@ -1305,6 +1313,21 @@ namespace Terminal {
             }
         }
 
+
```
jeremypw: Superfluous blank line

src/MainWindow.vala
```diff
@@ -1305,6 +1313,21 @@ namespace Terminal {
...
+        private void action_terminate () {
+            if (current_terminal.has_foreground_process ()) {
+                var d = new ForegroundProcessDialog.kill_process (this);
+
+                if (d.run () == Gtk.ResponseType.ACCEPT) {
```
jeremypw: No need to repeat `d.destroy ()` - move after the if clause and remove else clause.

jeremypw:
Congratulations on making your first contribution to elementary!   If you put "Fixes #nnn" in the description then the related issue is automatlically closed if the PR is merged into master.  I have done this for you.

I have made a few small code style points but a larger issue is that this does not work for processes that intercepts keystrokes and button presses on the terminal widget (I tried with htop although that has its own quit method). One possible solution would be to put the terminate process menuitem in the tab menu.

jeremypw:
I see you have moved the terminate process menuitem to the app menu rather than the tab menu.  The problem with this is that there is not a direct connection between that menu and the tab to which the option applies.  I think it would be more intuitive to have it in the tab menu (right-click on the tab of the relevant terminal).  Also you need to make sure the option appears (or is only sensitive) when there is a process to terminate.

I'm going to flag this for UX team input with regard to the optimum placement of the option.

jeremypw:
In order to add an item to the tab menu you will need to connect to the "popped-up" signal of the tab menu.  As the tab is associated with the terminal after the terminal is created, you need to do this in MainWindow::create_tab (). For example:

```
            term.tab.menu.popped_up.connect (() => {
                if (term.has_foreground_process ()) {
                    var item = .... //Create the menuitem here
                    item.show_all ();
                    item.sensitive = true;
                    term.tab.menu.append (item);
                }
            });
```

Ideally the menu item label should include the name of the process to be terminated but that is not so easily obtained.

src/Dialogs/ForegroundProcessDialog.vala
```diff
@@ -44,6 +44,20 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
...
 
+    public ForegroundProcessDialog.kill_process (MainWindow parent) {
+        Object (
+            transient_for: parent,
+            primary_text: _("Are you sure you want to kill this process?"),
+            secondary_text:
+                _("There is an active process on this terminal.") + " " +
+                _("Are you sure you want to kill it?"),
+            buttons: Gtk.ButtonsType.CANCEL,
+            button_label: _("Kill Process")
```
danrabbit: I feel like we should avoid the word "Kill". "Terminate" is probably better language

src/MainWindow.vala
```diff
@@ -464,6 +467,7 @@ namespace Terminal {
             menu_popover_grid.add (color_grid);
             menu_popover_grid.add (new Gtk.Separator (Gtk.Orientation.HORIZONTAL));
             menu_popover_grid.add (natural_copy_paste_button);
+            menu_popover_grid.add (terminate_process_button);
```
danrabbit: The settings menu probably isn't the right place for this. All of the other menu items here effect all of your tabs.

A more appropriate location is probably either the tab menu or the context menu

danrabbit:
I'd also like to mention @elementary/ux here for some additional thought on if a dialog is necessary here. We've never had a dialog when the user terminates a process using the keyboard shortcut.

My gut reaction is that if we feel like the menuitem needs a dialog then the keyboard shortcut should as well. We warn when closing a tab about running process because of possible unintended consequences. But if a user is intentionally terminating a process, I'm not certain that it makes sense to ask if they're sure.

jeremypw:
Another issue I noticed was that if you terminal a process such as `htop` then the terminal is not cleared and behaves differently to a normal terminal in that (e.g.) scrolling on it injects characters.  We probably want to keep any output of the process and subsequent input should (a) not overwrite the process output and (b) behave normally.

src/Dialogs/ForegroundProcessDialog.vala
cassidyjames: @danrabbit even "Terminate" seems a little… excessive. Couldn't it just be “End”?

jeremypw: There are technical differences between "Kill" and "End / Terminate" although we may not need trouble the user with them.  I presume the intended use of the option would be to forcibly end (kill) a process that could not be ended normally.

jeremypw:
@franciol Hi!  Are you still interested in working on this?

jeremypw:
Re: Having a dialog, I guess we can assume anyone running a process in the terminal probably knows what they are doing so probably need less hand holding that someone using a graphical app?

franciol:
> @franciol Hi! Are you still interested in working on this?

Hi @jeremypw! Right now i can't work on this. Maybe on 2 weeks i can look at it....

jeremypw:
Converting to draft as there are outstanding issues, conflicts and no recent activity.

jeremypw:
Closing due to age and lack of activity

PR #504, opened 2020-04-14
Use escape sequence and ignore clicks on command prompt

andreasfelix:


src/Widgets/TerminalWidget.vala
```diff
@@ -148,27 +149,23 @@ namespace Terminal {
                     get_clicked_cell_position (event, out clicked_row, out clicked_col);
                     remember_command_start_position ();
 
-                    if (clicked_row < remembered_command_start_row) {
+                    if (
+                        clicked_row < remembered_command_start_row
+                        || (clicked_row == remembered_command_start_row
+                        && clicked_col < remembered_command_start_col)
+                    ) {
```
jeremypw, line 156: This is not elementary code style -  should be:
```
                    if (clicked_row < remembered_command_start_row  || 
                       (clicked_row == remembered_command_start_row &&
                       clicked_col < remembered_command_start_col)) {

                            return Gdk.EVENT_PROPAGATE
                    }
```

src/Widgets/TerminalWidget.vala
```diff
@@ -148,27 +149,23 @@ namespace Terminal {
...
                     set_color_cursor (background_color);
                     Idle.add (() => { /* wait for button press event to be processed */
                         /* Cursor will move as close as possible to pointer */
                         var n_events = (int) delta_cells.abs ();
+                        var sequence = delta_cells > 0U ? "\033[C" : "\033[D";
                         for (int i = 0; i < n_events; i++) {
-                            key_event.time = (uint32)(get_monotonic_time ());
-                            key_press_event (key_event);
+                            feed_child (sequence, 3);
                         }
```
jeremypw, line 169: I think you can use a single sequence to move the cursor <N> places.  Either `\033[<N>C` or `\033[<N>D`

andreasfelix:
The last commit fixes an issue, where it was possible to lock the `cursor_blink_mode` to off when clicking very fast. This was because the `cursor_blink_mode` was reset in on button_release event. When two `button_press` events where processed in a row, the `previous_cursor_blink_mode` was set to `OFF`. The solution is to make the `previous_cursor_blink_mode` a local variable and reset it together with the cursor color.

Another issue was that `Gdk.threads_add_idle_full` does not wait for `feed_child`, so one has revert to a hard-coded 50ms second timeout. IMO this is still a lesser "evil"...

src/Widgets/TerminalWidget.vala
andreasfelix: I also thought so, but I did not have the desired effect. Instead it moves only one position and deletes the last character. Maybe I am doing something wrong... 

There is also a sequence to move the cursor to a fixed position `Esc[Line;ColumnH`. But I was also not able to make it work ...

jeremypw:
OK, I'll see if I can get it to work otherwise will merge as is.  I am beginning to wonder whether the whole PR is going to get too hacky to get approved anyway :-( .  A Vte terminal is not designed to be used like a text editor I guess.

andreasfelix:
But is it really that hacky?

**Challenge:**
1. Get the clicked cell in a reliable way.
2. Move the cursor by *delta* cells.

**Implementation:**

1. The `get_cell_clicked` uses the `vadjustment` method of the `Gtk.Scrollable` to determine the clicked cell. I see no hack there.

2. To my understanding the escape sequences are the official "API" to move the cursor in a terminal. So I would argue that this is also not a hack, it's just a weird "API".

The only thing that is a bit hacky is the code that avoids the flickering of the cursor. But the flickering is not specific to the click implementation. The flickering is always there, it is just more visible because the cursor is moved multiple positions at once. It for example also occurs when pressing the <kbd>Home</kbd> or <kbd>End</kbd> keys to move the cursor. One could simply leave this code out and accept the flickering. But as there is a reliable way to avoid it, I think it is reasonable to keep it.

andreasfelix:
I think the last outstanding problem is, that the current way of getting the `command_start_row/col` values (which are necessary to ignore clicks on output) is not super reliable. But this is also a general problem and not specific to this PR...

src/Widgets/TerminalWidget.vala
andreasfelix: @jeremypw  Do you understand why on https://valadoc.org/vte-2.91/Vte.Terminal.feed_child.html the `feed_child` has a diffenerent signature (in comparison to `feed_child (string, int)`)

andreasfelix:
I removed the `Idle.add`, which to my understanding should no longer be necessary, as the cursor is not moved via `key_press_events` anymore. Revert this commit if I am wrong.

andreasfelix:
The last commit uses the `allow_hyperlink` to ignore clicks when window was not focused. It think this is necessary, because otherwise the cursor gets moved if one copies a command from somewhere and then focuses the terminal. Maybe the variable should be renamed to `is_focused` or something if we also use it here.

jeremypw:
> But is it really that hacky?

Well, its up to reviewers like tintou/danrabbit to decide that uptimately (referring to #386).  Once we are happy with it we should request a review from them.

jeremypw:
Thanks for all your work on this anyway!

PR #505, opened 2020-04-17
Zoom per tab

### LINKED ISSUE #480: Separate zoom level for different windows
## Prerequisites

## Feature
I have laptop and external monitor both have the same resolution but different screen size. When I open two terminal windows, one on laptop screen and another on external screen they have the same zoom level so either text on laptop is too tiny or text on monitor is too big.

I'd like to change zoom level on both windows independently.

jeremypw:
Fixes #480 

This allows different zoom levels to be set, saved and restored for each tab.  

The existing "zoom" setting is used for the font-scale of new tabs and for when the "default-zoom" button is pressed.

To be resolved:
* Whether new tabs should, instead, follow the font-scale of the currently active terminal


jeremypw:
@cassidyjames Could I have a ruling on what the correct behaviour for new tabs is (i.e. default zoom or zoom of current tab)?

cassidyjames:
@jeremypw Yeah I'm kinda conflicted. I would say, if it's a new tab it should have the default zoom, but we don't have a way to set a default zoom separately so I'm kind of conflicted on this feature in the first place. :sweat_smile:

One thing is that if you're duplicating a tab, we should _probably_ duplicate that tab's zoom level, too?

jeremypw:
My feeling is that we should take the zoom level of the current tab both for "New Tab" and "Duplicate Tab".  I would guess in most cases users would choose a zoom level that is comfortable for them and stick with it for most tabs and only adjust it for certain outputs.

jeremypw:
Konsole has seperate zoom per tab but XFce Terminal synchronises the zoom across tabs so ...(*shrug*).  If you think this unlikely to be merged then I'll abandon it and mark the issue "Design Conflict".   The respondent seemed quite fired up about it though.

cassidyjames:
@jeremypw that's a good point that if you set a comfortable default, then we should just continue using that by default for new tabs. And then yeah, if you adjust a specific tab, it shouldn't affect the other tabs. I am fine with that logic.

jeremypw:
@cassidyjames  I have implemented using the last active tab zoom as the starting point for new and duplicate tabs and where "remember-tabs" is turned off. 
I have maintained the option to toggle this feature on and off, but now made the default as on (to be confirmed).
I am still a bit unsure about what the correct action of the "Default zoom" button should be.  It is a bit unpredictable at the moment as it uses the "zoom" setting which, at the moment, is the zoom level of the active tab when the app was closed.
Alternatives would be always to use 100% zoom for this button, or to use it to restore the zoom at the moment the menu was opened - i.e. undo zoom changes.

cassidyjames:
I think the expectation is that it always sets back to 100%, which is what browsers do.

jeremypw:
@cassidyjames OK, I have implemented that.

data/io.elementary.terminal.gschema.xml
```diff
@@ -173,5 +178,10 @@
       <summary>Alert the user about unsafe paste.</summary>
       <description>If the user paste's a command into the terminal that might be considered unsafe, warn them.</description>
     </key>
+    <key name="sync-zoom" type="b">
+      <default>false</default>
+      <summary>Whether all tabs have the same zoom</summary>
+      <description>Whether changes to one tab's zoom level should be applied to all tabs</description>
+    </key>
```
cassidyjames, line 185: I don't think we should add this to be honest. It's such a specific, finicky setting. We should just ensure the default behavior makes sense. :smile: 

```suggestion
```

src/MainWindow.vala
```diff
@@ -453,6 +442,26 @@ namespace Terminal {
...
+            sync_zoom_grid.column_spacing = 12;
+            sync_zoom_grid.hexpand = true;
+
+            sync_zoom_grid.attach (sync_zoom_label, 0, 0);
+            sync_zoom_grid.attach (sync_zoom_switch, 1, 0);
+
+            var sync_zoom_button = new Gtk.ModelButton ();
+            sync_zoom_button.get_child ().destroy ();
+            sync_zoom_button.add (sync_zoom_grid);
+
```
cassidyjames, line 464: See previous comment.
```suggestion
```

src/MainWindow.vala
```diff
@@ -464,6 +473,7 @@ namespace Terminal {
             menu_popover_grid.add (color_grid);
             menu_popover_grid.add (new Gtk.Separator (Gtk.Orientation.HORIZONTAL));
             menu_popover_grid.add (natural_copy_paste_button);
+            menu_popover_grid.add (sync_zoom_button);
```
cassidyjames, line 476:
```suggestion
```

src/MainWindow.vala
```diff
@@ -521,9 +531,35 @@ namespace Terminal {
...
+                if (sync_zoom_switch.active) {
+                    var scale = current_terminal.font_scale;
+                    foreach (TerminalWidget t in terminals) {
+                        t.font_scale = scale;
+                    }
+
+                    /* Also sync new tab zoom */
+                    Terminal.Application.saved_state.set_double ("zoom", scale);
+                }
+
```
cassidyjames, line 545:
```suggestion
```

src/MainWindow.vala
```diff
@@ -559,6 +595,11 @@ namespace Terminal {
                 return Gdk.EVENT_STOP;
             });
 
+            sync_zoom_button.button_release_event.connect (() => {
+                sync_zoom_switch.activate ();
+                return Gdk.EVENT_STOP;
+            });
+
```
cassidyjames, line 602:
```suggestion
```

src/MainWindow.vala
```diff
@@ -573,6 +614,14 @@ namespace Terminal {
...
             );
 
+            Application.settings.bind (
+                "sync-zoom",
+                sync_zoom_switch,
+                "active",
+                SettingsBindFlags.DEFAULT
+            );
+
+
```
cassidyjames, line 624:
```suggestion
```

jeremypw:
@cassidyjames OK, the sync-zoom stuff is gone :-)

src/MainWindow.vala
```diff
@@ -535,9 +524,25 @@ namespace Terminal {
...
+                                                              zoom_default_button,
+                                                              "label",
+                                                              BindingFlags.DEFAULT,
+                                                              (binding, from_val, ref to_val) => {
+                                                to_val.set_string (font_scale_to_zoom (from_val.get_double ()));
+                                                return true;
+                                                              });
+
+                menu_popover.set_data<Binding> ("zoom-binding", binding);
+            });
```
cassidyjames, line 544: I think something like this would more closely follow our whitespace conventions and is much easier to read:

```suggestion
            menu_button.pressed.connect (() => {
                zoom_default_button.label = font_scale_to_zoom (current_terminal.font_scale);
                var binding = current_terminal.bind_property (
                    "font-scale",
                    zoom_default_button,
                    "label",
                    BindingFlags.DEFAULT,

                    (binding, from_val, ref to_val) => {
                        to_val.set_string (font_scale_to_zoom (from_val.get_double ()));
                        return true;
                    }
                );

                menu_popover.set_data<Binding> ("zoom-binding", binding);
            });
```

src/MainWindow.vala
```diff
@@ -966,13 +978,17 @@ namespace Terminal {
...
+            int index = 0;
             foreach (string loc in tabs) {
                 if (loc == "") {
                     focus--;
-                    continue;
                 } else {
-                    new_tab (loc, null, false);
+                    var term = new_tab (loc, null, false);
+                    term.font_scale = double.parse (zooms[index]).clamp (TerminalWidget.MIN_SCALE,
+                                                                         TerminalWidget.MAX_SCALE);
```
cassidyjames, line 988: Similar to above, I think this more closely follows our modern conventions:

```suggestion
                    term.font_scale = double.parse (zooms[index]).clamp (
                        TerminalWidget.MIN_SCALE,
                        TerminalWidget.MAX_SCALE
                    );
```

cassidyjames (changes requested):
Behavior-wise I think this makes sense. Just a couple of code style/readability things.

data/io.elementary.terminal.gschema.xml
```diff
@@ -44,8 +44,13 @@
     </key>
     <key name="zoom" type="d">
       <default>1.0</default>
-      <summary>Most recent zoom level of terminal</summary>
-      <description>Most recent zoom level of terminal</description>
+      <summary>Default zoom level of new terminal</summary>
+      <description>Default zoom level of new terminal</description>
+    </key>
+    <key name="tab-zooms" type="as">
+      <default>[]</default>
```
cassidyjames, line 51:
```suggestion
      <default>['1.0']</default>
```

cassidyjames (changes requested):
@jeremypw hm, installing this on elementary OS 6, it defaulted to 25% zoom the next time I opened Terminal. This is reproducible by setting `/io/elementary/terminal/saved-state/tab-zooms` to the default, then re-opening Terminal.

I _think_ my suggested change will resolve that by defaulting to 1.0.

data/io.elementary.terminal.gschema.xml
jeremypw: Hmm, that is not the intended behaviour - there is already a separate "zoom" setting for the fallback zoom for new tabs.  I expected that to be used where a tab did not have a corresponding "tab-zoom" value (which cannot guaranteed as the settings could be manually sent to an empty or invalid array regardless of the default value).  I'll look at this part of the code more closely.

jeremypw: OK, recent commits should fix the problem.

jeremypw:
Had to fix weird regression where Terminals crashed if closed immediately after adjusting a tab zoom level.  Not obvious why this did not occur in master as the relevant code had not been changed by this PR.  Seems to be an obscure race where Terminal attempts to save the position of `notebook.current` after it becomes `null` during a delete event.  I guess the new code affected the winner of the race ...

data/io.elementary.terminal.gschema.xml
```diff
@@ -42,10 +42,15 @@
         <summary>The index of the tab that is focused.</summary>
         <description>The index of the tab from the saved tabs that is currently focused.</description>
     </key>
-    <key name="zoom" type="d">
+    <key name="zoom" type="d"><range min="0.25" max="4.0"/>
```
cassidyjames, line 45:
```suggestion
    <key name="zoom" type="d">
    <range min="0.25" max="4.0" />
```

cassidyjames (commented):
I'm generally happy with this UX-wise. I would still like to get a code review from @elementary/desktop-developers though.

jeremypw:
Need to decide what a suitable max scale should be - just noticed I used 4.0 in one place and 5.0 in another. Suggestions?

data/io.elementary.terminal.gschema.xml
```diff
@@ -43,9 +43,15 @@
         <description>The index of the tab from the saved tabs that is currently focused.</description>
     </key>
     <key name="zoom" type="d">
+    <range min="0.25" max="4.0" />
```
meisenzahl, line 46: What effect do these values have?

I tested in Terminal with 3 tabs: 25% (min), 100% (default) and 400% (max)

These were the values in the settings: `['0,2', '1,0', '4,0']`

jeremypw: These values prevent manual entering of out of range default zoom values into the settings using dconf-editor or otherwise.  They are not essential as the code clamps the values it reads in from the settings anyway, but they inform the user.

meisenzahl: Yes, but do the values make sense if I can bypass them via the UI?

`['0,2', '1,0', '4,0']`

From my point of view it would then have to be changed as follows:

```suggestion
    <range min="0.2" max="4.0" />
```

jeremypw:
Yes, as I noted in a previous comment there are some discrepancies in the minimum and maximum zoom levels set in different places and asked what the best values would be.  Once UX team have adjudicated on that I'll ensure these values are consistently applied.

jeremypw:
I have decided to make a choice for zoom range myself - it is now 0.25 - 4.0 which seems more than sufficient to me.

meisenzahl (approved):

PR #512, opened 2020-05-09
Validate colors restored from settings; correct and replace as required.

### LINKED ISSUE #511: Adding an extra color to the theme field in dconf breaks the terminal
If you accidentally add an extra color to the theme via dconf-editor, it crashes the terminal and stops it from opening again. It should either truncate the parameters, not let you save, or ignore the extra values.

jeremypw:
Fixes #511

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -247,32 +247,54 @@ namespace Terminal {
             Gdk.RGBA foreground_color = Gdk.RGBA ();
             foreground_color.parse (Application.settings.get_string ("foreground"));
 
-            string[] hex_palette = {
+            const int PALETTE_SIZE = 16;
+            const string[] HEX_PALETTE = {
                 "#073642", "#dc322f", "#859900", "#b58900",
                 "#268bd2", "#ec0048", "#2aa198", "#94a3a5",
                 "#586e75", "#cb4b16", "#859900", "#b58900",
                 "#268bd2", "#d33682", "#2aa198", "#EEEEEE"
             };
 
-            string current_string = "";
-            int current_color = 0;
-            for (var i = 0; i < Application.settings.get_string ("palette").length; i++) {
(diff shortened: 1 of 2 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -247,32 +247,54 @@ namespace Terminal {
...
-                    current_string += Application.settings.get_string ("palette")[i].to_string ();
-                }
+            var hex_palette = new string[PALETTE_SIZE];
+            var palette_setting_string = Application.settings.get_string ("palette");
+            var setting_palette = palette_setting_string.split (":", PALETTE_SIZE + 1);
+
+            bool settings_valid = setting_palette.length <= PALETTE_SIZE;
+
+            int i = 0;
+            foreach (string hex in setting_palette) {
```
tintou, line 265:
```suggestion
            foreach (unowned string hex in setting_palette) {
```

src/Widgets/TerminalWidget.vala
```diff
@@ -247,32 +247,54 @@ namespace Terminal {
...
+             /* Remove invalid colors from setting */
+                var sb = new StringBuilder ("");
+                foreach (string hex in hex_palette) {
+                    sb.append (hex);
+                    sb.append (":");
+                }
 
-                palette[i] = new_color;
+                sb.truncate (sb.len - 1);
+                Application.settings.set_string ("palette", sb.str);
```
tintou, line 297: What you want is this :slightly_smiling_face: 
```suggestion
                Application.settings.set_string ("palette", string.joinv (":", hex_palette));
```

src/Widgets/TerminalWidget.vala
jeremypw: Done

src/Widgets/TerminalWidget.vala
jeremypw: Much better - thanks.

PR #527, opened 2020-08-06
Application.css: increase padding

danrabbit:
Since that's our standard window padding

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -26,8 +26,7 @@
 }
 
 vte-terminal {
-    padding-left: 4px;
-    padding-right: 4px;
+    padding: 0 12px;
 }
 
 .color-button radio,
--- a/data/Application.css
+++ b/data/Application.css
@@ -26,7 +26,7 @@
 }
 
 vte-terminal {
-    padding: 0 12px;
+    padding: 0 9px;
 }
 
 .color-button radio,


cassidyjames:
Hrm, this feels really weird to me. Maybe because the padding is not even all the way around? But it feels like the sides are way too padded inside the terminal view itself.

danrabbit:
Yeah I can't see a way to pad the top and bottom without it affecting the scroll area which looks weird.

Maybe see if it grows on you. I could settle for 6px or 9px too. But the current padding is just so little

data/Application.css
```diff
@@ -26,8 +26,7 @@
 }
 
 vte-terminal {
-    padding-left: 4px;
-    padding-right: 4px;
+    padding: 0 12px;
```
lainsce, line 29: How about this? Makes it look more uniform and less lopsided. :smile: 

```suggestion
    padding: 12px 12px 0 12px;
```

cassidyjames:
Currently experimenting with `12px 6px` plus an undershoot.

Screenshot from 2020-08-10 12 13 10

data/Application.css
cassidyjames: So this still has the weird scrolling problem. I think just applying to the sides for now is fine unless/until we can sort out the undershoot.

cassidyjames: @danrabbit I think 0 and 12px feels too lopsided, so unless/until we can sort out some sort of undershoot or other way to handle that, how about `0 9px`? Still a notable improvement over master.

```suggestion
    padding: 0 9px;
```

cassidyjames (changes requested):
Let's try `0 9px` for now?

cassidyjames (approved):

PR #533, opened 2020-08-25
Ctrl+tab key bind: go to previous tab

### LINKED ISSUE #532: Enable "Ctrl+Tab" navigation between Tabs
## Prerequisites
- [ x ] I have searched open and closed issues for duplicates.

## Feature

Enable <kbd>ctrl</kbd>+<kbd>tab</kbd>  go to previous tab as is done on Files and on internet browser.

https://github.com/elementary/files/issues/389

rafaelcorsi:
closes #532

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -149,8 +149,9 @@ namespace Terminal {
             action_accelerators[ACTION_DUPLICATE_TAB] = "<Control><Shift>d";
             action_accelerators[ACTION_NEW_WINDOW] = "<Control><Shift>n";
             action_accelerators[ACTION_NEXT_TAB] = "<Control><Shift>Right";
+            action_accelerators[ACTION_NEXT_TAB] = "<Control>Tab";
             action_accelerators[ACTION_PREVIOUS_TAB] = "<Control><Shift>Left";
-            action_accelerators[ACTION_PREVIOUS_TAB] = "<Ctrl>Tab";
+            action_accelerators[ACTION_PREVIOUS_TAB] = "<Control><Shift>Tab";
             action_accelerators[ACTION_MOVE_TAB_RIGHT] = "<Control><Alt>Right";
             action_accelerators[ACTION_MOVE_TAB_LEFT] = "<Control><Alt>Left";
             action_accelerators[ACTION_ZOOM_DEFAULT_FONT] = "<Control>0";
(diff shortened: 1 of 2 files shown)


cassidyjames:
I feel like there may be a conflict with VTE stuff here? But I'm not sure. We should probably document what we consider reserved…

src/MainWindow.vala
```diff
@@ -150,6 +150,7 @@ namespace Terminal {
             action_accelerators[ACTION_NEW_WINDOW] = "<Control><Shift>n";
             action_accelerators[ACTION_NEXT_TAB] = "<Control><Shift>Right";
             action_accelerators[ACTION_PREVIOUS_TAB] = "<Control><Shift>Left";
+            action_accelerators[ACTION_PREVIOUS_TAB] = "<Ctrl>Tab";
```
jeremypw, line 153: Shouldn't `<Ctrl>Tab` navigate to *next* tab?  And `<Shift><Ctrl>Tab` navigate to the previous tab?  Problem is, Vte-91 captures `Tab` && `<Ctrl>Tab` for triggering autocompletion.

rafaelcorsi: Hi there,

1. I will change to *next tab* 
1. here it works nice and don't trig autocomplete 

Peek 2020-08-25 11-14

jeremypw:
`<Ctrl>Tab` does not *seem* to do anything different from `Tab` in Vte.Terminal, but I am not sure.  It is possible to implement tab navigation with `<Ctrl>Tab` but you have to use a key press handler to catch the event before it gets to the base Vte.Terminal - it doesn't work using the action accelerators.

src/MainWindow.vala
jeremypw: Hmm, you are right - cloning your branch it works - previously I just amended my own repo but I used "<Control>Tab" like the rest of the code.  Not sure why that should make a difference.

rafaelcorsi: > `<Ctrl>Tab` does not _seem_ to do anything different from `Tab` in Vte.Terminal, but I am not sure.

fix:

- [ x ] using `<Control>`
- [ x ] `<Control>Tab`  for next tab
- [ x ] `<Control><Shift>Tab` for previous tab

jeremypw (approved):
Works for me!

I have not found any documented use of `<Control><Tab>` in bash or vte so :shrug: I am going to approve.

PR #537, opened 2020-09-01
Only open `with_directory` mode when -w arg is not null

### LINKED ISSUE #499: Terminal is not remembering the last active tab
## Prerequisites

## Describe the bug
Terminal is not remembering the last active tab if you close while multiple tabs opened. It is focusing always on the first tab.

### To Reproduce
1. Open Terminal app.
2. Open 2 or 3 tab and select last tab
3. Close Terminal app without closing tabs.
4. Reopen Terminal app

### Expected behavior
When you open the Terminal app, the last active tab is expected to be remembered.

## Platform Information
```console
$ dpkg-query --show pantheon-terminal 
pantheon-terminal	5.5.1+r1640+pkg72~ubuntu5.1.2.1

$ head -2 /etc/os-release
NAME="elementary OS"
VERSION="5.1.2 Hera"
```

- [X] I'm using the latest released stable version

rafaelcorsi:
fix #499 
 
Starts application according to the `-w` argument.


jeremypw (changes requested):
This works, but I think we should really fix `start_terminal_with_terminal_directory (string? directory)` and/or the functions it calls so that they behave correctly with a `null` working directory.   Otherwise the parameter should be changed to `string directory` so that this function requires a non-null working directory.

rafaelcorsi:
> This works, but I think we should really fix `start_terminal_with_terminal_directory (string? directory)` and/or the functions it calls so that they behave correctly with a `null` working directory. Otherwise the parameter should be changed to `string directory` so that this function requires a non-null working directory.

Hi @jeremypw! I think I did what you suggest, but ould not check on master because `libhandy-1` . I checked on the last version before this dependence was introduced.

jeremypw:
@rafaelcorsi Yes, what I am doing is to develop on Ubuntu Focal, pulling things from the elementary focal repositories as required.  This is OK as long as your not working on styling issues.  Otherwise you probably need to install the prerelease version of elementary Odin on a VM.

jeremypw (changes requested):
The problem with this is that if you combine the `-w` flag with the `-n` flag, the new window reloads all the tabs from settings.  This is not the desired behaviour - the new window should only have a single tab.

jeremypw:
@rafaelcorsi Sorry for the slow review.  Are you still able to work on this?

jeremypw:
Closing as no response to last question.

PR #540, opened 2020-09-01
keep terminal focus when tab is closed

### LINKED ISSUE #495: Closing a tab that is not selected makes terminal loose focus.
1- Open two tabs.
2- Close the first tab without selecting it.

**Result:** The text cursor stops flashing and the terminal looses focus. This doesn't happen if an user selects a tab before closing it.

rafaelcorsi:
fix #495

keep focus on the terminal when a tab is closed.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -802,6 +802,7 @@ namespace Terminal {
                 }
             }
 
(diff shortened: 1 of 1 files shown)


jeremypw (approved):
Thanks for this! Works as expected.

PR #542, opened 2020-09-12
Add custom color schemes

### LINKED ISSUE #418: User-provided color scheme

sebastianlay:
This thing fixes #418.

The implementation is based on the recommendations by @danrabbit in the corresponding issue:

- [x] Doesn't require hunting around in a preferences dialog. We can add a ( + ) add the end of the list of color options in the popover we have currently. This could then throw a dialog that specifically and only manages the stored color schemes.
- [x] Includes a preview so we can see what the resulting color scheme actually looks like
_This is not actually needed since the selected theme is applied immediately to the opened console._
- [x] Labels the color selectors (of which it looks like only Tilix tries to do this)
- [x] If we're really smart, we can do some color contrast checking and warn about color schemes that would be illegible. At the very least we should do this between the background and foreground color and make sure it passes the base A level

Additionally this implementation is completely backwards compatible in that it only manipulates the already existing settings and also recognizes the existing default configurations. This has the added benefit that third-party tools and scripts also continue to work.

This also includes an excessive number of predefined themes (194 to be exact) which were all taken from https://github.com/Mayccoll/Gogh/ and the color contrast check uses the code from https://github.com/danrabbit/harvey/ 

### Open questions / calls for improvements:

- The ComboBox for for choosing a theme is pretty long and it can take a while to find the desired theme. Is there a better suited widget that allows keyboard navigation an/or search and/or autocompletion?
- The already existing color buttons modify the window theme and the console theme at the same time. Should it stay this way since there is now a way to set these two themes independently of each other?
- The data structure to hold the predefined themes in the `Themes.vala` is currently an ArrayList. I am pretty new to Vala so my question is whether there is a better data structure for this (something like a dictionary in C#, ideally with a `find` method) and a better way to initialize this (again something like collection initializers in C#)?

### Screenshots:

The fourth button in the color menu that opens the color preference dialog:

high-contrast

The open color preference dialog with a default theme (and good contrast):

default-dark

A theme selected that has poor contrast:

one-dark


src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+
+            var colors_grid = new Gtk.Grid () {
+                column_spacing = 12,
+                row_spacing = 6,
+                margin = 9,
+                margin_top = 0,
+                halign = Gtk.Align.CENTER
+            };
+
+            var window_theme_label = new SettingsLabel (_("Window Theme"));
```
danrabbit, line 77: We have prior art for using the word "Style" instead of "Theme" when referring to the dark and light styles. I'm not sure if we have anything for referring to the app chrome specifically. Maybe @cassidyjames has suggestions here.

These labels should be sentence case as suffixed with ":"
```suggestion
            var window_theme_label = new SettingsLabel (_("Window style:"));
```

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+
+            add_action_widget (close_button, 0);
+        }
+
+        private Gtk.Widget get_colors_box () {
+
+            var colors_grid = new Gtk.Grid () {
+                column_spacing = 12,
+                row_spacing = 6,
+                margin = 9,
```
danrabbit, line 72: This should be 6 to align it with the action area and bottom margin of 18 so that we have a total of 24px (double space) between the content and action areas:
```suggestion
                margin = 6,
                margin_bottom = 18,
```

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+            var red_color_label = new SettingsLabel (_("Red"));
+            var green_color_label = new SettingsLabel (_("Green"));
+            var yellow_color_label = new SettingsLabel (_("Yellow"));
+            var blue_color_label = new SettingsLabel (_("Blue"));
+            var magenta_color_label = new SettingsLabel (_("Magenta"));
+            var cyan_color_label = new SettingsLabel (_("Cyan"));
+            var grey_color_label = new SettingsLabel (_("Grey"));
+            var background_label = new SettingsLabel (_("Background"));
+            var foreground_label = new SettingsLabel (_("Foreground"));
+            var cursor_label = new SettingsLabel (_("Cursor"));
```
danrabbit, line 111: These should all be suffixed with ":" and the more common en_US spelling is "Gray"
```suggestion
            var black_color_label = new SettingsLabel (_("Black:"));
            var red_color_label = new SettingsLabel (_("Red:"));
            var green_color_label = new SettingsLabel (_("Green:"));
            var yellow_color_label = new SettingsLabel (_("Yellow:"));
            var blue_color_label = new SettingsLabel (_("Blue:"));
            var magenta_color_label = new SettingsLabel (_("Magenta:"));
            var cyan_color_label = new SettingsLabel (_("Cyan:"));
            var grey_color_label = new SettingsLabel (_("Gray:"));
            var background_label = new SettingsLabel (_("Background:"));
            var foreground_label = new SettingsLabel (_("Foreground:"));
            var cursor_label = new SettingsLabel (_("Cursor:"));
```

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+            contrast_image = new Gtk.Image.from_icon_name ("process-completed", Gtk.IconSize.LARGE_TOOLBAR) {
+                halign = Gtk.Align.START
+            };
+            var contrast_bottom_label = new Gtk.Label ("┘") {
+                halign = Gtk.Align.START,
+                margin_start = 9
+            };
+            contrast_box.pack_start (contrast_top_label);
+            contrast_box.pack_start (contrast_image);
+            contrast_box.pack_start (contrast_bottom_label);
```
danrabbit, line 169: It seems like it would be easier to do:
```suggestion
            var contrast_top_label = new Gtk.Label ("┐");
            contrast_image = new Gtk.Image.from_icon_name ("process-completed", Gtk.IconSize.LARGE_TOOLBAR);
            var contrast_bottom_label = new Gtk.Label ("┘");
            
            var contrast_box = new Gtk.Box (Gtk.Orientation.VERTICAL, 3) {
                  halign = Gtk.Align.START
            };
            contrast_box.pack_start (contrast_top_label);
            contrast_box.pack_start (contrast_image);
            contrast_box.pack_start (contrast_bottom_label);
```

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+                contrast_image.tooltip_text = _("Contrast is very low.");
+            } else if (contrast_ratio < 4.5) {
+                contrast_image.icon_name = "dialog-warning";
+                contrast_image.tooltip_text = _("Contrast is low.");
+            } else if (contrast_ratio < 7) {
+                contrast_image.icon_name = "process-completed";
+                contrast_image.tooltip_text = _("Contrast is high.");
+            } else {
+                contrast_image.icon_name = "process-completed";
+                contrast_image.tooltip_text = _("Contrast is very high.");
```
danrabbit, line 320: tooltips shouldn't have punctuation
```suggestion
                contrast_image.tooltip_text = _("Contrast is very low");
            } else if (contrast_ratio < 4.5) {
                contrast_image.icon_name = "dialog-warning";
                contrast_image.tooltip_text = _("Contrast is low");
            } else if (contrast_ratio < 7) {
                contrast_image.icon_name = "process-completed";
                contrast_image.tooltip_text = _("Contrast is high");
            } else {
                contrast_image.icon_name = "process-completed";
                contrast_image.tooltip_text = _("Contrast is very high");
```

src/MainWindow.vala
```diff
@@ -538,34 +548,37 @@ namespace Terminal {
...
-                Application.settings.set_string ("foreground", DARK_FG);
+                Terminal.Themes.set_active_name ("Default (Dark)");
+                if (color_preferences_dialog != null) {
+                    color_preferences_dialog.update_widgets_from_settings ();
+                }
+                return Gdk.EVENT_PROPAGATE;
             });
 
-            color_button_light.clicked.connect (() => {
+            color_button_light.button_press_event.connect (() => {
```
danrabbit, line 560: Is there a reason for changing these from clicked to button press event?

danrabbit (changes requested):
Hey thanks for working on this! This is looking really good and it's definitely a highly requested feature so I appreciate you tackling it.

I left some inline comments with some minor code cleanups and copy changes alignment etc.

I agree that the long combobox here is pretty unweildy. Perhaps instead of trying to include as many styles as we can find, it would be better to have some kind of import mechanism and stick with the same 3 default styles for now. Perhaps in the interest of getting this feature merged more quickly, we should reduce the scope of this branch to only supporting a single custom style and we can expand in later follow-up branches

I think a major interaction issue we have here is that if you ever switch away to another style, your custom style gets wiped. This seems like it would be pretty frustrating for someone who switches between styles. I imagine we should probably persist custom styles in gsettings. Maybe with a "styles" child schema, even if we only support saving one custom style for now

Yes I think we want to maintain the behavior of a style also setting the window chrome to dark or light. The quick style switcher should make it really easy to alternate between styles without having to fiddle with individual settings every time

data/Application.css
```diff
@@ -50,3 +50,13 @@ vte-terminal {
     background-color: #fff;
     color: #000;
 }
+
+.color-custom radio {
+    background-color: #fff;
+    color: #000;
+    -gtk-icon-source: -gtk-icontheme("list-add");
```
danrabbit, line 57: We should make sure to use the -symbolic variant here
```suggestion
    -gtk-icon-source: -gtk-icontheme("list-add-symbolic");
```

data/Application.css
```diff
@@ -50,3 +50,13 @@ vte-terminal {
     background-color: #fff;
     color: #000;
 }
+
+.color-custom radio {
+    background-color: #fff;
+    color: #000;
```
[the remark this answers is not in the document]
sebastianlay: Yeah, I used it since there does not seem to be a dark style for the "list-add" icon. But since we are switching to "list-add-symbolic" anyways this is not an issue anymore.

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+        private Gtk.ColorButton color07_button; // Cyan
+        private Gtk.ColorButton color08_button; // Light gray
+        private Gtk.ColorButton color09_button; // Dark gray
+        private Gtk.ColorButton color10_button; // Light Red
+        private Gtk.ColorButton color11_button; // Light Green
+        private Gtk.ColorButton color12_button; // Light Yellow
+        private Gtk.ColorButton color13_button; // Light Blue
+        private Gtk.ColorButton color14_button; // Light Magenta
+        private Gtk.ColorButton color15_button; // Light Cyan
+        private Gtk.ColorButton color16_button; // White
```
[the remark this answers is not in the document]
sebastianlay: This is matching the naming in Gogh. I think it makes sense since it is generally shorter and there are a number of themes where for example "blue" does not actually correspond to a blue color.

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,403 @@
...
+            );
+        }
+
+        construct {
+            get_content_area ().add (get_colors_box ());
+
+            var close_button = new Gtk.Button.with_label (_("Close"));
+            close_button.clicked.connect (destroy);
+
+            add_action_widget (close_button, 0);
```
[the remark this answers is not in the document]
sebastianlay: The more you know!

src/Dialogs/ColorPreferencesDialog.vala
sebastianlay: When deciding on grey/grey I used the same as Tilix. But I changed it now to american spelling :)

src/Dialogs/ColorPreferencesDialog.vala
sebastianlay: Definitely.

src/MainWindow.vala
sebastianlay: Actually yes :) I noticed that the `clicked` event for radio buttons is triggered once when the button is activated and (more surprising) once when the button is deactivated. This prevented the updating of the radio buttons from the color preferences dialog. This wasn't an issue until now since the order of the activating/deactivating of the different buttons was just right for this to work.

data/Application.css
sebastianlay: Could you elaborate on when to use which variant? Is there some documentation on this somewhere?

sebastianlay:
Hey! Thanks a lot for your review. I implemented almost all of your suggestions, see my comments above. 

> I agree that the long combobox here is pretty unweildy. Perhaps instead of trying to include as many styles as we can find, it would be better to have some kind of import mechanism and stick with the same 3 default styles for now. Perhaps in the interest of getting this feature merged more quickly, we should reduce the scope of this branch to only supporting a single custom style and we can expand in later follow-up branches

I would _really_ like to include at least some predefined styles. I am pretty sure that many people would like to customize their terminal at least a little bit. On the other hand I assume that not all of these people (myself included) would put in the effort to enter an existing style color by color or even create a completely custom style from the ground up. Maybe reducing the number of styles to the 10-20 most popular (by some measure) styles would be a good compromise?

> I think a major interaction issue we have here is that if you ever switch away to another style, your custom style gets wiped. This seems like it would be pretty frustrating for someone who switches between styles. I imagine we should probably persist custom styles in gsettings. Maybe with a "styles" child schema, even if we only support saving one custom style for now

Yeah, I am a bit torn about that. My thinking was that most people are not creating styles from the ground up and instead start with one of the existing styles and apply some minor tweaks to their liking. The way it is currently implemented allows that while also allowing the creation of a completely new style. I am not sure how we would allow both use cases if we would save the custom style. I am open to ideas!

data/Application.css
danrabbit: Basically the `-symbolic` variant will follow the text color whereas the normal variants are full color icons. It's only kind of a historical coincidence that the normal variant for `list-add` happens to be close to the text color in the light style. This will change in a future revision of the icon set. So in this context where we want a simple shape icon that follows the UI text color, `symbolic` is the way to go

src/MainWindow.vala
danrabbit: That makes sense to me!

danrabbit:
@sebastianlay Hey sorry it took a while to follow up with a review here! This is looking really good. I think we can probably just try to get something in and iterate on it as we go.

> I would really like to include at least some predefined styles
At the very least we need to make sure we actually have permission to redistribute any inluded styles. It doesn't look like the `Gogh` repo contains any licensing information and several of these styles do not have a license, so they should be removed unless you'd like to contact the authors and ask them to publish a license. There are also several styles included that don't pass the contrast check, so I imagine we should remove those as well.

sebastianlay:
@danrabbit Hey, you are right. I trimmed the list down to 24 styles (including our three default and the one custom) and checked the licenses for the remaining ones:

- 3024 | https://github.com/idleberg/3024.tmTheme | Public Domain
- Dracula | https://github.com/dracula/dracula-theme | MIT
- Gruvbox | https://github.com/morhetz/gruvbox | MIT
- Monokai | https://github.com/stephenway/monokai.terminal | MIT
- Nord | https://github.com/arcticicestudio/nord | MIT
- Solarized | https://github.com/altercation/solarized | MIT
- Spacegray | https://github.com/kkga/spacegray | MIT
- Summerfruit | https://github.com/cscorley/base16-summerfruit-scheme | MIT
- Tomorrow | https://github.com/chriskempson/tomorrow-theme | MIT

If I understand it correctly, the MIT license should be compatible with the LGPL from this repo here, right? I am just not sure what the right place for the above list would be. Some sort of CREDITS file?

Conan-Kudo:
For vendoring third-party styles into this codebase, a reference to the style, the commit it came from, and a copy of the license needs to be included for each vendored style.

You probably will want to make those styles able to be read from configuration files that are installed into somewhere in `/usr/share` so that the appropriate documentation can be included along with the style configurations.

Conan-Kudo:
To give a bit more detail of how _I_ suggest doing this:

* Break out the theme configurations into something like `/usr/share/elementary-terminal/themes/<theme>/` directory
* In that directory, a `theme.toml` and `LICENSE` file would be present.
* The `theme.toml` file would include the actual style configuration and fields indicating where it was sourced from if it comes from somewhere else.
* The `LICENSE` file would be a verbatim copy of the license text from the original source. If the style is unique to this codebase, then the second file can be omitted and the `theme.toml` can indicate that it's an original theme.

danrabbit:
Hey @sebastianlay, since this seems to have gotten a bit stuck on the licensing bit would you be interested in merging a version of this without the color schemes so we can get this feature in and then bring back some pre-loaded color schemes in a future branch?

sebastianlay:
> Hey @sebastianlay, since this seems to have gotten a bit stuck on the licensing bit would you be interested in merging a version of this without the color schemes so we can get this feature in and then bring back some pre-loaded color schemes in a future branch?

That sounds like a good idea. I will give the suggestions by @Conan-Kudo a shot and if it takes me too long I will reduce the merge request to the custom color scheme.

sebastianlay:
Okay, I will close this in favor of #552

PR #546, opened 2020-11-03
MainWindow: Add “follow system style” toggle

cassidyjames:
As discussed in a call recently. We probably want this to be off by default since Terminal defaults to the dark style. But toggling it on hides the color buttons and should follow the system style.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -401,6 +401,23 @@ namespace Terminal {
             font_size_grid.add (zoom_default_button);
             font_size_grid.add (zoom_in_button);
 
+            var follow_system_label = new Gtk.Label (_("Follow System Style"));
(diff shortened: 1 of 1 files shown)


cassidyjames:
The way to save and restore color schemes was pretty wild to follow, but this works within that—probably in a terrible way—to get this working at least. The behavior can be considered, and then we should probably clean a lot of the color scheme stuff up…

cassidyjames:
Ah found some more bad things that I need to fix. :sweat_smile:

cassidyjames:
I think this could use some refactoring still to be more clear, but it's _almost_ fully-functional. The one case I am still not handling is when `follow-system-style` is off and the style is mis-matched with the system, then `follow-system-style` is toggled on; for some reason, the background of the VTE is not updated to match.

PR #547, opened 2020-11-14
MainWindow: Use the Gtk.EventControllerKey controller to handle the key events

tintou:
This makes the copy function work again.
Fix GtkMenu warnings on the go.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -60,7 +60,7 @@ executable(
         dependency('gee-0.8'),
         dependency('glib-2.0'),
         dependency('gobject-2.0'),
-        dependency('gtk+-3.0'),
+        dependency('gtk+-3.0', version: '>=3.24'),
         dependency('granite', version: '>=5.5.0'),
         dependency('libhandy-1', version: '>=0.83.0'),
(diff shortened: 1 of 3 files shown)


jeremypw:
Hm, Gtk.EventControllerKey could be useful in other projects (e.g. Files) as well I think :thinking:

tintou:
@jeremypw yes, it is the future way to handle events and has been backported to GTK3 so that the transition to GTK4 can be prepared ( https://developer.gnome.org/gtk4/unstable/gtk-migrating-3-to-4.html#id-1.7.4.3.14 )

jeremypw (changes requested):
This seems to break natural copy - `<Ctrl>C` results in `^C` appearing in the terminal.  Not immediately obvious why as you have implemented similar code.

jeremypw:
AFAICT, the key press event is not actually reaching the EventControllerKey.key_pressed controller :disappointed:

tintou:
It indeed only work if the widget has focus which isn't the case here, closing in favor of another solution.

PR #548, opened 2020-11-14
TerminalWidget: fix copy not working on the first try

tintou:
MainWindow: Simplify the flags check


jeremypw:
I notice that the first menu item gets selected when opened with secondary-click but not when opened with the menu key. I think this should be consistent?

tintou:
@jeremypw indeed, applied there too

jeremypw (approved):
Does the job :smile:

PR #552, opened 2020-12-11
Danrabbit/custom colors

danrabbit:
Todos:
- [x] convert themes to a hashtable and get rid of the theme struct. This will make lookups simpler
- [ ] Persist the custom theme values in gsettings
- [ ] Avoid a lot of extra work combining and splitting strings
- [x] Save theme name in gsettings so we're not relying on looking up the palette and trying to make a match
- [x] Recalculate contrast only when background and foreground values change


jeremypw:
This would probably also affect PRs #546 and #630.

danrabbit:
Yeah if someone else wants to pick this up, feel free! I just haven't been able to revisit it for a while.

Since this is a pretty big branch, it probably makes sense to try to break it up into smaller cleanups that can be merged before adding full color scheme support

jeremypw:
@danrabbit Can this PR be closed now?

PR #556, opened 2020-12-27
Adds setting "block-multiline-paste-alert" (default: true)

etkaar:
Adds setting "block-multiline-paste-alert" (default: true) which opens a dialog if the user attempts to paste multiline content into the terminal, since newline characters could lead to an unintentionally command execution and harm the system.

**Example**

Someone would copy a `rm -R ...` command from a tutorial (or from the terminal itself) with the intention to edit it _after_ the paste. However, if the paste contains newline characters, it would be **unintentionally executed**. The new warning prevents this.

**Setting**

You can turn on/off this setting using:

```
gsettings set io.elementary.terminal.settings block-multiline-paste-alert true
gsettings set io.elementary.terminal.settings block-multiline-paste-alert false
```

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1187,6 +1187,20 @@ namespace Terminal {
         }
 
         private void on_get_text (Gtk.Clipboard board, string? intext) {
+            /* don't accidentally cause command executions from pasting multiline text */
+            if (Application.settings.get_boolean ("block-multiline-paste-alert")) {
+				
+				if (intext.index_of ("\n") != -1) {
+					var d = new BlockMultilinePasteDialog (this);
+                    if (d.run () == 1) {
+                        d.destroy ();
+                        return;
+                    }
+                    d.destroy ();
+				}
+				
+			}
+			            
             /* if unsafe paste alert is enabled, show dialog */
             if (Application.settings.get_boolean ("unsafe-paste-alert") && !unsafe_ignored ) {
 
--- a/data/io.elementary.terminal.gschema.xml
+++ b/data/io.elementary.terminal.gschema.xml
@@ -179,5 +179,10 @@
       <summary>Alert the user about unsafe paste.</summary>
       <description>If the user paste's a command into the terminal that might be considered unsafe, warn them.</description>
     </key>
+    <key name="block-multiline-paste-alert" type="b">
+      <default>true</default>
+      <summary>Alert the user about multiline paste.</summary>
+      <description>If the user pastes text into the terminal which contains newlines, warn them, since it could lead to an unintentionally command execution.</description>
+    </key>
   </schema>
 </schemalist>
--- a/meson.build
+++ b/meson.build
@@ -51,6 +51,7 @@ executable(
     'src/Application.vala',
     'src/DBus.vala',
     'src/MainWindow.vala',
+    'src/Dialogs/BlockMultilinePasteDialog.vala',
     'src/Dialogs/ForegroundProcessDialog.vala',
     'src/Dialogs/UnsafePasteDialog.vala',
     'src/Widgets/SearchToolbar.vala',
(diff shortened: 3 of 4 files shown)


cassidyjames (changes requested):
Left one copy change, but it makes me wonder: should this just be part of the UnsafePasteDialog? Much of the copy and UI would be the same; the only difference would be the title and possibly the setting we bind to, if we want it to be a separate setting.

etkaar:
@cassidyjames The problem is that I fear that people (like me) would permanently disable the UnsafePasteDialog for the `sudo ...` paste, but then the warning for the multiline pastes would be disabled too, which may be unwanted. I think in the most cases you want to paste your sudo command, but not a multiline command, so it makes sense to seperate it.

Regarding the warning message: Since a newline on the end alone would lead to an execution (not necessarily of multiple commands), what do you think of using **The pasted text contains newline characters and may be unintentionally executed** instead?

etkaar:
@cassidyjames I did now commit your suggestion regarding the text.

src/Dialogs/BlockMultilinePasteDialog.vala
```diff
@@ -0,0 +1,50 @@
+/*
+* Copyright 2011-2018 elementary, Inc. (https://elementary.io)
```
jeremypw, line 2: Needs updating

src/Dialogs/BlockMultilinePasteDialog.vala
```diff
@@ -0,0 +1,50 @@
...
+            buttons: Gtk.ButtonsType.NONE,
+            transient_for: parent
+        );
+    }
+
+    construct {
+        image_icon = new ThemedIcon ("dialog-warning");
+        primary_text = _("The pasted text may execute multiple commands");
+
+        secondary_text =
```
jeremypw, line 31: Shouldn't construct translatable text like this, if possible.  Suggest using `printf`.  Also there is a grammar error.
```
        secondary_text =  "%s\n%s".printf (
              _("The text you want to paste may be treated as a command and could unintentionally be executed."),
              _("Be sure your paste doesn't contain newline characters.")
        );
```

src/Dialogs/BlockMultilinePasteDialog.vala
```diff
@@ -0,0 +1,50 @@
...
+            _("The text you want to paste may be treated as command and could unintentionally be executed.") + " " +
+            _("Be sure your paste doesn't contain newline characters.");
+
+        var show_protection_warnings = new Gtk.CheckButton.with_label (_("Show paste protection warnings"));
```
jeremypw, line 35: Distinguish from unsafe paste alert setting.
`"Show multi-line paste warning"`

src/MainWindow.vala
```diff
@@ -1217,6 +1217,20 @@ namespace Terminal {
         }
 
         private void on_get_text (Gtk.Clipboard board, string? intext) {
+            /* don't accidentally cause command executions from pasting multiline text */
+            if (Application.settings.get_boolean ("block-multiline-paste-alert")) {
+
```
jeremypw, line 1222: The two if statements can be merged.
```
if (Application.settings.get_boolean ("block-multiline-paste-alert") &&
    intext.contains ("\n")) {

        var d = ......
```

jeremypw (changes requested):
Made a few suggestions but it would be good to progress this if possible - thanks.

jeremypw:
Failing lint test - install io.elementary.vala-lint  from https://github.com/vala-lang/vala-lint and run `io.elementary.vala-lint -d <root folder>` to locate the problems.

etkaar:
You're free to take over this PR, unfortunately it is unlikely that I have time to fix that this year.

jeremypw:
@etkaar  OK, thanks for letting us know. I'll try and finish it off sometime.  Thanks for the contribution.
 :smile:

cassidyjames:
@etkaar @jeremypw I've created a PR based on this branch at https://github.com/elementary/terminal/pull/595 that does the same thing while keeping them under one setting. I can see a small reason to keep them separate, but I think the intent was always to group various "unsafe" pastes under one setting—especially if we want to add a GUI setting for it later.

PR #557, opened 2021-01-03
Adds context menu option "Paste Selected"

### LINKED ISSUE #553: Select > Right click > Copy and Paste
In PuTTY you can paste selected text from the terminal into the command line just by selecting it and right clicking, which can be very helpful sometimes. For elementary/terminal a new option "Copy and Paste" for the right-click context menu would be helpful.

etkaar:
Closes #553

Adds context menu option "Paste Selected" (`Shift+Ctrl+X`) to easily copy and paste selected text from terminal into the terminal (similar to PuTTYs select & right click to paste).

**Note**
I did not touch the translation files in this PL because I think this is exclusively done using [Weblate](https://l10n.elementary.io/projects/terminal/terminal/).

Screenshot from 2021-01-03 08 46 30


jeremypw:
Need input from design team as to whether the value of this function justifies lengthening the menu. (Note, the same functionality can be achieved by pressing `<Shift><Ctrl>C` + `<Shift><Ctrl>V`).

etkaar:
> (Note, the same functionality can be achieved by pressing `<Shift><Ctrl>C` + `<Shift><Ctrl>V`).

Unfortunately this is not fast enough. However, what I would really prefer would be `<Ctrl><RightClick>` (or something similar) instead of `<Shift><Ctrl>X`, but I was not sure if that is possible.

jeremypw:
Also `<Shift><Ctrl>X` might be expected to cut the selected text (although this could only work on an uncompleted command)

jeremypw (commented):
Needs input from the design team as to whether the option is needed and what shortcut is appropriate

andreasfelix:
> Unfortunately this is not fast enough. However, what I would really prefer would be `<Ctrl><RightClick>` (or something similar) instead of `<Shift><Ctrl>X`, but I was not sure if that is possible.

There is already middle-click-paste to paste with current selection with the mouse (can be enabled under *Settings -> Mouse & Touchpad -> Clicking -> Middle Click Paste*). This works in all applications, not only terminal. 

There is also the <kbd>Shift</kbd> <kbd>Insert</kbd> shortcut which pastes the current selection. I think this is implemented by VTE.

jeremypw:
I can confirm `<Shift>Insert` works in Terminal to paste the current selection into the commandline.

@etkaar Are you content for this PR to be closed in view of the above comments?

etkaar:
Yes, the PR can be closed :)

PR #563, opened 2021-03-05
Add text to paste in details pane for UnsafePasteDialog.

### LINKED ISSUE #562: Show clipboard contents in paste protection dialog
## Problem

I just went to past a string into Terminal and paste protection popped up. I was confused because the string I was pasting was not a command, but a branch name. Why is this dialog appearing? How weird. The string was `fix-sudo-extraneous-output` ;p

## Proposal

It would probably be helpful to show the string that will be pasted in the paste protection dialog so that the user can review it before deciding to continue or cancel

igordsm:
Fix #562 .


tintou (approved):

tintou:
Thanks for this!

PR #566, opened 2021-03-12
Use Granite.SwitchModelButton

pongloongyeat:
As title implies. There's some other deprecations that I'd like to check out but perhaps in a separate PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A super lightweight, beautiful, and simple terminal. Comes with sane defaults, b
 ## Building, Testing, and Installation
 
 You'll need the following dependencies:
-* libgranite-dev >= 5.3.0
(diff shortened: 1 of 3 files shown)


pongloongyeat:
Could anyone comment on whether or not the builds were supposed to pass before commit 4b544124609d3c156f09d9d6ac1643bd447afc86 (Granite version bump)? I'm using `Granite.SwitchModelButton` which should require minimum 6.0.0 but the build still passed on `Granite` 5.5.0.

danrabbit (changes requested):
Works as expected. Nice to simplify all this!

Can you bump the required version in the README as well? :)

pongloongyeat:
@danrabbit it's done!

danrabbit (approved):
Great! Thanks again

PR #570, opened 2021-03-28
Add Reload item to terminal tabs.

### LINKED ISSUE #544: Add option to reload tab/restart shell
## Prerequisites

## Feature
**Is your feature request related to a problem? Please describe.**
When installing new applications, changing config files or if one just wants to reset the current shell environment, it is necessary open a new shell. Currently the fastest way to achieve this is to duplicate the current tab and close the old one.

**Describe the solution you'd like**
Add a *Reload Tab* or *Restart Shell* option to the context menu of a tab.

Slice 1

**Describe alternatives you've considered**
Source config file, but this does not remove old config, does not reset the environment and can lead to duplicated paths in the `$PATH` variable:

```sh
. ~/.bashrc
```

Clear cache for program locations, but does not do anything else:
```sh
hash -r
```
Almost the same as opening a new tab, but does also not produce the same environment [(see here)](https://medium.com/@rajsek/zsh-bash-startup-files-loading-order-bashrc-zshrc-etc-e30045652f2e):
```sh
exec sudo --login --user $USER
# or
exec sudo -u $USER bash
```

igordsm:
Fix #544 . The solution adopted was to kill the current shell child process and create a new one using the same terminal. 

terminal-reload

I'm not sure if the Reload option was added in the correct place. It seems all exiting options come from DynamicNotebook, so I just added an extra option after a tab is created.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1135,6 +1135,11 @@ namespace Terminal {
             term.tab = tab;
             tab.ellipsize_mode = Pango.EllipsizeMode.START;
 
+            var reload_menu_item = new Gtk.MenuItem.with_label (_("Reload"));
+            tab.menu.append (reload_menu_item);
+            reload_menu_item.activate.connect (term.reload);
+            tab.menu.show_all ();
+
             return tab;
         }
 
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -592,5 +592,12 @@ namespace Terminal {
         public bool has_output () {
             return !resized && get_last_output ().length > 0;
         }
+
+        public void reload () {
+            var old_loc = get_shell_location ();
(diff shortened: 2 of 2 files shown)


jeremypw:
I would question the term "Reload" here as that implies that the current state is just refreshed - whereas here the current state is destroyed and a new clean state replaces it.   Also as this could be a destructive action perhaps some warning is required - especially if a foreground process is present.  Needs UX team input though.

jeremypw:
"Restart" or "Reset" sounds better to me but its up to the UX team.  Maybe adding " ... Shell" would make it clearer but that might confuse a non-technical user I guess.

jeremypw:
I am inclined to merge this now as is and iterate later if required. The term "Reload" was suggested by one of the UX team.  A warning dialog might be useful if reloading a shell will kill an ongoing process.

jeremypw (approved):

Issue #541, opened 2020-09-06
Flatpak package


minioin:
Hi, is this terminal available as flatpak. I use Fedora with GNOME and wanted to try this one as I've heard good things about it.

If not, can this be done? It can be packaged with full privilege at first and gradually drop them in future releases.

jeremypw:
I guess we just need someone to maintain the flatpak ....

cassidyjames:
It will likely eventually be available in the AppCenter Flatpak remote once that happens, but I don't think we can commit to an official Flatpak hosted anywhere other than that. I'm also not 100% sure how a Terminal in a Flatpak would work, other than just being allowed to break through the sandbox to access everything. :)

cassidyjames:
I guess I can re-open this to be more specific; we're open to maintaining a Flatpak manifest with the intent of shipping it as a Flatpak in elementary OS. So that would be welcome, at least. And it might be useful to others.

PR #571, opened 2021-03-29
Preliminary Flatpak manifest

elsiehupp:
Fixes https://github.com/elementary/terminal/issues/541

On a preliminary basis, this appears to work correctly.

You can test the build using the command:
```
$ flatpak-builder --force-clean build io.elementary.terminal.json
```
And then build and install using the command:
```
$ sudo flatpak-builder --install --force-clean build io.elementary.terminal.json
```

The main obvious build issues seem to be:
1. `open-pantheon-terminal-here.desktop` doesn't work because it's a non-allowed export filename, and
2. The exported icons in general (but mainly the custom app icon) don't work because they also have non-allowed export filenames.

I used `org.gnome.Platform` and `org.gnome.Sdk` because they include certain dependencies that you would otherwise be building from scratch, but one of the other of them is probably unnecessary, and [`io.elementary.BaseApp`](https://github.com/flathub/io.elementary.BaseApp), might help, as well.

### DIFF
--- a/io.elementary.terminal.appdata.xml
+++ b/io.elementary.terminal.appdata.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="application">
+    <id>io.elementary.terminal</id>
+    <metadata_license>FSFAP</metadata_license>
+    <project_license>GPL-3.0-only</project_license>
+    <name>elementary Terminal</name>
+    <summary>The terminal of the 21st century.</summary>
+    <description>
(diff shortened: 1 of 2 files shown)


elsiehupp:
Yes. This was more of a proof of concept than anything else, and I didn’t really test it thoroughly at the time. Mainly I used it as an opportunity to learn how to do Flatpak packaging. FWIW you could comment on Issue #541, as well, since that was my impetus for doing this. Also feel free to close this PR if it doesn’t seem like something worth pursuing (considering the limitations you’ve explained above).

PR #575, opened 2021-04-10
Catch Uri.unescape_string returning null

tom95:
Not checking this can lead to crashes when a string cannot be escaped, for example if the string is the contents of an open Vim window. As such, when doing "select all" while having Vim open would crash the terminal as it tried to check if the new selection contains a link.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1258,11 +1258,13 @@ namespace Terminal {
 
         private void action_open_in_browser () {
             get_current_selection_link_or_pwd ((clipboard, uri) => {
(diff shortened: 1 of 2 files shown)


jeremypw:
You need to change the return type of the function to `string?` and check all callers of this function deal with possible null return properly (if you haven't already done so).

tom95:
Fair point. I had already checked both call sites, one was already doing a null check, the other could only be reached if there was a valid link under the cursor, but I went ahead and added a null check in the 2nd place as well to make sure valac agrees, too :)

jeremypw (approved):
LGTM - thanks!

PR #578, opened 2021-05-19
Add shortcut for reload tab action

andreasfelix:
Adds <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>R</kbd> as shortcut for the new reload tab action (#544, #570)


jeremypw:
What do you feel about supporting `F5` as a shortcut for this (cf Files and Epiphany)?

andreasfelix:
Sure, if the symmetric with Files and Epiphany is intended, adding <kbd>F5</kbd> as alternative shortcut seams reasonable. I included it with the last commit. But this should probably not up to me to decide that :D

I still find the symmetric somewhat misleading, because Files and Epiphany update/reload a given resource, while this action resets/restarts the shell.

On concern is also that this action/shortcut could be misused to "just clear" the terminal screen.

jeremypw:
Yes, I see what you are saying - that is why I was unsure about using "Reload" for the menuitem :thinking:

jeremypw:
The current shortcut `<Ctrl><Shift>R` is very similar to `<Ctrl>R` used in Files for the "Refresh" action so there could still be confusion.

jeremypw:
What about `<Ctrl>D` - this is used to exit the current shell anyway except it does not work when a process is running.

andreasfelix:
As far as I understand bash interprets <kbd>Ctrl</kbd><kbd>D</kbd> as the forth Unicode character End of Transmission (because **D** is the forth letter in the alphabet). I am not sure if this counts as a "real" shortcut. And in other command line programs like "cat" or "netcat" it can also be used to end the current input, but it does not exit the process. So I think <kbd>Ctrl</kbd><kbd>D</kbd> cannot be thought of as a universal exit shortcut.

In my opinion the <kbd>Shift</kbd><kbd>Ctrl</kbd><kbd>R</kbd> and <kbd>F5</kbd> shortcuts are fine, as it is conceptually somewhat similarity with reloading the browser or refreshing Files. IMO only the word "Reload" is misleading here.

But this is of course up to the @elementary/ux to decide that.

jeremypw:
OK, let's wait for their input.

danrabbit:
This looks like a 0 diff since we merged in the other branch with accel labels that included this

PR #579, opened 2021-05-19
Add dialog to reload tab action if there is a foreground process

andreasfelix:
Related to #570, #578

### DIFF
--- a/src/Dialogs/ForegroundProcessDialog.vala
+++ b/src/Dialogs/ForegroundProcessDialog.vala
@@ -44,6 +44,18 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
         );
     }
 
+    public ForegroundProcessDialog.before_tab_reload (MainWindow parent) {
+        Object (
+            transient_for: parent,
+            primary_text: _("Are you sure you want to reload this tab?"),
+            secondary_text:
+                _("There is an active process on this tab.") + " " +
+                _("If you reload it, the process will end."),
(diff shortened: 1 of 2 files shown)


src/Dialogs/ForegroundProcessDialog.vala
```diff
@@ -44,6 +44,18 @@ public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
         );
     }
 
+    public ForegroundProcessDialog.before_tab_reload (MainWindow parent) {
+        Object (
+            transient_for: parent,
+            primary_text: _("Are you sure you want to reload this tab?"),
+            secondary_text:
+                _("There is an active process on this tab.") + " " +
+                _("If you reload it, the process will end."),
```
jeremypw, line 53: I don't think it good practice to construct messages like this - it can prevent proper translation into some languages (and there is no context for "it" in the second translation).
```suggestion
                _("There is an active process on this tab. If you reload it, the process will end.")
```

jeremypw (changes requested):
Just a comment on the translatability of the message, otherwise looks good and works as expected. Thanks.

src/Dialogs/ForegroundProcessDialog.vala
andreasfelix: It just copy-pasted this part from above, it should probably also be changed there then!

jeremypw:
Yes. I don't know why those other messages were accepted but better to change those in a separate cleanup PR.

jeremypw (approved):

PR #580, opened 2021-05-19
Add accellabel to reload menuitem

andreasfelix:
Should be merged after #578 

This should probably wait until there is an API to set the accel-labels in the  [`Granite.DynamicNotebook`](https://valadoc.org/granite/Granite.Widgets.DynamicNotebook.html), otherwise the `reload_menuitem` is the only item with an accel-label.


jeremypw:
There is this: https://github.com/elementary/granite/pull/236 which, hopefully can be extended to include the tab context menus

danrabbit (approved):
Nice fix :+1:

jakubklos77:
Hello,
the f44d189 commit is too much. It interferes with basic terminal functionality. You cannot press F5 in Midnight Commander anymore and there are other downsides as well. Isn't it enough to have the Ctrl+Shift+R to Reload?

jeremypw:
`F5` was implemented to match the `Reload` action shortcut in Files.  We should probably ignore it when a foreground process like Midnight Commander is running.

jakubklos77:
Yes, that would be nice. Thank you

PR #568, opened 2021-03-24
Update window title and tab labels properly when custom prompt

jeremypw:
Fixes #567 

* Do not rely on terminal.window_title property
* Get tab basename from shell location
* Check when content changes when window_title is empty (custom prompt)

For some reason, when there is a custom prompt defined in `~/.bashrc`, the Vte terminal widget does not update the window title property when the CWD changes.  Terminal was relying on being notified of that change to schedule setting the tab labels (including disambiguation).

In this, not very satisfactory, solution, when the window title is empty, the name checking is scheduled every time the tab content changes.  To avoid too much unnecessary work, the name checking does not occur until 100ms has passed without the function being called and the contents changed signal handler does not trigger unless the window title is empty.

For comparison:  Left = Mate Terminal, Right = this PR both with two tabs one at `/home/jeremy` and one at `/home/jeremy/TestFolder/jeremy` and with a custom prompt defined by `PS1="custom prompt >" in `~/.bashrc`

Screenshot from 2021-03-24 18 22 01

PR #584, opened 2021-05-24
Refactor name check and title update so works with custom prompt

### LINKED ISSUE #567: 4-5% CPU usage for idle terminal
I think I found a small-to-medium performance issue. I want to say in advance, I didn't spend much time investigating this. So I just opened this issue so it is documented. Maybe I can pick it up later or someone else decides to investigate this. 

The lines below introduce introduce a timeout which is never canceled. It saves the open terminals tabs **20 times** per second, which causes ~4-5% CPU and ~700 READ/WRITE syscall per second on my machine. (Increases with the number of tabs)

When looking at the debug logs (`G_MESSAGE_DEBUG=all`), one can also see that the open tabs are saved in a 50 ms interval.

<details>
  <summary>Debug Logs</summary>

```console
...
[DEBUG 15:30:39.288819] MainWindow.vala:1477: save open terminals!
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_notify: /io/elementary/terminal/saved-state/zoom
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_notify: /io/elementary/terminal/saved-state/tabs
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.288: change_notify: /io/elementary/terminal/saved-state/tab-zooms
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.289: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.289: change_notify: /io/elementary/terminal/saved-state/focused-tab
[DEBUG 15:30:39.339260] MainWindow.vala:1477: save open terminals!
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_notify: /io/elementary/terminal/saved-state/zoom
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_notify: /io/elementary/terminal/saved-state/tabs
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_notify: /io/elementary/terminal/saved-state/tab-zooms
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_fast
(io.elementary.terminal:143665): dconf-DEBUG: 15:30:39.339: change_notify: /io/elementary/terminal/saved-state/focused-tab
```
</details>

If uncommenting these lines the CPU usages drops to 0-0.5%.

https://github.com/elementary/terminal/blob/12cb5d40825dc923aa49726ae90ec11f5b8ff40f/src/MainWindow.vala#L1423-L1433

---
**Edit:** the below does not make sense

<strike>
Is it possible that it should just be

```vala
if (check_for_tabs_with_same_name) {
   // ...
}
```
instead of

```vala
if (!check_for_tabs_with_same_name) {
   // ...
}
```
?
</strike>

### LINKED ISSUE #583: Tabs with same name not disambiguated with custom prompt
## What Happened

Tabs with the same name are not disambiguated if a custom prompt is used

## Expected Behavior

Enough of the path to disambiguate the tab name should be used as the tab label.

## Steps to Reproduce

1 Edit `~/.bashrc` to include `PS1="custom prompt >"`
1 Open a tab at e.g. ..../xxx/build
2 Open a second tab at e.g. .../yyy/build
3. Both tabs are labelled "build"
4. Window title stays as "Terminal" regardless of selected tab

## Platform Information

Elementary 6 daily

jeremypw:
Fixes #583 
Fixes #567 

Replaces #568 

* Run update on new tab, switch tab, close tab and contents changed
* Do not use terminal.window_title

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1129,7 +1129,7 @@ namespace Terminal {
             sw.add (term);
             var tab = new Granite.Widgets.Tab (label, icon, sw);
             term.tab = tab;
-            tab.ellipsize_mode = Pango.EllipsizeMode.START;
+            tab.ellipsize_mode = Pango.EllipsizeMode.MIDDLE;
 
             var reload_menu_item = new Gtk.MenuItem.with_label (_("Reload"));
             tab.menu.append (reload_menu_item);
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -752,7 +752,6 @@ namespace Terminal {
 
         private bool on_close_tab_requested (Granite.Widgets.Tab tab) {
             var t = get_term_widget (tab);
-            bool closing_current_terminal = t == current_terminal;
 
             if (t.has_foreground_process ()) {
                 var d = new ForegroundProcessDialog (this);
@@ -775,7 +774,7 @@ namespace Terminal {
             }
 
             current_terminal.grab_focus ();
-            if (!closing_current_terminal) { // Otherwise current_terminal will change triggering a name check
+            if (t != current_terminal) { // Otherwise current_terminal will change triggering a name check
                 check_for_tabs_with_same_name ();
             }
 
@@ -795,6 +794,7 @@ namespace Terminal {
             notebook.insert_tab (tab, -1);
             notebook.current = tab;
             term.grab_focus ();
+            check_for_tabs_with_same_name ();
         }
 
         private void on_tab_moved (Granite.Widgets.Tab tab, int x, int y) {
@@ -958,8 +958,6 @@ namespace Terminal {
 
                 return false;
             });
-
-            check_for_tabs_with_same_name ();
         }
 
         private void open_tabs () {
@@ -1438,14 +1436,13 @@ namespace Terminal {
                     string term2_path = terminal2.current_working_directory;
                     string term2_name = Path.get_basename (term2_path);
 
-                    if (terminal2 != terminal && term2_name == term_label) {
-                        if (term2_path != term_path) {
-                            terminal2.tab_label = disambiguate_label (term2_path, term_path);
-                            terminal.tab_label = disambiguate_label (term_path, term2_path);
-                            // Overwrite tooltip_text set by changing tab_label
-                            terminal2.tooltip_text = term2_path;
-                            terminal.tooltip_text = term_path;
-                        }
+                    if (terminal2 != terminal && term2_name == term_label && term2_path != term_path) {
+                        terminal2.tab_label = disambiguate_label (term2_path, term_path);
+                        terminal.tab_label = disambiguate_label (term_path, term2_path);
+                        // Overwrite tooltip_text set by changing tab_label
+                        terminal2.tooltip_text = term2_path;
+                        terminal.tooltip_text = term_path;
+
                     }
                 }
             }
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -585,6 +585,8 @@ namespace Terminal {
                             current_terminal.remember_command_end_position ();
                             get_simple_action (ACTION_COPY_LAST_OUTPUT).set_enabled (false);
                         }
+
+                        schedule_name_check (); // CWD may have changed
                         break;
 
                     case Gdk.Key.@1: //alt+[1-8]
@@ -939,7 +941,6 @@ namespace Terminal {
                                      Granite.Widgets.Tab new_tab) {
 
             current_terminal = get_term_widget (new_tab);
-            title = current_terminal.tab_label ?? TerminalWidget.DEFAULT_LABEL;
             /* The font-scales of all terminals are currently the synchronized through saved-state binding */
             new_tab.icon = null;
             Idle.add (() => {
@@ -956,6 +957,8 @@ namespace Terminal {
 
                 return false;
             });
+
+            update_title ();
         }
 
         private void open_tabs () {
@@ -1077,17 +1080,6 @@ namespace Terminal {
                 }
             });
 
-            /* This signal is not emitted when the .bashrc is missing or does not force terminal to
-             * set its title. So we cannot detect path changes in the shell. Tab name remains "Terminal".
-             */
-            t.window_title_changed.connect (() => {
-                if (t == current_terminal) {
-                    title = t.window_title;
-                }
-
-                schedule_name_check ();
-            });
-
             t.set_font (term_font);
 
             if (current_terminal != null) {
@@ -1125,6 +1117,7 @@ namespace Terminal {
                 t.run_program (program, location);
             }
 
+            schedule_name_check ();
             return t;
         }
 
@@ -1301,6 +1294,7 @@ namespace Terminal {
         private void action_close_tab () {
             current_terminal.tab.close ();
             current_terminal.grab_focus ();
+            schedule_name_check ();
         }
 
         private void action_new_window () {
@@ -1447,7 +1441,7 @@ namespace Terminal {
 
             foreach (TerminalWidget terminal in terms) {
                 string term_path = terminal.get_shell_location ();
-                string term_label = terminal.window_title;
+                string term_label = Path.get_basename (term_path);
 
                 if (term_label == "") { /* No point in continuing - tabs not finished updating */
                     return false; /* Try again later */
@@ -1462,22 +1456,25 @@ namespace Terminal {
 
                 foreach (TerminalWidget terminal2 in terms2) {
                     string term2_path = terminal2.get_shell_location ();
-                    string term2_name = terminal2.window_title;
+                    string term2_name = Path.get_basename (term2_path);
 
                     if (terminal2 != terminal && term2_name == term_label) {
                         if (term2_path != term_path) {
                             terminal2.tab_label = disambiguate_label (term2_path, term_path);
                             terminal.tab_label = disambiguate_label (term_path, term2_path);
-
-                            if (terminal == current_terminal) {
-                                title = terminal.tab_label;
-                            }
                         }
                     }
                 }
+
+                terminal.tab.tooltip = term_path;
             }
 
+            update_title ();
             return true;
+        }
+
+        private void update_title () {
+            title = current_terminal.get_shell_location ();
         }
 
         private void save_opened_terminals () {
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1419,25 +1419,28 @@ namespace Terminal {
         }
 
         private uint name_check_timeout_id = 0;
+        private bool next_time = false;
         private void schedule_name_check () {
             if (name_check_timeout_id > 0) {
-                Source.remove (name_check_timeout_id);
+                next_time = false; // Keep existing timeout running
+                return;
             }
 
-            name_check_timeout_id = Timeout.add (50, () => {
-                save_opened_terminals ();
-
-                if (!check_for_tabs_with_same_name ()) {
-                    return true;
-                } else {
-                    name_check_timeout_id = 0;
-                    return false;
+            name_check_timeout_id = Timeout.add (250, () => {
+                if (!next_time) {
+                    next_time = true;
+                    return Source.CONTINUE;
                 }
+
+                check_for_tabs_with_same_name ();
+                name_check_timeout_id = 0;
+                next_time = false;
+                return Source.REMOVE;
             });
         }
 
         /** Compare every tab label with every other and resolve ambiguities **/
-        private bool check_for_tabs_with_same_name () {
+        private void check_for_tabs_with_same_name () {
             /* Take list copies so foreach clauses can be nested safely*/
             var terms = terminals.copy ();
             var terms2 = terminals.copy ();
@@ -1446,12 +1449,10 @@ namespace Terminal {
                 string term_path = terminal.get_shell_location ();
                 string term_label = Path.get_basename (term_path);
 
-                if (term_label == "") { /* No point in continuing - tabs not finished updating */
-                    return false; /* Try again later */
-                }
+                if (term_label == "" ||
+                    terminal.tab_label == TerminalWidget.DEFAULT_LABEL) {
 
-                if (terminal.tab_label == TerminalWidget.DEFAULT_LABEL) { /* Absent or incorrect .bashrc */
-                    return true;
+                    continue;
                 }
 
                 /* Reset tab_name to basename so long name only used when required */
@@ -1473,7 +1474,7 @@ namespace Terminal {
             }
 
             update_title ();
-            return true;
+            return;
         }
 
         private void update_title () {
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1418,7 +1418,7 @@ namespace Terminal {
             var terms2 = terminals.copy ();
 
             foreach (TerminalWidget terminal in terms) {
-                string term_path = terminal.get_shell_location ();
+                string term_path = terminal.current_working_directory;
                 string term_label = Path.get_basename (term_path);
 
                 if (term_label == "" ||
@@ -1431,7 +1431,7 @@ namespace Terminal {
                 terminal.tab_label = term_label;
 
                 foreach (TerminalWidget terminal2 in terms2) {
-                    string term2_path = terminal2.get_shell_location ();
+                    string term2_path = terminal2.current_working_directory;
                     string term2_name = Path.get_basename (term2_path);
 
                     if (terminal2 != terminal && term2_name == term_label) {
@@ -1445,7 +1445,7 @@ namespace Terminal {
                 terminal.tab.tooltip = term_path;
             }
 
-            title = current_terminal.get_shell_location ();
+            title = current_terminal.current_working_directory;
             return;
         }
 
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -1077,14 +1079,7 @@ namespace Terminal {
...
-            /* This signal is not emitted when the .bashrc is missing or does not force terminal to
-             * set its title. So we cannot detect path changes in the shell. Tab name remains "Terminal".
-             */
-            t.window_title_changed.connect (() => {
-                if (t == current_terminal) {
-                    title = t.window_title;
-                }
-
+            t.contents_changed.connect (() => {
                 schedule_name_check ();
```
andreasfelix, line 1083: This gets called "whenever the visible appearance of the terminal has changed". This means it is called on every keystroke and also when a program which outputs a log messages or graphical program like htop changes the contents of the terminal. So this means it is called approximately 50-500 times per minute for normal terminal usage. So it should be made sure that this calls are not too expensive.

At the moment this callback eventually calls `save_opened_terminals`, which is relatively expensive operation. For my machine the CPU usage more than doubled when using vim.

On counter measure would be to check if the terminal has a running foreground process and return immediately if this is the case.

andreasfelix: Maybe this should also only be run on user input and not if the contents of the terminal are changed by a program.

andreasfelix:
Do you know what's the point of [`current_directory_uri_changed`](https://valadoc.org/vte-2.91/Vte.Terminal.current_directory_uri_changed.html), it sounds like a signal which should get emitted when the directory changes, but it seems to never get emitted and calling [`get_current_directory_uri`](https://valadoc.org/vte-2.91/Vte.Terminal.get_current_directory_uri.html) returns `null` on my machine.

Shouldn't it be technically possible to get notified whenever the CWD changes, without relying on continuously polling for it? Similar how `get_shell_location` 
https://github.com/elementary/terminal/blob/ac1bcc066302044529ae25d76b8ab067e1bf3ff4/src/Widgets/TerminalWidget.vala#L450-L460

use the Linux `/proc/<pid>/cwd` file, should it be possible to set up an watcher using something like [`inotify_add_watch`](https://valadoc.org/linux/Linux.inotify_add_watch.html) ? Or maybe trace the `chdir` syscalls of the child processes?

src/MainWindow.vala
andreasfelix: Denouncing the `contents_changed` by ~500ms could probably also help.

andreasfelix: Another idea: Doesn't changing the directory always require to press <kbd>Enter</kdb>? Wouldn't be the simplest solution to schedule the name_check every time <kbd>Enter</kdb> is pressed and no foreground process is running?

jeremypw:
@andreasfelix Thanks for the review!
I thought I took out `save_open_terminals` - I intended to.  It doesn't seem necessary when updating the labels.  I originally only had the labels update when the user entered `Return` then changed it to `contents_changed` so that if a script changed the cwd it was reflected in the labels.

The updating is throttled by a timeout so I do not think it is particularly expensive.

You are right that a better solution would to be notified of a cwd change rather than reacting to signals that `might` have changed the cwd (would still have to react to adding, removing and switching tabs though).  I did not know how to do that but I'll look into your suggestions.

jeremypw:
Isn't `/proc/` a virtual directory? So I don't know whether you can monitor it.

andreasfelix:
oh yes :/ according to stackoverflow seems like [watching the proc filesystem is not possible.](https://stackoverflow.com/questions/6003830/how-to-monitor-the-change-of-a-processs-working-directory)

andreasfelix:
IMO, the easiest solution would be to implement `check_cwd_changed` function in the `TerminalWidget`, which emits a `cwd_changed` signal if the CWD has changed.

Then you could register the `t.cwd_changed.connect (schedule_name_check);` and call `current_terminal.check_cwd_changed` every time <kbd>Enter</kbd> is pressed.

I think this way the code would be a little bit more self documenting, because it makes clear that the names has to be checked every time a terminal widget's cwd changes. And if in the future somebody comes up with some `Posix/Linux` magic, they just have to replace the logic of the `TerminalWidget.cwd_changed` signal.

But this way would catch the case the CWD is changed by a script. But this way the elementary terminal would still be better than Mate terminal, Gnome terminal or tilix when it comes to displaying the CWD.

jeremypw:
The last commit reduces the cost of calling `schedule_name_check` rapidly. If two calls arrive within 500mS then the existing timeout just keeps running for another 250+mS.  The call simply performs an int comparison and checks a boolean.

I do not see a lot of advantage in creating a new terminal signal tbh instead of monitoring the existing terminal signals.

andreasfelix:
If you rather want to cover all cases by using `contents_changed`, I would still move this to the `TerminalWidget` and store the `cwd` into an instance variable and emit an `cwd_changed` if the `cwd` changes. This should scale better with a larger number of tabs because the `check_for_tabs_with_same_name` only has to run if one of the `TerminalWidgets` cwd actually changes.

jeremypw:
But then you have each terminal running code everytime its contents changes instead of MainWindow.  Each terminal would have to connect to and process the `contents_changed` signal instead of MainWindow and *then* emit another signal which MainWindow has to process.

I suppose it might be worth it as `check_for_tabs_with_same_name` is fairly expensive but does not run very often now.

andreasfelix:
Isn't the `contents_changed` not a signal of the `TerminalWidget` anyway? So isn't it also processed for each individual terminal in the current state?

andreasfelix:
Isn't the `contents_changed` a signal of the `TerminalWidget`? To my understanding it gets emitted for each `TerminalWidget` individually and each of them should process only if their contents change.

jeremypw:
`contents_changed` is a signal of `Vte.Terminal`, the TerminalWidget does not do any processing on it currently so would have to connect to maintain a cwd property and decide whether to notify MainWindow.

andreasfelix:
Yes, this is what I suggested. But if you don't like that, I could do a separate PR afterwards. The PR is probably still worth merging, as it is already a big improvement over the current state.

jeremypw:
It should be possible for MainWindow to maintain a cwd property on each terminal and ignore signals where it hasn't changed but it would mean calls to `get_shell_location ()` and comparison with the current cwd every time the contents changes - this is more expensive (and more code) than calling `schedule_name_check`.

andreasfelix:
schedule_name_check calls the get_shell_location for all terminal widgets. so it should be much more expensive than a single call to get_shell_location. at least to my understanding of the code.

jeremypw:
One advantage of your approach is that it could be more responsive - it may not be necessary to throttle updating the labels.  I'll give it a try.

jeremypw:
I have to admit your way works at least as well and is more responsive :smile: because I removed the throttle.

The only case it doesn't work (and probably the original didnt) is when you drag a tab out of a window - the remaining tab labels are not updated.

src/MainWindow.vala
```diff
@@ -1072,20 +1071,11 @@ namespace Terminal {
...
-             */
-            t.window_title_changed.connect (() => {
-                if (t == current_terminal) {
-                    title = t.window_title;
-                }
-
-                schedule_name_check ();
+            t.cwd_changed.connect (() => {
+                check_for_tabs_with_same_name ();
             });
```
andreasfelix, line 1079:
```suggestion
            t.cwd_changed.connect (check_for_tabs_with_same_name);
```

Could this be simplified?

jeremypw:
@andreasfelix You might like to look at https://github.com/elementary/code/pull/1050 for comparison.  This uses a different method of disambiguation based on File functions which is a little simpler but in theory could result in one tab name being much longer than the other whereas my method ensures they are the same length.

src/MainWindow.vala
```diff
@@ -1421,63 +1412,42 @@ namespace Terminal {
...
                             terminal.tab_label = disambiguate_label (term_path, term2_path);
-
-                            if (terminal == current_terminal) {
-                                title = terminal.tab_label;
-                            }
                         }
                     }
                 }
+
+                terminal.tab.tooltip = term_path;
```
andreasfelix, line 1444: Would it be possible to move this to `TerminalWidget` and make use of `bind_property` to automatically update the tooltip?

andreasfelix: Or, just move this to the `check_cwd_changed` function.

src/Widgets/TerminalWidget.vala
```diff
@@ -331,6 +336,7 @@ namespace Terminal {
         void on_child_exited () {
             child_has_exited = true;
             last_key_was_return = true;
+            check_cwd_changed ();
```
andreasfelix, line 339: why do you have to check here? To my understanding a child process shouldn't be able to change the cwd of the parent.

src/Widgets/TerminalWidget.vala
```diff
@@ -608,6 +616,15 @@ namespace Terminal {
             Posix.kill (child_pid, Posix.Signal.TERM);
             reset (true, true);
             active_shell (old_loc);
+            check_cwd_changed ();
```
andreasfelix, line 619: if you call check_cwd_changed in active_shell you can remove it here.

src/MainWindow.vala
jeremypw: OK, I'll move it to `check_cwd_changed` it doesn't need to be here.

jeremypw: Tried `bind_property` but doesnt work because `Granite.Widgets.Tab` resets the tooltip_text everytime the label text is reassigned - which may happen even when the cwd has not changed.  So I think its better to leave it as it is tbh. We do not want to reassign it on every call to `check_cwd_changed`

src/Widgets/TerminalWidget.vala
jeremypw: OK, I wasnt sure so I played safe.  Removing.

src/Widgets/TerminalWidget.vala
jeremypw: Well spotted.

src/MainWindow.vala
jeremypw: The cwd can be changed by scripts, which are not foreground processes. If there is a way of detecting when a bash script is running (or detecting when a script inserts `Enter` then that would be more efficient

src/MainWindow.vala
```diff
@@ -776,6 +775,10 @@ namespace Terminal {
             }
 
             current_terminal.grab_focus ();
+            if (!closing_current_terminal) { // Otherwise current_terminal will change triggering a name check
```
andreasfelix, line 778:
```suggestion
            if (!t != current_terminal) { // Otherwise current_terminal will change triggering a name check
```

Is the `closing_current_terminal` variable necessary?

src/MainWindow.vala
```diff
@@ -1421,63 +1415,43 @@ namespace Terminal {
...
                 foreach (TerminalWidget terminal2 in terms2) {
-                    string term2_path = terminal2.get_shell_location ();
-                    string term2_name = terminal2.window_title;
+                    string term2_path = terminal2.current_working_directory;
+                    string term2_name = Path.get_basename (term2_path);
 
                     if (terminal2 != terminal && term2_name == term_label) {
                         if (term2_path != term_path) {
                             terminal2.tab_label = disambiguate_label (term2_path, term_path);
                             terminal.tab_label = disambiguate_label (term_path, term2_path);
```
andreasfelix, line 1444: As the loop goes over all permutations instead of all combinations, I think this can be simplified to:

```suggestion
                    if (terminal2 != terminal && term2_name == term_label && term2_path != term_path) {
                        terminal.tab_label = disambiguate_label (term_path, term2_path)
```

sorry github doesn't makes it possible to make suggestions for deleted lines, so I had to split this up in to.

src/MainWindow.vala
```diff
@@ -1421,63 +1415,43 @@ namespace Terminal {
...
-
-                            if (terminal == current_terminal) {
-                                title = terminal.tab_label;
-                            }
+                            // Overwrite tooltip_text set by changing tab_label
+                            terminal2.tooltip_text = term2_path;
+                            terminal.tooltip_text = term_path;
                         }
```
andreasfelix, line 1448:
```suggestion
                        // Overwrite tooltip_text set by changing tab_label
                        terminal.tooltip_text = term_path;
```

src/Widgets/TerminalWidget.vala
```diff
@@ -223,6 +226,8 @@ namespace Terminal {
                 resized = true;
             });
 
+            contents_changed.connect (check_cwd_changed);
```
andreasfelix, line 229: The only thing that still bother me a little bit, is that this is also triggered for foreground processes. Guarding it with a `has_foreground_process` check probably doesn't make sense, because I assume they are comparatively expensive. 

Anyway it probably doesn't matter so much, I didn't see any performance differences on my machine.

andreasfelix: One solution would be to unregistered the callback when a foreground process is running and re-register it on `child_exited`. But this is probably a unnecessary micro optimizations, just came to my mind.

src/MainWindow.vala
jeremypw: I thought `current_terminal` may have already changed by that point so this seemed safer. I guess the check is only needed when the current_terminal is not closed in which it will not have changed so your way may work - I'll try it.

src/MainWindow.vala
```diff
@@ -956,6 +958,8 @@ namespace Terminal {
 
                 return false;
             });
+
+            check_for_tabs_with_same_name ();
```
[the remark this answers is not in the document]
jeremypw: Because pages are switched when a tab is closed or dragged out of the window - but it would probably be better to deal with those cases directly.

src/MainWindow.vala
jeremypw: With these changes I find that closing a tab and then re-opening it only disambiguates one of the tabs - the other keeps the basename. I don't think this is what is wanted.

jeremypw: This is because when first terminal has already been disambiguated, the basename of the second terminal no longer matches it so the second term is never disambiguated.

src/MainWindow.vala
jeremypw: Yes, that seems to work OK - thanks.

src/Widgets/TerminalWidget.vala
jeremypw: Yes, if foreground processes cannot change the cwd then this check is unnecessary. But I doubt whether the extra complexity of code is worth the (undetectable) performance gain.

src/MainWindow.vala
andreasfelix: If the script is run in a sub-process (which is in my understanding equivalent to foreground process) like this
```
bash script.sh or ./script.sh
```
it shouldn't be able to change the CWD. But if somebody sources a script which changes the cwd script like this

```
source script.sh
```
yes then it could the change the CWD. This is probably really rare so you have to decide if it is worth taking into acount..

Also there is already some kind of precedent set with the `remember_position` which relies on the <kbd>Enter</kbd> key being pressed. Using the `contents_changed` signal introduces a second mechanism for similar issue: Both want to check if a command was executed. IMO, ideally both should rely on the same mechanism.

src/MainWindow.vala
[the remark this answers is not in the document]
andreasfelix: ok makes sense. I don't have an opinion what is better here.

src/MainWindow.vala
andreasfelix: hmm ok, still I think it should be possible to somehow to get away with doing half the work as the loop runs over all the permutations. Anyway this is a different issue.

src/Widgets/TerminalWidget.vala
```diff
@@ -28,6 +28,7 @@ namespace Terminal {
         internal const string DEFAULT_LABEL = _("Terminal");
         public Terminal.Application app;
         public string terminal_id;
+        public string current_working_directory { get; private set; default = "";}
```
andreasfelix, line 31: Codestyle
```suggestion
        public string current_working_directory { get; private set; default = ""; }
```

andreasfelix (approved):
The code looks good to me! It solves both linked issues. And even though the callback of the `contents_changed` signal is called multiple times per second when using the terminal, I can't measure any performance differences.

jeremypw:
Thanks for your patience in collaborating on this! I don't particularly like connecting to `contents-changed` but at least the handler is now fairly inexpensive.

PR #586, opened 2021-05-28
Update window title on switch page

jeremypw:
Reinstate fix that was inadvertantly reverted in final version of c3e36fb2ab64c18028ff2b4a6da5bfb2171c1c04


andreasfelix (approved):
Sorry, I also missed this in my review! I can confirm this solves the issue.

PR #587, opened 2021-05-28
Ensure tab tooltip shows full path

jeremypw:
Change the tooltip back after being changed by Granite.  Not an elegant solution - is there a better one without changing Granite?

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1137,6 +1137,11 @@ namespace Terminal {
             sw.add (term);
             var tab = new Granite.Widgets.Tab (label, icon, sw);
             term.tab = tab;
+            /* We have to rewrite the tooltip everytime the label changes to override Granite annoying habit of 
+             * automatically changing the tooltip to be the same as the label. */
+            term.tab.notify["label"].connect_after (() => {
+                term.tab.tooltip = term.current_working_directory;
+            });
             tab.ellipsize_mode = Pango.EllipsizeMode.MIDDLE;
 
             var reload_menu_item = new Gtk.MenuItem.with_label (_("Reload"));


andreasfelix:
This could be an indicator that the code in the `DynamicNotebook` may be a little bit over abstracted. It makes many assumptions and has many parameters to customize the behavior. While this is very effective to reduce boilerplate code in the other apps, it also somewhat limits the possibilities. The only way to change some behavior for an application on a deeper level is to make changes to the `DynamicNotebook` by adding additional options (see elementary/granite/pull/502) or to "monkey-patch" the behavior (this PR).

Maybe the most elegant solution would be to drastically reduce the area of responsibility of the `DynamicNotebook` by removing all predefined tool-tips and menu-items and implement most of the behavior directly in the different elementary apps. This would of course lead to some code duplication for the different elementary apps, but would make the code somewhat more flexible and less complex. Sometimes there is no one-fits-all-solution and IMO sometimes code duplication can be the lesser evil than the wrong abstraction.

As the elementary apps already using libhandy, another solution could be to switch to the the `Hdy.TabBar` and `Hdy.TabView`, which are a little bit more vanilla than the `DynamicNotebook` when it comes to the tooltips and tab context menus.

This is of course only my opinion and it could be totally wrong, but it came to my mind before when reading the code of the `DynamicNotebook`.

andreasfelix (approved):
Otherwise I can confirm that this works as expected. The stated above should probably discussed in a different issue.

jeremypw:
I agree with your sentiments about DynamicNotebook and will look into whether libhandy can provide a suitable altermative - thanks.  I raised an issue about this in the Granite repo long ago - https://github.com/elementary/granite/issues/201 - but never got round to proposing a fix.  Maybe I'll look at it again.

PR #590, opened 2021-05-31
Replace some deprecated functions

andreasfelix:
This gets rid of some warning messages during compilation.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -331,7 +331,7 @@ namespace Terminal {
         protected bool match_keycode (int keyval, uint code) {
 #endif
             Gdk.KeymapKey [] keys;
-            Gdk.Keymap keymap = Gdk.Keymap.get_for_display (get_display ());
+            var keymap = Gdk.Keymap.get_for_display (get_display ());
             if (keymap.get_entries_for_keyval (keyval, out keys)) {
                 foreach (var key in keys) {
                     if (code == key.keycode)
@@ -726,7 +726,7 @@ namespace Terminal {
             default_height = rect.height;
 
             if (default_width == -1 || default_height == -1) {
-                Gdk.Rectangle geometry = get_display ().get_primary_monitor ().get_geometry ();
+                var geometry = get_display ().get_primary_monitor ().get_geometry ();
 
                 default_width = geometry.width * 2 / 3;
                 default_height = geometry.height * 3 / 4;
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -331,7 +331,7 @@ namespace Terminal {
         protected bool match_keycode (int keyval, uint code) {
 #endif
             Gdk.KeymapKey [] keys;
-            Gdk.Keymap keymap = Gdk.Keymap.get_default ();
+            Gdk.Keymap keymap = Gdk.Keymap.get_for_display (get_display ());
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -331,7 +331,7 @@ namespace Terminal {
         protected bool match_keycode (int keyval, uint code) {
 #endif
             Gdk.KeymapKey [] keys;
-            Gdk.Keymap keymap = Gdk.Keymap.get_default ();
+            Gdk.Keymap keymap = Gdk.Keymap.get_for_display (get_display ());
```
jeremypw, line 334: Might as well follow current style and use `var keymap = ...` here.

src/MainWindow.vala
```diff
@@ -726,8 +726,7 @@ namespace Terminal {
             default_height = rect.height;
 
             if (default_width == -1 || default_height == -1) {
-                Gdk.Rectangle geometry;
-                get_screen ().get_monitor_geometry (get_screen ().get_primary_monitor (), out geometry);
+                Gdk.Rectangle geometry = get_display ().get_primary_monitor ().get_geometry ();
```
jeremypw, line 729: Might as well use `var geometry = .....` here.

jeremypw (changes requested):
Good cleanup. Just a couple of very minor comments.

src/MainWindow.vala
andreasfelix: done

src/MainWindow.vala
andreasfelix: done

andreasfelix:
Is it save to remove this line? As far as I understand [this setting is ignored now](https://developer.gnome.org/gtk3/stable/GtkSettings.html#GtkSettings--gtk-menu-bar-accel).

https://github.com/elementary/terminal/blob/7c970414b8a811e2985ec6a0d2e9ebc289ae0fe3/src/MainWindow.vala#L191-L193

jeremypw:
I would stick to the stated scope of the PR tbh if you want a quick merge.

jeremypw:
Have you looked into replacing the `spawn_sync ()` function? (see my comment on Slack). I couldn't get `spawn_async ()` to work - looks like the `.vapi` is wrong.

andreasfelix:
> I would stick to the stated scope of the PR tbh if you want a quick merge.

The `gtk_settings.gtk_menu_bar_accel` also throw a deprecation warning, therefore I considered it to be the same scope.

andreasfelix:
> Have you looked into replacing the `spawn_sync ()` function? (see my comment on Slack). I couldn't get `spawn_async ()` to work - looks like the `.vapi` is wrong.

tbh I intentionally left that out because I haven't used async in vala before. It works a bit different than in python, js or rust, the languages I happen to know :D

I am only in the community slack. I suppose there is another one, but I am not part of it. So I couldn't see you comment.

danrabbit:
@andreasfelix I would be happy to send you an invite to the contributor slack. If you want to DM me on the community slack or have an email address I can send an invite link to. I can't seem to find you on the community slack

jeremypw (approved):
Removing the gtk_menu_bar_accel override should be safe enough I guess - it isn't used elsewhere in elementary or within Terminal.  I guess it is to be sure that terminal programs like `htop` can be exited with it.

PR #595, opened 2021-06-10
Unsafe Pastes: Handle multi-line pastes as well

cassidyjames:
Based on (and supersedes) https://github.com/elementary/terminal/pull/556. Keeps multi-line pastes under the same setting as unsafe pastes to simplify things, especially if we ever add a GUI setting for this.

Updated the UnsafePasteDialog to accept a title so we can be more contextual while not creating an entirely new dialog.

Might be able to DRY it up a little more?


src/MainWindow.vala
```diff
@@ -1229,8 +1228,28 @@ namespace Terminal {
                 }
                 var text = intext.strip ();
 
+                /* warn on pasting text with `sudo` */
                 if ((text.index_of ("sudo") > -1) && (text.index_of ("\n") != 0)) {
-                    var dialog = new UnsafePasteDialog (this, text);
+                    var dialog = new UnsafePasteDialog (
```
JoseExposito, line 1233: This 2 blocks of code (the if...else) are duplicated. I think we could refactor them since the only bit that changes is the string literal.

JoseExposito (changes requested):

src/MainWindow.vala
```diff
@@ -1217,7 +1217,6 @@ namespace Terminal {
         }
 
         private void on_get_text (Gtk.Clipboard board, string? intext) {
-            /* if unsafe paste alert is enabled, show dialog */
             if (Application.settings.get_boolean ("unsafe-paste-alert") && !unsafe_ignored ) {
```
jeremypw, line 1221: Extraneous white space

src/MainWindow.vala
```diff
@@ -1229,8 +1228,28 @@ namespace Terminal {
                 }
                 var text = intext.strip ();
 
+                /* warn on pasting text with `sudo` */
                 if ((text.index_of ("sudo") > -1) && (text.index_of ("\n") != 0)) {
-                    var dialog = new UnsafePasteDialog (this, text);
+                    var dialog = new UnsafePasteDialog (
+                        this,
+                        _("This command is asking for Administrative access to your computer"),
```
jeremypw, line 1235: Strictly speaking, the word "sudo" might be inside quotes in the pasted text (which may not be a command). Maybe rethink this text?

src/MainWindow.vala
```diff
@@ -1229,8 +1228,28 @@ namespace Terminal {
...
+                        text
+                    );
+
+                    if (dialog.run () != Gtk.ResponseType.ACCEPT) {
+                        dialog.destroy ();
+                        return;
+                    }
+                    dialog.destroy ();
```
jeremypw, line 1243: Blank line after if clause recommended.

src/MainWindow.vala
cassidyjames: @jeremypw what's extraneous here?

src/MainWindow.vala
cassidyjames: This copy is exactly as-is from master, but I can rephrase to `may` to match the multi-line paste text.

src/MainWindow.vala
cassidyjames: Thanks! I suspected it could be cleaned up but hadn't thought about how. I refactored it now, let me know if that makes more sense.

src/MainWindow.vala
Marukesu: Isn't the space on the end of the if clasule?

src/MainWindow.vala
JoseExposito: Looking much better, thanks!!

src/MainWindow.vala
```diff
@@ -1226,25 +1226,40 @@ namespace Terminal {
...
+                        text
+                    );
+
+                    if (unsafe_paste_dialog.run () != Gtk.ResponseType.ACCEPT) {
+                        unsafe_paste_dialog.destroy ();
                         return;
                     }
-                    dialog.destroy ();
+
+                    unsafe_paste_dialog.destroy ();
```
danrabbit, line 1262: Am I missing something here? It looks like we destroy no matter what and the `return` would only stop us from calling the destroy that we call anyways before returning?

danrabbit: Nvm, I was indeed missing the rest of the function :sweat_smile:

src/Dialogs/UnsafePasteDialog.vala
```diff
@@ -28,11 +29,10 @@ public class Terminal.UnsafePasteDialog : Granite.MessageDialog {
 
     construct {
         image_icon = new ThemedIcon ("dialog-warning");
-        primary_text = _("This command is asking for Administrative access to your computer");
 
         secondary_text =
-            _("Copying commands into the Terminal can be dangerous.") + " " +
-            _("Be sure you understand what each part of this command does.");
+            _("Copying commands into Terminal can be dangerous.") + " " +
+            _("Be sure you understand what each part of the pasted text does before continuing.");
```
danrabbit, line 35: If we're changing these anyways, we should probably remove the forced space and make this a single translatable string
```suggestion
            _("Copying commands into Terminal can be dangerous. Be sure you understand what each part of the pasted text does before continuing.");
```

danrabbit (changes requested):
One small string change :)

src/Dialogs/UnsafePasteDialog.vala
cassidyjames: This was separated in master and I have no idea why, so I left it as-is. But I can combine them.

JoseExposito (approved):
The code is looking much better, good job! If everybody else is happy, I think we can merge it.

jeremypw (approved):

PR #602, opened 2021-06-24
MainWindow: Use the XTerm title if defined

### LINKED ISSUE #593: Cannot Set Custom XTerm Title
## What Happened

After #584, terminal don't update the title when printing a xterm escape sequence.

* **Before**
Terminal Window With The Custom Title

* **After**
Terminal Window Without The Custom Title

## Expected Behavior

That the title is updated when a xterm escape sequence is passed.

## Steps to Reproduce

1. Open terminal.
2. Run `echo -n "\033]2;this is a custom title\007"`.
3. Nothing happens.

## Platform Information

elementaryOS 6.0 Early Access

Marukesu:
re-enable the use of `window_title` if there's a defined title. if the title is empty, we fallback to the current directory for tabs and set the title with `TerminalWidget.DEFAULT_LABEL`

closes #593


src/MainWindow.vala
```diff
@@ -961,7 +963,8 @@ namespace Terminal {
             Idle.add (() => {
                 get_term_widget (new_tab).grab_focus ();
                 update_copy_output_sensitive ();
-                title = current_terminal.current_working_directory;
+                title = current_terminal.window_title != "" ? current_terminal.window_title
+                                                            : TerminalWidget.DEFAULT_LABEL;
```
jeremypw, line 967: We need to show the path in as the window title, if not overridden.
```suggestion
                title = current_terminal.window_title != "" ? current_terminal.window_title
                                                            : current_terminal.current_working_directory;
```

src/MainWindow.vala
```diff
@@ -1488,7 +1493,8 @@ namespace Terminal {
                 }
             }
 
-            title = current_terminal.current_working_directory;
+            title = current_terminal.window_title != "" ? current_terminal.window_title
+                                                        : TerminalWidget.DEFAULT_LABEL;
```
jeremypw, line 1497: We need to show the path in as the window title, if not overridden.
```suggestion
            title = current_terminal.window_title != "" ? current_terminal.window_title
                                                        : current_terminal.current_working_directory;
```

jeremypw (changes requested):
This PR causes a regression in showing the path as the window title under normal conditions.

I've added suggestions as to how to fix this.  I am not sure whether you need to check whether current_working_directory is something valid but it is probably safer to test this and use TerminalWidget.DEFAULT_LABEL as an ultimate fallback.

src/MainWindow.vala
```diff
@@ -1462,11 +1466,11 @@ namespace Terminal {
 
             foreach (TerminalWidget terminal in terms) {
                 string term_path = terminal.current_working_directory;
-                string term_label = Path.get_basename (term_path);
+                string term_label = terminal.window_title != "" ? terminal.window_title
+                                                                : Path.get_basename (term_path);
```
jeremypw, line 1470: This causes a regression when there *isn't* a custom prompt set in `.bashrc`.  In that case the terminal label is the same as the terminal prompt (and window title) which is not what is wanted.  The label should always show the terminal path (disambiguated).

src/MainWindow.vala
```diff
@@ -1475,7 +1479,8 @@ namespace Terminal {
 
                 foreach (TerminalWidget terminal2 in terms2) {
                     string term2_path = terminal2.current_working_directory;
-                    string term2_name = Path.get_basename (term2_path);
+                    string term2_name = terminal2.window_title != "" ? terminal2.window_title
+                                                                     : Path.get_basename (term2_path);
```
jeremypw, line 1483: See above comment re tab label regression.

src/MainWindow.vala
Marukesu: that's because `.bashrc` set a title by default:
```bash
# If this is an xterm set the title to user@host:dir
case "$TERM" in
xterm*|rxvt*)
    PS1="\[\e]0;${debian_chroot:+($debian_chroot)}\u@\h: \w\a\]$PS1"
    ;;
*)
    ;;
esac
```
elementary use a patch to set the title as `dir: last_cmd` or `actual_cmd`, see: https://github.com/elementary/os-patches/blob/712e7a310b32c205fde0c73025c54f1be9181e3e/debian/skel.bashrc#L114-L171

jeremypw:
@Marukesu You do not seem to have fixed the tab label regression when no custom prompt?

Marukesu:
> You do not seem to have fixed the tab label regression when no custom prompt?

I think i misunderstand you review. as i said before, the `window_title` is the same of the prompt because `.bashrc` set both the same by default, and i don't think it's a regression but what this PR do (use a xterm title if we have one). However, if it was about the label and title be the same, i can update the PR fixing that.

jeremypw:
> However, if it was about the label and title be the same, i can update the PR fixing that.

This PR should only change the appearance of Terminal sufficient only to fix the linked issue - that is the setting a custom window title from the command-line.  It should not change the appearance in any other way.  If you feel Terminal behaves incorrectly in another way then you should raise another issue about that.  You are right that Terminal does not show tab labels the same way as e.g. Mate Terminal but that is, I think, by design.

Marukesu:
> This PR should only change the appearance of Terminal sufficient only to fix the linked issue - that is the setting a custom window title from the command-line.

i did it this way because was how it worked before #584 when there was a title.

jeremypw:
@Marukesu Sorry for the delay in re-reviewing.  I think this is now acceptable subject to resolution of the conflicts.

jeremypw:
Note: I had to use `echo -ne '\033]2;this works\007'` to get the title to change - your issue did not use the `-e` flag - perhaps your shell is different?

Marukesu:
> Note: I had to use `echo -ne '\033]2;this works\007'` to get the title to change - your issue did not use the `-e` flag - perhaps your shell is different? 


Yeah, I'm using zsh and echo apply the escape characters automatically here, while bash needs the `-e` flag for that.

jeremypw (approved):
Allows setting window title from commandline without affecting tab labelling. Thanks!

PR #604, opened 2021-06-26
Save and restore last tab closed

### LINKED ISSUE #603: Text size/zoom not remembered when closing tab
## What Happened

@techdev5521 and I discovered this during/after his edw talk. If you close a tab specifically with its close button, with `exit` in the Terminal, or with <kbd>Ctrl</kbd><kbd>D</kbd>, Terminal will not save (and then later restore) that tab's text zoom.

## Expected Behavior

Closing a tab would save its text zoom, just like it does when closing the window.

## Steps to Reproduce

1. Open Terminal with a single tab
2. Set a new text size, e.g. 200%
3. Close that tab with the tab's close button, by typing `exit`, or by pressing <kbd>Ctrl</kbd><kbd>D</kbd>
4. Re-open Terminal and see the original zoom level

## Platform Information

elementary OS 6 beta 2 with Terminal from master

jeremypw:
Fixes #603 

If closing a tab results in the window closing then the details of that tab (including zoom level) should be saved and restored.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -757,10 +757,11 @@ namespace Terminal {
 
         private void on_tab_removed (Granite.Widgets.Tab tab) {
             var terminal_widget = get_term_widget (tab);
-            terminals.remove (terminal_widget);
-
             if (notebook.n_tabs == 0) {
+                save_opened_terminals ();
                 destroy ();
+            } else {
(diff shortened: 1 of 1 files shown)


danrabbit (approved):
Appears to resolve the issue, nice work!

PR #607, opened 2021-07-10
MainWindow: Don't Close the Window During Drag and Drop

### LINKED ISSUE #577: Dragging tab causes closing Terminal (eOS 6 Public Beta).
Dragging tab to some positions (inside window AFAIK) makes Terminal close: 

https://user-images.githubusercontent.com/32203382/118425888-56d4e380-b6f4-11eb-9489-1c9b74a58ec6.mp4

Marukesu:
When there's only one tab, dragging the tab out and back of the notebook area closes the window because of the check in `on_tab_removed`. Instead, only run the check if the function is not called during a drag and do it afterwards when the `drag_end` signal is emitted.

fixes #577.


danrabbit (approved):
Fixes the issue here. Nice work!

PR #615, opened 2021-08-18
Use locale independent conversion for zoom

### LINKED ISSUE #614: Font size is reset when terminal is opened anew
## What Happened
When I reopen the terminal after having altered the font size, it is always set back to 100%

## Expected Behavior
Font size should be remembered, so the next time the terminal is opened, if it was left at 120%, it remains so.

## Steps to Reproduce
1. Open the terminal, and increase the font size by however much you want.
2. Close the terminal
3. Open the terminal and notice that the font size is back to 100%

## Platform Information
elementary OS 6
io.elementary.terminal 6.0.0

jeremypw:
Fixes #614 

Locale independent functions are use to convert zoom settings to and from double to string.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -1007,23 +1007,23 @@ namespace Terminal {
                 tabs = saved_tabs;
                 var n_tabs = tabs.length;
 
-                foreach (string zoom_s in saved_zooms) {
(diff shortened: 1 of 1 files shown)


tintou (approved):

tintou:
Good catch

PR #623, opened 2021-09-20
Avoid F5 shortcut clash with some foreground processes

### LINKED ISSUE #617: Option to disable reloading a tab with F5
## Problem
I use [ctrlp](https://github.com/kien/ctrlp.vim) in vim, and ctrlp uses F5 to refresh a list of files to search against. Terminal intercepts F5 and tries to reload a tab, and it prevents ctrlp from refreshing the list of files.

## Proposal
It would be great if Terminal provides an option to disable this behavior.

## Prior Art
N/A.

Thanks!

jeremypw:
Fixes #617 

* Use `<Shift><Control>R` as preferred accel for reload action
* Retain `F5` as alternative but modify it with `<Shift>` to avoid clashes

Other solutions are possible but this seems the simplest non-destructive one.

### DIFF
--- a/data/io.elementary.terminal.appdata.xml.in
+++ b/data/io.elementary.terminal.appdata.xml.in
@@ -31,6 +31,16 @@
     </screenshot>
   </screenshots>
   <releases>
+    <release version="6.0.1" date="2020-06-15" urgency="medium">
+      <description>
+        <p>Other updates:</p>
+        <ul>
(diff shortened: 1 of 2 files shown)


BCsabaEngine (approved):

danrabbit (approved):
Makes sense!

BCsabaEngine:
Thx 4 commit

PR #626, opened 2021-10-13
Show zoom level in the center of terminal

### LINKED ISSUE #442: Show percentage while zooming in with Ctrl+Scroll
## Prerequisites

## Feature
While zooming, I'd like to have an overlay which displays the current zoom level in percent in the middle of the terminal window.

**Existing work**
Zooming with Ctrl+Scroll was added in 5.5.0

igordsm:
Discussion about design in the issue. Will close #442


jeremypw:
@danirabbit Is this idea OK from design viewpoint? If so, you may wish to adjust the CSS.

danirabbit (changes requested):
Interesting enough. I don't have any real objections to the idea.

I think there's some issues from a code perspective, like Gtk.Overlay and Revealer are both sealed in GTK4 so you can't subclass them and adding a style provider to a widget is deprecated.

It might actually be better to just have the zoom widget class set css name to "tooltip" to recycle tooltip styles instead of adding a new style here

jeremypw:
@danirabbit Thanks for the input.  @igordsm Are you interested in taking this up again? Otherwise I will try to implement @danirabbit 's suggestions.

danirabbit:
Closing since #772 was merged

PR #631, opened 2021-10-17
Escape '#' except when fragment identifier

### LINKED ISSUE #625: Some characters are incorrectly escaped in links in Terminal
### What Happened?

When clicking on a link in a Terminal window, the character '#' is escaped when it shouldn't be, so the link the browser sees includes '%23' instead of '#'.

### Steps to Reproduce

In a Terminal window:
1. cat > test.txt
 - `https://treeherder.mozilla.org/#jobs
  ^D`
2. cat test.txt
3. Click on the link.
4. Notice that in the browser URL bar, what appears is `https://treeherder.mozilla.org/%23/jobs`.

### Expected Behavior

I would expect the link to be passed to the browser literally without escaping the '#'.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #625 
Rather than allowing all '#' characters in URLs this only allows those after the final `/`' which might be fragment identifiers.  

Open to suggestions for a more elegant method of doing this.

This PR fixes the URL given in the issue report.  Suggestions for other corner cases to test welcome.  

It would be good to add a test framework to CI test this (and maybe other functions) but that is left for another PR.


cassidyjames (changes requested):
URLs can end without a trailing slash and still have a `#` fragment; for example, the link https://github.com/nodesource/distributions/blob/master/README.md#debian-and-ubuntu-based-distributions is perfectly valid but would not be escaped properly here.

jeremypw:
@cassidyjames Hmm, OK thanks - I'll have to think again :thinking:

jeremypw:
These rules from w3.org may help:
```
2.6.4 URL manipulation and creation
To fragment-escape a string input, a user agent must run the following steps:

Let input be the string to be escaped.

Let position point at the first character of input.

Let output be an empty string.

Loop: If position is past the end of input, then jump to the step labeled end.

If the character in input pointed to by position is in the range U+0000 to U+0020 or is one of the following characters:

U+0022 QUOTATION MARK character (")
U+0023 NUMBER SIGN character (#)
U+0025 PERCENT SIGN character (%)
U+003C LESS-THAN SIGN character (<)
U+003E GREATER-THAN SIGN character (>)
U+005B LEFT SQUARE BRACKET character ([)
U+005C REVERSE SOLIDUS character (\)
U+005D RIGHT SQUARE BRACKET character (])
U+005E CIRCUMFLEX ACCENT character (^)
U+007B LEFT CURLY BRACKET character ({)
U+007C VERTICAL LINE character (|)
U+007D RIGHT CURLY BRACKET character (})
...then append the percent-encoded form of the character to output. [RFC3986]

Otherwise, append the character itself to output.

This escapes any ASCII characters that are not valid in the URI <fragment> production without being escaped.

Advance position to the next character in input.

Return to the step labeled loop.

End: Return output.
```

zeebok (approved):

Issue #632, opened 2021-10-20
Opacity of terminal makes text illegible for the poorly sighted 


KeironO:
### Problem

I've recently switched to eOS for my desktop after years of it being on my laptop. My laptop's screen isn't the best, so when I installed it on my desktop machine I'm finding it difficult to read the text because of the transparency.

### Proposal

I've been able to rectify this through `gsettings`

```
gsettings set io.elementary.terminal.settings background 'rgb(46, 46, 46)'
```

Maybe having something in the settings panel that allows you to alter the opacity would be a nice addition. I'll have no problem doing this, just give me the nod if you're down.

PR #633, opened 2021-10-21
Removed alpha values from palletes.

KeironO:
Fixes: https://github.com/elementary/terminal/issues/632

Removes alpha channel from BG_DARK and BG_SOLARIZED_LIGHT.


jeremypw:
@KeironO The logic here is still faulty and the opacity does not change until the settings are changed.  Would you like me to push a fix?

jeremypw:
Looks like this PR would not be necessary if #552  were progressed.

danrabbit:
Hasn't had an update in a long time and now conflicts. Superseded by https://github.com/elementary/terminal/pull/670

PR #638, opened 2021-11-07
Focus current tab

### LINKED ISSUE #73: Open relevant tab when clicking on a task finished notification
I had two tabs open, with a process running in tab two. When it finished, I received a notification (sweet!) but clicking the tab showed me Terminal with the first tab active, not the second one (not so sweet).

Ideally, whichever tab emitted the notification should be the active tab if a notification is clicked upon.

Launchpad Details: [#LP1642970](https://bugs.launchpad.net/bugs/1642970) Ruppal Singh - 2016-11-18 14:08:49 +0000

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44562426-open-relevant-tab-when-clicking-on-a-task-finished-notification?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

igordsm:
Fix #73 

Tested this with multiple terminal windows and tabs. should work without issues.


jeremypw:
Is it necessary to add another dependency?  Is using "get_server_time" essential?

igordsm:
@jeremypw Unfortunately it seems `get_server_time` is essential. Apparently `present_with_time` needs a timestamp to solve focus stealing and the only timestamp I passed to it that worked was this one. Using regular datetime functions did not work and I found an answer in SO that linked to this function.

src/Application.vala
```diff
@@ -56,6 +56,27 @@ public class Terminal.Application : Gtk.Application {
...
+        act.activate.connect ((v) => {
+            var now = new DateTime.now_local ();
+            size_t len;
+            var tid = v.get_string (out len);
+            foreach (var window in windows) {
+                foreach (var terminal in window.terminals) {
+                    if (terminal.terminal_id == tid) {
+                        window.set_active_terminal_tab (terminal.tab);
+                        var ts = Gdk.X11.get_server_time ((Gdk.X11.Window) window.get_window ());
+                        window.present_with_time (ts);
```
Marukesu, line 71: should be using `Gtk.get_current_event_time ()`
```suggestion
                        window.present_with_time (Gtk.get_current_event_time ());
```

src/Application.vala
```diff
@@ -56,6 +56,27 @@ public class Terminal.Application : Gtk.Application {
...
+                        window.grab_focus ();
+                        terminal.tab.grab_focus ();
+                        return ;
+                    }
+                }
+            }
+            get_last_window ().present_with_time ((uint32) now.to_unix ());
```
Marukesu, line 78:
```suggestion
            get_last_window ().present_with_time (Gtk.get_current_event_time ());
```

src/Application.vala
igordsm: @Marukesu Does this work on your end? It doesn't for me. From what I've read, the  timestamp that `present_with_time` receives must be a X11 server timestamp and the only way I was able to get it was using this Gdk.X11 call. I've tried `Gdk.CURRENT_TIME` and many types of datetime functions returning the current time as well. 

I'm not a fan of adding this dependency too, but it was literally the only timestamp that worked for me.

jeremypw:
Might be worth delaying trying to fix this until https://github.com/elementary/notifications/issues/153 is fixed. It should not be necessary to add that extra dependency.

jeremypw:
Need to find a solution that works on Wayland.

src/Application.vala
jeremypw: I can confirm that `current_event_time` does not work on X11 or Wayland, probably because the event did not occur on the Terminal window but was received over DBus.  I have suggested a hack to get it work on X11 at least with that extra dependency.

jeremypw:
@Marukesu  I think a dependency on `gdk-wayland-3.0` may be required to make it work under Wayland.  This should not be necessary under Gtk4, so maybe better to wait for the Gtk4 port to land to fix Wayland?   If so, maybe this can be merged so that it at least works on OS7.1 and in "Compatability Mode".

Marukesu:
I think it should be fine for what the notification sever can do right now.

We really need to have the Gtk Notification interface implemented there to have the true timestamp of the event, but that can be done later.

jeremypw (approved):

PR #639, opened 2021-11-09
UnsafePasteDialog: use paste icon

danrabbit:
Not 100% sure if this is better or worse, but figured I'd throw it out there:

## Before

Screenshot from 2021-11-09 14 28 59

## AFTER

Screenshot from 2021-11-09 14 28 36

### DIFF
--- a/src/Dialogs/UnsafePasteDialog.vala
+++ b/src/Dialogs/UnsafePasteDialog.vala
@@ -28,7 +28,8 @@ public class Terminal.UnsafePasteDialog : Granite.MessageDialog {
     }
 
     construct {
-        image_icon = new ThemedIcon ("dialog-warning");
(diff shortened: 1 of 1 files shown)


jhaygood86:
The biggest change I would make here honestly is swapped the suggested action, to make it "easier" to do the safe thing.

danrabbit:
@jhaygood86 that's not really inside the scope of this PR. I'm only changing the icon here.

But also, that would kind of go against our HIG and could lead to misclicks. We want to make sure the "Cancel" action is in a consistent place

vjr:
How about swapping the icons? Keep the alert icon as before and make the paste icon as the smaller overlay?

jeremypw:
For me, it makes the icon over-complicated and reduces the impact of the dialog.

PR #647, opened 2021-12-31
Display character size on dragging window edges

### LINKED ISSUE #443: Display size while dragging windows edges
## Prerequisites

## Feature
While increasing the window size, I'd like to have an overlay in the middle of the terminal window which displays the current window size.

**Existing work**
The Ubuntu Desktop default terminal already does that.

Roshan-R:
Should resolve #443 

things to address upon : 

- At present, it connects the size-allocate signal of the grid to call the function to show size, this could probably be improved by connecting to some other signal 
- It shows the character size upon first launch
- Improve Design

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -521,7 +521,7 @@ namespace Terminal {
             var tab_bar_behavior = Application.settings.get_enum ("tab-bar-behavior");
             notebook.tab_bar_behavior = (Granite.Widgets.DynamicNotebook.TabBarBehavior)tab_bar_behavior;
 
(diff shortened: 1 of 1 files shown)


jeremypw:
@Roshan-R : This now conflicts with master so needs attention. Also, the character size of the window will change when the font size is changed - this does not seem to handle that?  @danirabbit Is this something worth pursuing? I am not sure whether a Gtk4 version of Terminal is on the roadmap but this solution probably will not work as is.

jeremypw:
Closing as there has been no progress with this for a long time.

PR #650, opened 2022-01-09
MainWindow: Switch tab with Alt+[1-9] when there are at least 2 tabs

JamiKettunen:
This allows `Alt+[1-9]` to be handled by applications such as WeeChat (with just one tab open) while functioning like before when another is opened (previously `Alt+1` and `Alt+9` were completely unusable within WeeChat hindering its featureset).

Without more than one tab open it doesn't make sense attempting to handle the `Alt+1` and `Alt+9` shortcut actions anyway.


jeremypw (approved):
Makes sense.

It is not clear why [Alt+9] is handled separately (any number greater than the number of tabs could select the last tab) and also the code does not exclude another modifier also being pressed but those issues are present in master and can be fixed separately

Issue #507, opened 2020-04-24
Add custom terminal profile


dagolinuxoid:
Let's say I want to use a custom 'Nord' theme for the elementary terminal. How to achieve it? At the moment the terminal has three radio buttons: *high contrast*, *solarized light*, *dark*. Clicking on corresponding button changes only `foreground` and `background` to some default defined value for each button. Right now I can define a theme 
```bash
#!/bin/bash

# Nord_ISH theme
# palette="#3b4252:#bf616a:#a3be8c:#ebcb8b:#81a1c1:#b48ead:#88c0d0:#e5e9f0:#4c566a:#bf616a:#a3be8c:#ebcb8b:#81a1c1:#b48ead:#8fbcbb:#eceff4"
foreground="#eceff4"
background="rgba(46, 52, 64, 0.95)"

gsettings set io.elementary.terminal.settings foreground "$foreground"
gsettings set io.elementary.terminal.settings background "$background"
gsettings set io.elementary.terminal.settings palette "$palette"
```
and run such `./setMyTheme.sh` to change *palette* permanently. The problem is as soon as I click on the radio buttons they dispatch setting of some default background and foreground settings overriding the ones described in a custom theme. Actually I think it's a good thing. The question is how can I change the default values for only *dark* radio button to values of my custom defined colors for *foreground and background* to make it permanent?
The other solution would be to consider adding a new theme (fourth, fifth ... radio button with a custom name something similar to gnome-terminal profiles feature). But I'm a bit skeptical about the approach: three buttons is more than enough.

jeremypw:
See #418

danrabbit:
I believe this is now fixed. Custom theme settings now can be selected in the UI and persist even when switching between default themes

Issue #501, opened 2020-04-08
Filename not readable when permissions are 777


rajkumaar23:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Describe the bug
The filenames are not at all readable when the permissions are 777 (applies to all the 3 themes provided)
![image](https://user-images.githubusercontent.com/37476886/78750727-75639980-798e-11ea-83a0-ebdd27c0833a.png)

### To Reproduce
1. Go to terminal
2. `chmod` one or two files to `777`
3. Do an `ls -a`
4. See the filenames which are with `777` being not readable.

### Expected behavior
The filenames should be displayed in a better color.

## Platform Information

- [ ] I'm using the latest version from git that I've manually compiled
- [x] I'm using the latest released stable version

jeremypw:
You can temporarily turn off the coloration with `ll --color=none` but need to change the color palette of the terminal for a permanent fix.  This is initially set in the code but can be modified by changing the setting `io.elementary.terminal.settings.palette`  However, it is not obvious how to change it for the better manually. I agree that the color contrast for this combination is not great.  I assume it is the `ls` command that chooses which pair of colors to choose (blue on green) so it is never going to be very readable for color-blind people.

jeremypw:
You can improve things by lightening the 3rd item in the palette (green) (e.g. :#99BB22) and darkening the 5th item (navy) (e.g. #2626d2).  Need design team to specify acceptable default colors for elementary.

rajkumaar23:
@jeremypw Cool. Thanks for that. Just edited with `dconf-editor` and it looks better. 😁️

dougabugg:
I ran into this issue as well, I tried changing the terminals theme from the gear icon menu, but it just changed the background, not the color of the text. This app is completely unusable for me.

jeremypw:
see #418

Issue #71, opened 2017-04-26
Syntax highlighting makes text unreadable because of color scheme


elementaryBot:
The syntax highlighting in some text editor (I'm using vi in the screenshot) makes the code almost unreadable.

UPDATE: This happens only when working remotely, via SSH.

Launchpad Details: [#LP1634711](https://bugs.launchpad.net/bugs/1634711) Antonio Camargo - 2016-10-19 01:07:04 +0000

![Screenshot from Launchpad](https://launchpadlibrarian.net/290005378/Screenshot%20from%202016-10-18%2022:44:48.png)

Issue #39, opened 2017-04-26
Terminal unusable with mc due to bad standard color scheme


elementaryBot:
Well, I had to use Gnome terminal until elementary's terminal has some sort of setup functionality (or profiles like Gnome terminal). At the moment the terminal is unusable due to the standard colors - this makes the blue color scheme of mc unreadable. And, believe it or not, mc is THE most used app here and I assume I'm not the only one.

Launchpad Details: [#LP1368893](https://bugs.launchpad.net/bugs/1368893) Umeneri - 2014-09-12 18:03:49 +0000

elementaryBot:
I'm not sure what mc is, but I agree the default theme is not good enough for active use. Luckily you can change the color scheme using the dconf-editor.

Here's an example of a nice gnome-like theme with a touch of elementary in it.

org.pantheon.terminal.settings.palette:
#303030:#e1321a:#6ab017:#ffc519:#165a9e:#ec2361:#2aa7e7:#d1d1d1:#5d5d5d:#ff4933:#74bd1e:#ffd526:#0071ff:#eb2a65:#4bb8fd:#a020f0

org.pantheon.terminal.settings.foreground:
#a9b6b8﻿

Launchpad Details: [#LPC](https://bugs.launchpad.net/pantheon-terminal/+bug/1368893/comments/1) Samu Nuutamo - 2014-10-09 15:40:17 +0000

elementaryBot:
Hi Samu

mc is the Midnight Commander -  a tool which I use frequently in a terminal. Actually a Linux without mc is no Linux ;)

(try sudo apt-get install mc)

And thanks so much for the colour codes, I'll try them.


-----Ursprüngliche Nachricht-----
Gesendet: Donnerstag, 09 Oktober 2014 um 17:40:17 Uhr
Von: "Samu Nuutamo" <email address hidden>
An: <email address hidden>
Betreff: [Bug 1368893] Re: Terminal unusable with mc due to bad standard color scheme
I'm not sure what mc is, but I agree the default theme is not good
enough for active use. Luckily you can change the color scheme using the
dconf-editor.

Here's an example of a nice gnome-like theme with a touch of elementary
in it.

org.pantheon.terminal.settings.palette:
#303030:#e1321a:#6ab017:#ffc519:#165a9e:#ec2361:#2aa7e7:#d1d1d1:#5d5d5d:#ff4933:#74bd1e:#ffd526:#0071ff:#eb2a65:#4bb8fd:#a020f0

[...]

PR #662, opened 2022-07-01
Add a custom theme dialog

### LINKED ISSUE #418: User-provided color scheme

jeremypw:
Pulled from #552 
Fixes #418 

Potentially address #39 
Potentially addresses #71
Potentially address #501 
Partially addresses #507 
Potentially address #632 

Additional work:
 * [x] Fix appearance of dialog under RTL
 * [x] Add a reset to default palette button
 * [x] Name variables according to linked nominal color
 * [x] Add labels for light colors
 * [x] Tweak layout and add some margins  
 * [x] Persist user custom theme after switching to a builtin theme
 * [ ] Import a theme

### DIFF
--- a/src/Dialogs/ColorPreferencesDialog.vala
+++ b/src/Dialogs/ColorPreferencesDialog.vala
@@ -72,9 +72,9 @@ public class Terminal.Dialogs.ColorPreferences : Gtk.Dialog {
             margin_bottom = 12,
             tooltip_text = _("Reset to elementaryos default color palette")
         };
-        default_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
         default_button.clicked.connect (() => {
             Application.settings.set_string ("palette", string.joinv (":", HEX_PALETTE));
+            update_buttons_from_settings ();
         });
 
         var black_color_label = new SettingsLabel (_("Black:"));
--- a/src/Dialogs/ColorPreferencesDialog.vala
+++ b/src/Dialogs/ColorPreferencesDialog.vala
@@ -1,0 +1,310 @@
+/*
+* Copyright (c) 2020 elementary, Inc. (https://elementary.io)
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU Lesser General Public
+* License version 3, as published by the Free Software Foundation.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+public class Terminal.Dialogs.ColorPreferences : Gtk.Dialog {
+    private Gtk.ColorButton color01_button; // Black
+    private Gtk.ColorButton color02_button; // Red
+    private Gtk.ColorButton color03_button; // Green
+    private Gtk.ColorButton color04_button; // Yellow
+    private Gtk.ColorButton color05_button; // Blue
+    private Gtk.ColorButton color06_button; // Magenta
+    private Gtk.ColorButton color07_button; // Cyan
+    private Gtk.ColorButton color08_button; // Light gray
+    private Gtk.ColorButton color09_button; // Dark gray
+    private Gtk.ColorButton color10_button; // Light Red
+    private Gtk.ColorButton color11_button; // Light Green
+    private Gtk.ColorButton color12_button; // Light Yellow
+    private Gtk.ColorButton color13_button; // Light Blue
+    private Gtk.ColorButton color14_button; // Light Magenta
+    private Gtk.ColorButton color15_button; // Light Cyan
+    private Gtk.ColorButton color16_button; // White
+    private Gtk.ColorButton background_button;
+    private Gtk.ColorButton foreground_button;
+    private Gtk.ColorButton cursor_button;
+    private Gtk.Image contrast_image;
+
+    public ColorPreferences (Gtk.Window? parent) {
+        Object (
+            deletable: false,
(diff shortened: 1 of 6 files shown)


jeremypw:
@danrabbit I would appreciate your comments on this. I added the extra labels as it didn't seem obvious to a new user why there were two very similar color buttons with only one label.

The main flaw with this is that it is not possible to save a custom theme permanently - if you switch back to a builtin theme it overwrites your custom theme and you cannot return to it without rebuilding it.  It might be better to fix that in a new PR as there is more than one possible solution - e.g. a new setting or a file in the user home.

jeremypw:
I see from the prior art in #418 that no-one else thinks the extra labels are necessary so I'll remove if required.

jeremypw:
There are also some possible extra refinements like using palette colors for foreground/background but that can be done in another PR.

danrabbit:
I think we should solve the issue of persisting the custom palette before exposing this toggle so that we don't have an immediately broken feature. We should probably make sure that the `palette` setting is only used to store values for the custom theme and that Terminal is loading the palette for built-in themes from the Theme class and not from gsettings

jeremypw:
Yeh, that makes sense.

jeremypw:
@danrabbit  After merging master and making relevant changes, the custom theme now persists. I have moved the "Color Palette" header to before the background/foreground buttons because although the palette strictly speaking does not include these, to the average user they are just some more color settings.  I have also changed the reset button to reset the foreground/background/cursor colors for the same reason.

Can importing/saving custom themes be left for a later PR, if required?

src/Themes.vala
```diff
@@ -40,12 +40,34 @@ public class Terminal.Themes {
...
+    }
+
+    // DARK_PALETTE is used as a fallback when custom palette colors are invalid
+    public const string[] DARK_PALETTE = {
+        "#073642", "#dc322f", "#859900", "#b58900", "#268bd2", "#ec0048", "#2aa198", "#94a3a5",
+        "#586e75", "#cb4b16", "#859900", "#b58900", "#268bd2", "#d33682", "#2aa198", "#6c71c4",
+        "rgba(46, 46, 46, 0.95)", "#a5a5a5", "#839496"
+    };
+    public const int PALETTE_SIZE = 19;
+    public static string[] get_string_palette (string theme) {
```
danrabbit, line 70: Does this need to be public?

src/Themes.vala
```diff
@@ -40,12 +40,34 @@ public class Terminal.Themes {
...
+            }
+
+            rgba_palette[i] = new_color;
+        }
 
+        return rgba_palette;
+    }
+
+    // DARK_PALETTE is used as a fallback when custom palette colors are invalid
+    public const string[] DARK_PALETTE = {
```
danrabbit, line 64: Since we have `get_string_palette` now, should this not be a constant and have `get_rgba_palette` use `get_string_palette (prefer_dark ? DARK : LIGHT)` for its fallback?

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -0,0 +1,353 @@
+/*
+* Copyright (c) 2020 elementary, Inc. (https://elementary.io)
```
danrabbit, line 2:
```suggestion
* Copyright 2022 elementary, Inc. (https://elementary.io)
```

danrabbit (changes requested):
Two small questions, but other than that totally happy with this as an initial implementation. Let's get this in and then smooth out the flow a little bit/do other features in future branches

src/Themes.vala
jeremypw: This is possible only if we replace the *whole* of the palette if *any* of the custom colors is invalid.  Is that OK? At the moment, we only replace individual colors.

jeremypw:
I have implemented your suggestion re fallback palette as it simplifies the code and the issue of possible invalid colors in the settings (presumably due to manual editing) is unlikely.  We now fallback to a default palette if *any* of the custom colors are invalid rather than replacing only the invalid color.

src/Themes.vala
```diff
@@ -40,12 +41,27 @@ public class Terminal.Themes {
...
+
+        var rgba_palette = new Gdk.RGBA[PALETTE_SIZE];
+        for (int i = 0; i < PALETTE_SIZE; i++) {
+            var new_color = Gdk.RGBA ();
+            // If custom palette invalid use a fallback one
+            if (!new_color.parse (string_palette[i])) {
+                critical ("Color %i '%s' is not valid - using fallback palette", i, string_palette[i]);
+                return get_rgba_palette (
+                    Application.settings.get_boolean ("prefer-dark-style") ? DARK : LIGHT
+                );
```
danrabbit, line 55: Couldn't we do this?

```suggestion
                critical ("Color %i '%s' is not valid - replacing with default", i, string_palette[i]);
                settings_valid = false;

                var fallback_palette = get_string_palette (
                    Application.settings.get_boolean ("prefer-dark-style") ? DARK : LIGHT
                );
                string_palette[i] = fallback_palette[i];
                new_color.parse (fallback_palette[i]);
```

src/Themes.vala
```diff
@@ -63,37 +79,30 @@ public class Terminal.Themes {
...
-                settings_valid = false;
-            }
-
-            rgba_palette[i] = new_color;
-        }
-
-        if (!settings_valid) {
-            /* Remove invalid colors from setting */
-            Application.settings.set_string ("palette", string.joinv (":", string_palette));
-        }
```
danrabbit, line 95: I don't think this was meant to be removed. This overwrites the setting with the sanitized palette if we detect that any of the values are invalid

danrabbit (changes requested):
One suggestion and a comment

jeremypw:
Ok, so you want to keep correcting individual colors; I'll revert to that. I just thought that this pretty unlikely to be required in practice so simplified it. I guess if we implementing importing customs themes it raises the risk of invalid colors being imported.

src/Themes.vala
jeremypw: Reinstated

danrabbit (approved):
Yeah it's just a little bit less destructive. This looks good!

PR #663, opened 2022-07-01
Perpetuate custom theme

danrabbit:
Solves the problem presented in #662 that custom themes are overwritten by the default ones.

Also spend less time reading/writing to settings hopefully


danrabbit:
Converting back to draft for a second because there's some obvious simplifications that can be made

jeremypw (commented):
Apart from the first time of running, when the selected theme in the popover did not match the visuals, this works as expected.  I do not think the initial glitch is significant - maybe due to previously working on the custom palette other PR.

Is there a reason why only one of the builtin themes is stored as as a constant string array?

danrabbit:
Ah yeah I bet that is because the theme switcher doesn't currently handle the custom theme, so it should automatically be resolved once we add a switcher item for it.

Yes, because it's used to replace invalid values in the custom theme. That's diff lines 82-87

jeremypw (approved):

PR #665, opened 2022-07-06
MainWindow: Style color buttons from theme

danrabbit:
Style color buttons using theme data instead of hardcoding values


jeremypw (commented):
This works as expected and the code looks good but the problem is, if the custom theme has the same or very similar background and foreground colors, you end up with two buttons looking the same.  Is there some way of making the builtin and custom theme buttons distinguishable other than by tooltip?

jeremypw:
Screenshot from 2022-07-07 10 04 50

PR #666, opened 2022-07-06
Add "Follow System Style" toggle

### LINKED ISSUE #618: Terminal theme should adapt current active system theme
## What Happened
Changed system theme (light vs. dark) in eOS 6 to be dependent on sunset/sunrise
As it was daytime, system theme changed to light but terminal stayed dark.

## Expected Behavior
Terminal automatically switches to light theme, too

## Steps to Reproduce
See above

## Platform Information
ElementaryOS 6

jeremypw:
Fixes #618

The UX is copied from #546 

The mechanism for getting the widgets to update their color scheme when the system default changes is somewhat non-obvious (the application settings are signaled to have changed even though they haven't) so as to make this work with the existing mechanism which requires an application settings change.

### DIFF
--- a/data/io.elementary.terminal.gschema.xml
+++ b/data/io.elementary.terminal.gschema.xml
@@ -85,6 +85,11 @@
       <summary>Whether or not to use an audible bell event alert</summary>
       <description>Used when hitting the end of a line like in GTK entries, but also for tab-completion when there are either no or multiple possible completions.</description>
     </key>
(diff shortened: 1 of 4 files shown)


jeremypw:
Screenshot from 2022-07-07 19 46 21
Screenshot from 2022-07-07 19 45 59

danrabbit (approved):

danrabbit (approved):

PR #668, opened 2022-07-06
MainWindow: Put search toolbar in a title stack

danirabbit:
Instead of adding a whole new toolbar, re-use this title area that has nothing else in it

Screenshot from 2022-07-06 14 13 56


jeremypw (changes requested):
There is a noticeable visual glitch on startup caused by the title label being revealed.  Apart from that this works as expected and the code looks good.

jeremypw (dismissed):

jeremypw (approved):
Not sure why this wasn't merged before. Code good and no glitch now.

PR #671, opened 2022-07-07
MainWindow: Update active color button on theme change

### LINKED ISSUE #424: AppMenu out of sync after external settings change to background color
- [* ] I have searched open and closed issues for duplicates.

## Describe the bug
If the background color is changed externally while Terminal is running, the AppMenu color chooser does not update (although the background does).

### To Reproduce
1. Open Terminal
2. Choose a light or solarized color scheme
3. Edit settings using dconf-editor or otherwise
4. Change "io.elementary.terminal.settings.background" to the default (dark) value.
5. The terminal background goes dark.
6. Open the App Menu.  The light/solarized choice is still ticked.

### Expected behavior
The tick should appear in the item reflecting the actual background

elementary OS
Juno daily
- [* ] I'm using the latest version from git that I've manually compiled

danirabbit:
Fixes #424


jeremypw (approved):
This works as expected and the code looks good.  However, it does not fix the linked issue exactly.  The settings 'foreground' and 'background' and 'cursor-color' now only affect the custom palette and this already updates correctly on external change so the issue can be closed anyway.

PR #673, opened 2022-07-07
ColorPreferencesDialog: Add contrast checking for all main colors

danrabbit:
Screenshot from 2022-07-07 14 18 59

Adds contrast checking for all of the main colors.

I was originally trying to add contrast checking for all colors, but it's not honestly realistic and even this doesn't give a complete picture. It looks like the way the palette is defined is pairs of "Dark foo/Light foo", which is not the same as "Background foo, Foreground foo".  BASH and it seems most commands like `ls` use the "Dark foo" colors as foreground colors, so since that's the probably primary usage for most people, I imagine we want to optimize for that case.

It seems like maybe there was an intention that all of the main colors are expected to be able to be used as "Background foo" and all the light colors as "Foreground foo", as seen in apps like `mc`, but in practice you can basically expect any color to appear on top of any other color, so ensuring contrast is basically impossible (as seen below).

Screenshot from 2022-07-07 14 24 41

I know stuff like "Dark white" and "light black" sounds dumb, but it helps keep the code more clear since that's how the colors are defined in the Xresources format and we can use only the user-facing strings to paper over that with more human-sounding names


jeremypw:
Not entirely convinced of the utility of this as most contrast problems occur *between* palette colors not with the background - and there will always be at least one palette color with a warning against it unless you assign a color value that is different from the color name (e.g. assign a light color to "Black" when there is a black background).  Also custom palettes are presumably being personally tailored to an individual user's vision and preference (and which programs they run) so an objective test may not be correct.  If you are designing a public theme then you would probably not use Terminal to do it.

To be honest I was confused about how to label the palette colors as there is nothing preventing the user assigning a color completely different to the name.  But using names like "Color 1", "Light color 2" is not very helpful either :shrug:

danrabbit:
Gonna close since this is of questionable utility

PR #674, opened 2022-07-12
TerminalWidget: Connect specific settings to functions

danirabbit:
Make sure we only write changes when relevant settings change

### DIFF
--- a/po/he.po
+++ b/po/he.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: pantheon-terminal\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2022-07-12 22:12+0000\n"
-"PO-Revision-Date: 2022-07-09 17:13+0000\n"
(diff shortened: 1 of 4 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -293,12 +314,6 @@ namespace Terminal {
                 }
             }
 #endif
-
-            /* Disable bell if necessary */
-            audible_bell = Application.settings.get_boolean ("audible-bell");
-
-            /* Cursor shape */
-            set_cursor_shape ((Vte.CursorShape) Application.settings.get_enum ("cursor-shape"));
```
zeebok, line 301: I get pulling this out into a new function but the way it is written now, being called in the constructor, those update functions won't be called when Application.settings.changed is signaled. Is that alright?

zeebok (approved):
Just have the one question, otherwise this looks good

src/Widgets/TerminalWidget.vala
danrabbit: No they should only be called when the specific settings change that require a new set here

PR #675, opened 2022-07-12
Require libvte >=0.59

danrabbit:
Drop support for old VTE. Focal has 0.60


tintou (dismissed):

jeremypw:
After removing the pre libvte-0.60 code in master, `restore_settings` becomes empty so I have removed it.

danrabbit:
Thanks @jeremypw!

tintou (approved):

PR #679, opened 2022-08-12
build: Drop unused ubuntu-bionic-patched-vte option

bobby285271:
Grepping this does not return other result for me, probably no longer useful after #675 (yeah we carry this option in nixpkgs since initial packaging so it is really nice to see we can drop this)


jeremypw (approved):
Yes, looks like the previous PR you mentioned removed all reference to this option.  No problem compiling and running on EOS6.1 and EOS7

PR #680, opened 2022-08-14
Fix foreground process shortcuts

### LINKED ISSUE #676: Keybinds for micro text editor do not work, making it less efficient to use the editor
### What Happened?

keybinds like ctrl+shift+right don't work

### Steps to Reproduce

start micro editor
use keybinds ctrl+shift+right/left

### Expected Behavior

keybinds should work

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #676

When a foreground process is current all key presses received by the MainWindow are passed straight to the terminal (MainWindow actions are therefore inhibited).  As a consequence, the search entry and button are also made insensitive.

This has the drawback that MainWindow actions are inhibited whether or not the foreground uses any of the same shortcuts as MainWindow but there is no obvious way of determining whether or not a shortcut has been used by a foreground process. But as all actions can be initiated with the mouse this does not seem too much of an issue.


danrabbit (commented):
Hm, I'm not sure about the changes to the searchbar here, especially since if you've opened the searchbar it's now stuck open until the foreground process ends. It also seems like we'll get issue reports about not being able to create or switch tabs while a command is running.

I wonder if we should expand the behavior of the current "Natural Copy/Paste" toggle to have more of a "CLI app mode" or have a per-tab option to disable all app shortcuts while the current tab is focused. Or just generally something that feels more deliberate.

jeremypw:
OK, thanks for the review - I'll rethink it.

jeremypw:
>  It also seems like we'll get issue reports about not being able to create or switch tabs while a command is running.
You can always create a new tab or switch tabs with the mouse.  If we allow the foreground process to receive all key presses first (which seems necessary to fix the issue) then there is likely to be this issue, especially if the process eats the key event.

jeremypw:
@danrabbit One solution may be to special case certain key combinations such as those that switch tabs or create a new tab?  These would not be passed to the foreground process.  Let me know which actions should be special cased and I'll try and implement it.

jeremypw:
It doesnt really make sense to perform a search on a tab that has a foreground process so it seems reasonable to make the searchbar insensitive when such a tab is focused?

jeremypw:
Closing as linked issue has been closed.

Issue #11, opened 2017-04-26
Make findbar arrows insensitive while no matches


elementaryBot:
what it says on the tin. If there are no matches (especially if there are no search terms), make the next/previous arrows insensitive.

Launchpad Details: [#LP1426470](https://bugs.launchpad.net/bugs/1426470) Daniel Fore - 2015-02-27 17:25:00 +0000

jeremypw:
Because the terminal widget does search only as far as the next (or previous) hit it proved intractable to set the arrow sensitivities reliably under all conditions. In my opinion the complexity of the code required is not justified by a small cosmetic change.

jeremypw:
If anyone is able to provide a compact reliable solution they are welcome to re-open this.

PR #687, opened 2022-09-18
#11: Findbar arrow sensitivity

hollyschilling:
Added changes to the sensitivity of next and previous buttons when no results are found.


jeremypw:
@hollyschilling The linter is `vala-lint` with the default configuration. You can download and install locally from https://github.com/vala-lang/vala-lint.  You can then check you code locally before committing.

jeremypw:
Further info on elementary code style, which the linter enforces can be found here: https://docs.elementary.io/develop/writing-apps/code-style

There should probably be info about linting locally there but as yet it has not been added.

hollyschilling:
I'm sorry to keep bugging you for simple questions, but I'm having an error with the linting that I can't understand. Specifically, I received and error stating: 

```
  117.26    error   Expected variable name in underscore_convention   naming-convention
```

The referenced line of code is `bool initialFound = next_search ();`. I can't understand why this declaration would be preceded with an underscore. For context, the line is in a try block, in a lambda, in the constructor. What is the rule that is triggering this to get that error?

lenemter:
@hollyschilling in Vala variables are named using snake_case, and `initialFound` should be `initial_found`

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
-        public void previous_search () {
+        public bool previous_search () {
             window.current_terminal.search_set_wrap_around (cycle_button.active);
-            window.current_terminal.search_find_previous ();
+            bool found = window.current_terminal.search_find_previous ();
+            window.get_simple_action (MainWindow.ACTION_SEARCH_PREVIOUS).set_enabled (found);
+
+            if (cycle_button.active) {
+                // If CYCLE is enabled then NEXT and PREVIOUS should match
+                window.get_simple_action (MainWindow.ACTION_SEARCH_NEXT).set_enabled(  found);
```
jeremypw, line 139:
```suggestion
                window.get_simple_action (MainWindow.ACTION_SEARCH_NEXT).set_enabled (found);
```

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
+            } else if (found) {
+                // If CYCLE is disabled then enable NEXT if search succeeded
+                window.get_simple_action (MainWindow.ACTION_SEARCH_NEXT).set_enabled(  true);
```
jeremypw, line 142:
```suggestion
                window.get_simple_action (MainWindow.ACTION_SEARCH_NEXT).set_enabled (true);
```

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
+            }
+            return found;
```
jeremypw, line 144:
```suggestion

            return found;
```

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
-        public void next_search () {
+        public bool next_search () {
             window.current_terminal.search_set_wrap_around (cycle_button.active);
-            window.current_terminal.search_find_next ();
+            bool found = window.current_terminal.search_find_next ();
+            window.get_simple_action (MainWindow.ACTION_SEARCH_NEXT).set_enabled (found);
+
+            if (cycle_button.active) {
+                // If CYCLE is enabled then NEXT and PREVIOUS should match
+                window.get_simple_action (MainWindow.ACTION_SEARCH_PREVIOUS).set_enabled(  found);
```
jeremypw, line 154:
```suggestion
                window.get_simple_action (MainWindow.ACTION_SEARCH_PREVIOUS).set_enabled (found);
```

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
+            } else if (found) {
+                // If CYCLE is disabled then enable PREVIOUS if search succeeded
+                window.get_simple_action (MainWindow.ACTION_SEARCH_PREVIOUS).set_enabled(  true);
```
jeremypw, line 157:
```suggestion
                window.get_simple_action (MainWindow.ACTION_SEARCH_PREVIOUS).set_enabled (true);
```

src/Widgets/SearchToolbar.vala
```diff
@@ -126,14 +129,34 @@ namespace Terminal.Widgets {
...
+            }
+            return found;
```
jeremypw, line 159:
```suggestion
            return found;
```
```suggestion

            return found;
```

jeremypw (changes requested):
I have suggested some corrections to remaining lint.

jeremypw:
A side-effect of this change is that if, after the search, the terminal produces output containing the search term then the navigation buttons remain insensitive even if the cyclic search button is toggled.  The search term has to be modified or re-entered in order to initiate a new search.  A more convenient way of refreshing the search would be good. At least if the cyclic search button is toggled then the sensitivity of the navigation buttons should be revisited.

src/Widgets/SearchToolbar.vala
hollyschilling: What is this change?

lenemter: A blank line before return

hollyschilling:
I completely agree that this is potentially undesirable behavior. Is there an event emitted from the terminal that would indicate that additional content has been added? 

There are a few such edge cases that are not handled well. I feel that adding messaging to indicate that the end of search has been reached would be appropriate.

jeremypw:
> Is there an event emitted from the terminal that would indicate that additional content has been added?

There is `Vte.Terminal.content-changed` signal.  But you may need to throttle that so that a large number of searches are not performed unnecessarily when the content is changing rapidly.  Maybe better leave that for another PR.

src/Widgets/SearchToolbar.vala
```diff
@@ -114,7 +114,10 @@ namespace Terminal.Widgets {
                      */
                     var regex = new Vte.Regex.for_search (GLib.Regex.escape_string (search_term), -1, PCRE2.Flags.CASELESS | PCRE2.Flags.MULTILINE);
                     term.search_set_regex (regex, 0);
-                    next_search (); /* Search immediately - not after ENTER pressed */
+                    bool initial_found = next_search (); /* Search immediately - not after ENTER pressed */
+                    if (!initial_found && !cycle_button.active) {
```
jeremypw, line 118: As `initial_found` is only used once we can omit it:
```suggestion
                     /* Search immediately - not after ENTER pressed */
                    if (!next_search () && !cycle_button.active) {
```

jeremypw (changes requested):
Another small bug is that if cyclic search is OFF and you navigate to the last found term so that the `next` button becomes insensitive and then toggle cyclic search ON then the sensitivity of the `next` button does not change - it should become sensitive.

A fix for this would be to emit a `search-changed` signal whenever the cyclic search is toggled.

jeremypw:
I looked into completing this but found that because of the way the terminal widget does search one hit at a time (unlike Code which searches for all hits within scope) it is extremely difficult/impossible to set the search arrow sensitivities correctly all the time.  I do not think it is worth spending more time on it tbh.

PR #690, opened 2022-10-08
Use fc to get previous command

### LINKED ISSUE #689: Terminal notification displays timestamp when HISTTIMEFORMAT variable is set
### What Happened?

The notification for terminal displays the timestamp as part of the message when the HISTTIMEFORMAT bash variable is set.

Captura de pantalla de 2022-09-26 21-53-55

I believe this particular function is relevant to this issue

https://github.com/elementary/terminal/blob/5a9ac9197e89191c1477a3ae0d7d694dd152365b/src/Application.vala#L98

### Steps to Reproduce

For reproducing this bug, the HISTTIMEFORMAT variable needs to be set to anything. And then wait for the notification popup for a finished background process.

### Expected Behavior

The expected behavior is the notification body not including the timestamp.

### OS Version

6.x (Odin)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

davidmhewitt:
Fixes #689 

`fc` is a BASH builtin than we can use as an alternative to `history` that will give us some more consistently formatted results so we don't accidentally end up with other bits of `history` output in our notifications.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -73,7 +73,7 @@ namespace Terminal {
                                                   "/io/elementary/terminal " +
                                                   "io.elementary.terminal.ProcessFinished " +
                                                   "string:$PANTHEON_TERMINAL_ID " +
-                                                  "string:\"$(history 1 | cut -c 8-)\" " +
+                                                  "string:\"$(fc -nl -1 | cut -c 3-)\" " +
                                                   "int32:\$__bp_last_ret_value >/dev/null 2>&1";
 
         /* Following strings are used to build RegEx for matching URIs */


tintou:
How does this work with Dash and others?

davidmhewitt:
It's a builtin with the same syntax in dash as well.

Considering bash is our default, this variable is called `SEND_PROCESS_FINISHED_BASH `, and we have separate config and methods for handling this in fish and zsh, I'd say we have pretty good coverage of popular shells.

davidmhewitt:
Actually, I don't think other shells matter here at all, because this command only gets used in the `PROMPT_COMMAND` environment variable, which as far as I can tell is bash specific.

tintou (approved):

PR #691, opened 2022-10-11
TerminalWidget: Fix terminal freeze when closing in GLib 2.73.2+

### LINKED ISSUE #688: GLib 2.73.2+: Terminal freezes when closing with the close button
### What Happened?

I have a terminal with only one tabs left and nothing except zsh itself running and I want to close the terminal with the close button (both the one from the tab and the one from the window is okay), but the terminal freezes.

After some quick scan of the code, insert some debug lines and some rebuild etc. I noticed that it freezes because of:

https://github.com/elementary/terminal/blob/5a9ac9197e89191c1477a3ae0d7d694dd152365b/src/Widgets/TerminalWidget.vala#L307-L316

I actually insert this line above `Thread.usleep (100);`:

```vala
warning (this.child_pid.to_string ());
```

to get the pid and ran e.g. `ps -Al | grep 85231` and it is zsh and it is a zombie process, so we cannot get rid of the while loop and the terminal freezes. Unfortunately, after changing zsh to bash, I can still trigger the issue in the same way so I actually don't think it is related to what shell I am using.

I was aware that many years ago, there is a commit https://github.com/elementary/terminal/commit/ad127db5fc6088eba5379c4d8002074e83054bb7 that introduces this situation, and reverting https://github.com/elementary/terminal/commit/ad127db5fc6088eba5379c4d8002074e83054bb7 of course resolves the issue for me since we can break the loop eventually.

Also by using Ctrl + D or the `exit` command you can still close the terminal without issues.

It can be very possible that this can be cannot reproduced on elementary OS because I only saw this issue recently on a NixOS unstable + Pantheon system with GNOME 43 packages shipped. Edit: later we noticed that this only happens with GLib 2.73.2+.

Thanks for reading this!

### Steps to Reproduce

- Build the terminal with GLib 2.73.2+
- Open the terminal
- Close the terminal with the close button

### Expected Behavior

The terminal closes and no freeze.

### OS Version

Other Linux

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

bobby285271:
GLib 2.73.2 [dropped](https://gitlab.gnome.org/GNOME/glib/-/commit/f615eef4bafaa2f) the global SIGCHLD handler, we need to wait the shell process ourself

Closes #688

/cc @jtojnar


bobby285271:
Tested closing the terminal with kernel 4.19 and 6.0

jeremypw (approved):
Unable to fully test but does not cause any regression on Horus

PR #699, opened 2023-01-28
Add `SettingsPopover` Class

Marukesu:
move the settings menu popover, and related functions to they own class.

the settings object is now, also used to keep the state between the widgets synchronized, and the actions provided by it make the new class mostly self-contained, only needing a single signal to indicate the window that the theme dialog should be shown.

### DIFF
--- a/src/Widgets/SettingsPopover.vala
+++ b/src/Widgets/SettingsPopover.vala
@@ -1,0 +1,204 @@
+/*
+ * Copyright 2023 elementary, Inc (https://elementary.io)
+ * SPDX-License-Identifier: LGPL-3.0-only
+ */
+
+public sealed class Terminal.SettingsPopover : Gtk.Popover {
+    public signal void show_theme_editor ();
+
+    public Vte.Terminal terminal {
+        owned get {
+            return terminal_binding.source as Vte.Terminal;
+        }
+
+        set {
+            terminal_binding.source = value;
+        }
+    }
+
+    private const string STYLE_CSS = """
+        .color-button radio {
+            background-color: %s;
+            color: %s;
+            padding: 10px;
+            -gtk-icon-shadow: none;
+        }
+    """;
+
+    private BindingGroup terminal_binding;
+    private Gtk.Box theme_buttons;
+
+    public SettingsPopover () {
+        Object ();
+    }
+
+    construct {
+        var zoom_out_button = new Gtk.Button.from_icon_name ("zoom-out-symbolic", MENU) {
+            action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_ZOOM_OUT_FONT,
+            tooltip_markup = Granite.markup_accel_tooltip (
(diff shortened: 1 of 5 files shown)


jeremypw:
@Marukesu This seems like a good idea especially as `MainWindow.vala` is larger than I would like.  Could you fix the conflicts please?

Marukesu:
@jeremypw, sorry for the delay in rebasing this, should be good to review.

don't know what caused the CI error after the rebase, but with the `--print-errorlogs` option, it should be more easy to know what test case failed in the future.

src/Widgets/SettingsPopover.vala
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    private Gtk.RadioButton add_theme_button (string theme, out Gtk.CssProvider css_provider = null) {
+        var button = new Gtk.RadioButton (null) {
+            action_name = ACTION_THEME,
+            action_target = new Variant.string (theme),
+            halign = Gtk.Align.CENTER
+        };
+
+        if (css_provider == null) {
```
jeremypw, line 180: This gives two new compile warnings with valac 0.56.7 
```
../src/Widgets/SettingsPopover.vala:180.13-180.32: warning: Use of possibly unassigned parameter `css_provider'
  180 |         if (css_provider == null) {
      |             ^~~~~~~~~~~~~~~~~~~~   
../src/Widgets/SettingsPopover.vala:186.9-186.90: warning: Use of possibly unassigned parameter `css_provider'
  186 |         style_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);

```
Do you know if later versions of Vala will support it? I tendto avoid this pattern though it is useful.

src/Widgets/SettingsPopover.vala
```diff
@@ -0,0 +1,204 @@
...
+            css_provider = new Gtk.CssProvider ();
+        }
+
+        unowned var style_context = button.get_style_context ();
+        style_context.add_class (Granite.STYLE_CLASS_COLOR_BUTTON);
+        style_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        update_theme_provider (css_provider, theme);
```
jeremypw, line 187: You are potentially providing a `null` or uninitialised parameter to these functions, contrary to their signature.

jeremypw:
Despite the coding technicalities there do not seem to be any regressions in operation.

jeremypw:
@Marukesu Another regression with this is that the font scale is not set individually on each tab.

Marukesu:
I'm tempted to revert the ActionGroup change for now, the widget won't be as self-contained as i would like, but would still make this a good start in reducing the MainWindow scope.

jeremypw:
Yes two smaller PRs would be easier to review.

jeremypw (approved):
Code looks good, no regressions or new terminal warnings seen.

PR #700, opened 2023-02-04
Add shortcuts Ctrl + PageUp and Ctrl + PageDown to switch between tabs

Faelian:
_Control + PageUp_ and _Control + PageDown_ are common shortcuts to switch between tabs.
There are used by other apps such as Web, Chrome and Firefox, and other terminals like gnome-terminal.

This commit add these shortcuts to io.elementary.terminal.


jeremypw (approved):
Works as expected.  Shouldn't be any problem merging this as other common apps already use this shortcut for a similar purpose.

PR #701, opened 2023-02-09
Update Meson Files

Marukesu:
Declare meson targets in the files in they respective folders and make the main meson file handle only dependencies, version checking, and compiler arguments. Also set meson minimal version to 0.58 and stop using deprecated functions.


meson.build
```diff
@@ -1,27 +1,42 @@
-project(
-    'io.elementary.terminal',
-    'vala', 'c',
-    version: '6.1.1'
-)
+project('io.elementary.terminal', 'vala', 'c', version: '6.1.1', meson_version: '>=0.58')
```
danirabbit, line 1: I think this is less legible than having arguments separated into their own lines, personally

data/meson.build
```diff
@@ -1,52 +1,47 @@
...
 )
 
-test (
-    'Validate desktop file',
-    find_program('desktop-file-validate'),
-    args: join_paths(meson.current_build_dir (), meson.project_name() + '.desktop')
-)
+install_data(meson.project_name() + '.gschema.xml', install_dir: datadir / 'glib-2.0' / 'schemas')
+install_data('pantheon_terminal_process_completion_notifications.fish', install_dir: datadir / 'fish' / 'vendor_conf.d')
+install_data('enable-zsh-completion-notifications', install_dir: datadir / meson.project_name())
```
danirabbit, line 32: Can we keep these expanded on several lines as they were? I think this is a lot harder to read at a glance

meson.build
```diff
@@ -1,27 +1,42 @@
...
-        '-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()),
-        '-DHANDY_USE_UNSTABLE_API'
-    ],
-    language:'c'
-)
+glib_version = '>=2.40'
+gtk_version = '>-3.24'
+granite_version = '>=6.1'
+handy_version = '>=0.83'
+vte_version = '>=0.59'
```
danirabbit, line 17: Is there a good reason to have these all separated out like this? it seems like it kind of just makes it more complicated and I'm not sure we ever want to re-use these values besides `glib_version`

po/extra/meson.build
```diff
@@ -1,5 +1 @@
-i18n.gettext('extra',
-    args: '--directory=' + meson.source_root(),
-    preset: 'glib',
-    install: false,
-)
+i18n.gettext('extra', args: '--directory=' + meson.project_source_root(), preset: 'glib', install: false)
```
danirabbit, line 1: Same as before, I think this makes things a bit harder to read at a glance. I'm not sure there's a benefit to compacting it like this

danirabbit (commented):
This mostly looks good, just a few inline comments. I think generally some of these lines are a bit harder to read compared to how they were before. Having arguments separated onto different lines for some of these methods that take a bunch of arguments makes it just easier to read at a glance what's happening

meson.build
Marukesu: i would say it make more easy to find minimal versions of dependencies, but i can inline the not re-used ones if preferred.

danirabbit (approved):
Looks solid. Nice work!

Issue #498, opened 2020-04-03
-t opens two tabs when no windows are open


GammaGames:
Using the command `io.elementary.terminal -t` opens two tabs if non are currently open.

### To Reproduce
1. Add custom keyboard shortcut for `io.elementary.terminal -t` in the settings app
2. Close all terminal windows
3. Press shortcut
4. Notice there are two tabs

### Expected behavior
If there is currently no open terminal, it should just create the single window with the single tab

### Screenshots or screen recordings
![Screen record from 2020-04-03 11 35 47](https://user-images.githubusercontent.com/7832163/78389095-643a1780-759f-11ea-9a73-052d68e0025d.gif)

## Platform Information
![image](https://user-images.githubusercontent.com/7832163/78389509-0d810d80-75a0-11ea-9caa-073eb81193b3.png)

x I'm using the latest released stable version

GammaGames:
Maybe related to #356 too, but different in that right clicking a folder and opening it in terminal opens one tab in your home directory and the second tab in the target directory.

t0suj4:
The same problem appears when open with `--new-window` and `--execute=<command>`

jeremypw:
On elementary 6.0 I am getting *two* extra tabs :disappointed:

jeremypw:
This may be the result of a recent commit #478

jeremypw:
Weirdly, launching e.g. `io.elementary.terminal -w ~/Documents` with a custom shortcut also opens *three* tabs - all at `~` not `~/Documents`.  Launching the same command from e.g. xterm opens a single tab at `~/Documents` as expected.

So there seems something amiss with the way custom shortcuts work with Terminal.

GammaGames:
Is there a way to open a new window on the current desktop, or add a new tab if one already exists? I would love for ctrl+T to be more useful

jeremypw:
In OS7.1 you can open a new window with the "New Window" item in the dock context menu for Terminal.  This opens a new window on the current workspace (even if Terminal is already open on another workspace). There is a keyboard shortcut `<Shift><Control>N` for opening a new window when a Terminal window is focussed but unfortunately it is currently broken (see #814 ) and a fix is awaiting approval (see #816).  You can also drag a tab out of a Terminal window and drop onto the desktop to create a new window containing that tab. Starting Terminal from the commandline pretty much always restored any tabs you had open when Terminal was closed unless the "Privacy/History" setting is off. 

[...]

PR #702, opened 2023-02-12
Rework Application Class

Marukesu:
Modernize the Application class to better reflect the GLib.Application API, it now have a distinction between local and primary, command line and DBus invocations. 

The command line logic was split and simplified:
   - the working directory of the local invocation is used instead of the primary when creating a new tab/window.
  - `--working-directory` is a command line only option, with DBus invocations expecting the `cwd` option in the `platform-data` to be used instead.
  - `--commandline` works without `=`, matching the behaviour of others long options.
  - `ApplicationFlags.CAN_OVERRIDE_APP_ID` was added to facilitate testing changes without needing to close a already open instance of terminal.

Test are added, to assure that we don't cause any regression in the command line/options parsing when modifying the class in the future.


jeremypw:
@Marukesu I do not pretend to understand every aspect of this PR but it looks good and the test framework is impressive. However, when I install and run it from a different terminal app (I've tried XTerm and Blackbox) using no parameters, I get an immediate crash. The backtrace is :

```
Thread 1 "io.elementary.t" received signal SIGSEGV, Segmentation fault.
0x0000555555571c17 in __lambda5_ (self=0x5555555e21c0, id=0x555555eb8e00 "0", process=0x5555555f4620 "gdb io.elementary.terminal", exit_status=0) at ../src/Application.vala:120
120	                    if (terminal.terminal_id == id) {
(gdb) bt
#0  0x0000555555571c17 in __lambda5_
    (self=0x5555555e21c0, id=0x555555eb8e00 "0", process=0x5555555f4620 "gdb io.elementary.terminal", exit_status=0) at ../src/Application.vala:120
#1  0x0000555555572122 in ___lambda5__terminal_dbus_finished_processPython Exception <class 'ValueError'>: Variable 'static_fundamental_type_nodes' not found.

    (_sender=, terminal_id=0x555555eb8e00 "0", process=0x5555555f4620 "gdb io.elementary.terminal", exit_status=0, self=0x5555555e21c0) at ../src/Application.vala:111
#2  0x0000555555573d2e in g_cclosure_user_marshal_VOID__STRING_STRING_INT
    (closure=0x555555601110, return_value=0x0, n_param_values=4, param_values=0x7fffffffd5d0, invocation_hint=0x7fffffffd550, marshal_data=0x0) at ../src/DBus.vala:22
#3  0x00007ffff7e17d2f in g_closure_invoke () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#4  0x00007ffff7e33c36 in  () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x00007ffff7e35614 in g_signal_emit_valist () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff7e35863 in g_signal_emit () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x0000555555573c46 in terminal_dbus_process_finishedPython Exception <class 'ValueError'>: Variable 'static_fundamental_type_nodes' not found.

    (self=, terminal_id=0x555555ebc3f0 "0", process=0x555555dbbd50 "gdb io.elementary.terminal", exit_--Type <RET> for more, q to quit, c to continue without paging--
status=0, error=0x7fffffffd920) at ../src/DBus.vala:30
#8  0x0000555555574805 in _dbus_terminal_dbus_process_finishedPython Exception <class 'ValueError'>: Variable 'static_fundamental_type_nodes' not found.

    (self=, _parameters_=0x7fffdc010a60, invocation=0x555555ec4ec0)
    at /home/jeremyw/Documents/Elementary/terminal/build/DBus.c:537
#9  0x0000555555574997 in terminal_dbus_dbus_interface_method_call
    (connection=0x7fffe400b020, sender=0x7fffdc011bd0 ":1.136", object_path=0x7fffdc011f60 "/io/elementary/terminal", interface_name=0x7fffdc011f10 "io.elementary.terminal", method_name=0x7fffdc006790 "ProcessFinished", parameters=0x7fffdc010a60, invocation=0x555555ec4ec0, user_data=0x5555555fdc80)
    at /home/jeremyw/Documents/Elementary/terminal/build/DBus.c:570
#10 0x00007ffff7d34bfe in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#11 0x00007ffff7eb8c44 in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#12 0x00007ffff7f0d6c8 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#13 0x00007ffff7eb63e3 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff7d0bfb5 in g_application_run () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#15 0x0000555555573aab in _vala_main (args=0x7fffffffde28, args_length1=1)
--Type <RET> for more, q to quit, c to continue without paging--
    at ../src/Application.vala:227
#16 0x0000555555573af8 in main (argc=1, argv=0x7fffffffde28) at ../src/Application.vala:226
(gdb) 
```

jeremypw:
Putting a null check in `\src\Application.vala` line 120 stops the crash at least.

jeremypw:
Other observations so far when launching from `gnome-terminal`:

* Using the `-h` works as expected and does not launch `io.elementary.terminal`.
* Using the `-v` option in a different terminal app opens `io.elementary.terminal` as well as returning the version .
* Using `-n` and `-t` works as master. See issue #498 which persists
*  Using `-n` and `-t` (when an instance of `io.elementary.terminal` already running) works as master 
* Using `-w` works as master. Also restores saves tabs
* Using `-w` in combination with `-n` or `-t` works as master.

jeremypw:
* Using`-e` flag works the same as in master, additional arguments ignored unless preceded by further flags
* Using `--execute=` flag works if there is no space after `=`. If there is a space then the app opens but no tab is created and no error message given.  This differs from master, which gives a warning about spaces both in the launching terminal and in an additional tab in the launched terminal.

jeremypw:
As far as tested so far the `-x` flag works the same as master.  Further testing will be done later.

Marukesu:
@jeremypw the last commits should fix opening the terminal without arguments and the -v argument not returning. i would like to add test cases for them too, but i couldn't find a way to do that right now.

> Using `-n` and `-t` works as master. See issue #498 which persists

about #498, it's out of scope for this PR. The issue is that the restoration code will create a tab in the home directory if no tab is to be restored. I'm working in cleanup this code and fix it in another branch.

> Using `--execute=` flag works if there is no space after `=`. If there is a space then the app opens but no tab is created and no error message given. This differs from master, which gives a warning about spaces both in the launching terminal and in an additional tab in the launched terminal.

the new code ignores `-e` and `-x` if they are a empty line arguments. i don't really think crating a new tab with a warning is the best way to warn the user about it, we could either print a warning and return while in the local handling, or show a error dialog/notification if empty lines was sent to the primary instance.

jeremypw:
@Marukesu 
> about https://github.com/elementary/terminal/issues/498, it's out of scope for this PR. The issue is that the restoration code will create a tab in the home directory if no tab is to be restored. I'm working in cleanup this code and fix it in another branch.

That's fine, I wasn't expecting it to be fixed in this PR.

> the new code ignores -e and -x if they are a empty line arguments. i don't really think crating a new tab with a warning is the best way to warn the user about it, we could either print a warning and return while in the local handling, or show a error dialog/notification if empty lines was sent to the primary instance.

I agree that creating a new tab is not necessary. Either a printed warning or a notification depending on means of  calling would be good.

jeremypw:
While `io.elementary.terminal -w ~/Music` works, `io.elementary.terminal --working-directory=~/Music` does not.  However the same happens in master.

jeremypw:
There is a slight difference between master and PR regarding the `-x` flag.  Running `io.elementary.terminal -x ls` from a directory other the home directory results in a new tab containing a listing of the current directory with this PR but with master it results in a listing of the home directory.

jeremypw (approved):

PR #710, opened 2023-03-09
Application: Implement `startup()`

Marukesu:
Handy initialization and CSS loading is done here now, the `new-window` and `quit` action are implemented here too since there's no reason for them to be part of the window class. one behaviour diference is that the quit action now closes all windows instead of only the current window matching others elementary applications.


jeremypw:
@Marukesu #709 is merged - a conflict needs to be resolved.

jeremypw (changes requested):
There seems to be a regression in that quitting with the quit action now does not save tabs. Reopening just restores the home tab.

jeremypw (changes requested):
No regressions found, but the CI is currently failing on tests

Marukesu:
@jeremypw testing locally, the CI was falling because of the Idle in `on_switch_page()` in the MainWindow class. #720 should fix it.

jeremypw:
For me it was failing in the "quit" action test.  I tried to fix it but failed.  We could always leave that test out and fix it later if necesssary.

jeremypw:
#720 now merged

jeremypw:
Hmm, contrary to my previous diagnosis the CI is (now) failing on the "new-window" action test. See #723

Marukesu:
> Hmm, contrary to my previous diagnosis the CI is (now) failing on the "new-window" action test. See #723

the new-window test is fine, `GLib.Test.skip()` don't return by itself, so the assertion in the quit test got triggered.

jeremypw (approved):

PR #720, opened 2023-04-05
Remove Unecessary Idle During Page Switch

Marukesu:


jeremypw:
This idle loop was added for a reason here: https://github.com/elementary/terminal/pull/204#issuecomment-359936797.  However I cannot find any difference with master now with the idle loop removed, with the cursor or otherwise so I guess the underlying cause was fixed elsewhere.

jeremypw (approved):
No regression found.

PR #721, opened 2023-04-06
Jpw/gapplication test

jeremypw:
Just for test CI on Github - not for merging


jeremypw:
Passes on stable but fails on unstable and development builds.

jeremypw:
Need to compare in to maru/gapplication-startup

PR #722, opened 2023-04-06
Jpw/gapplication test

jeremypw:


jeremypw:
This passes locally.  Looks like a race between the windows being destroyed in the quit action test and the assertion being made.  Sometimes CI passes in one or other builds (not always the same one)  but not in all three.

jeremypw:
Window still undestroyed when assertion made (passes locally).

PR #724, opened 2023-04-07
Application: Handle Minimized in The Primary Instance

Marukesu:
Stop storing the flag value locally and only check for it in the primary instance. This fix a issue with subsequent command line invocations inheriting the minimized flag of the first launch.


jeremypw (approved):
Fixes the described issue

PR #731, opened 2023-05-31
Install man page with meson

jeremypw:
Needs consideration of what to do about deb-packaging and old compressed man file that will be present and take precedence on systems where Terminal has already been installed.

### DIFF
--- a/data/io.elementary.terminal.1
+++ b/data/io.elementary.terminal.1
@@ -1,0 +1,41 @@
+.TH io.elementary.terminal 1 "May 29, 2014"
+.SH NAME
+io.elementary.terminal \- Modern terminal emulator from elementary project
+.PP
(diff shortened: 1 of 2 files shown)


eli-schwartz:
The deb packaging, per the Debian maintainers guide, expects packages to run an installation process that installs a man page without compression, and then dpkg itself handles compressing it.

This is also how other distros work too -- manpages are installed without compression by the build tool, because after `meson install` or `./configure && make && make install`, the packaging tool collects all the files meson has installed, handles compression of manpages among other post-processing steps, and turns it into an archive.

jeremypw:
@eli-schwartz Thanks for the information.  So, we should also remove man page related files from the `debian` folder in `deb-packaging` branch?

eli-schwartz:
Yes, it's a straight-up duplicate.

PR #737, opened 2023-08-21
Fix extra tab

### LINKED ISSUE #736: When creating a new terminal window with the setting the working directory, there is an additional tab that starts at the working directrory
### What Happened?

When I run this command for example:
```sh
io.elementary.terminal -n -w my_new_directory
```
A new terminal window opens with 2 tabs:
1. Current working directory is the `$HOME` directory
2. Current working directory is `my_new_directory`.

Only tab 2 should be open.

### Steps to Reproduce

1. Open the terminal app
2. Enter this command in the terminal app: `io.elementary.terminal -n -w ~/Downloads`
3. A new terminal window will appear with a tab where the current working directory is the home directory and a second tab where the current working directory is the "Downloads" folder.-

### Expected Behavior

A new terminal window should open with just one tab where the current directory is what I specified in the command I use to open a new terminal window.

### OS Version

7.x (Horus)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #736 

* MainWindow: Do not add default tab - responsibility of Application now
* MainWindow: Save tabs from each window  (settings permitting) - the last window closed will overwrite previous windows tabs
* Application:  Leave it to MainWindow to restore tabs or not.  The first window will always restore tabs if settings permit
* Application: Distinguish when a working directory was explicitly requested (`-w` option) when an instance is already running in order to allow different behaviour for `-t` option   It is an open question whether the `-t` option should behave the same as the new-tab action  (honor "follow-last-tab" setting).  It also fixes a small issue in master where if Terminal is launched with `io.elementary.terminal` when an instance is already added then a new unrequested tab is added to the first window.

To clarify what is going on, MainWindow does not automatically add a default tab. 

 ### TESTING

Where <folder-path> is the path to a folder that is *not* amongst the saved tabs.

 #### With History/restore-tabs ON:

Open first instance of Terminal from command line with `io.elementary.terminal -nw <folder-path>`
Terminal opens containing one tab at the requested path plus restored tabs
Close Terminal and reopen with `io.elementary.terminal`
Terminal opens with restored tabs including the requested path

Now open second instance of Terminal from the command line with `io.elementary.terminal -nw <folder-path>`
A new window opens containing only the requested tab (difference from master - fixes issue)

Close the first window then close the second window
Reopen Terminal
Terminal opens with only the tab from the second (last closed) window

Open second instance with `io.elementary.terminal -w <folder-path>
The first window is focused with the tab selected (already exists)

Open second instance with `io.elementary.terminal -w <different-folder-path>
The first window is focused with an additional tab at the new path - the new tab is selected

Open second instance with `io.elementary.terminal -t`
The first window is focused with an additional tab at the current working directory.

Open second instance with `io.elementary.terminal -nt`
A new window is opened with a single tab at the current working directory.

Open a second instance `io.elementary.terminal`
The first window is focused and *no* extra tab added (difference from `master`)

 #### With History/restore-tabs OFF

Open first instance of Terminal from command line with `io.elementary.terminal -nw <folder-path>`
Terminal opens containing one tab at the requested path.
Close Terminal and reopen with `io.elementary.terminal`
Terminal opens with the default tab

Now open second instance of Terminal from the command line with `io.elementary.terminal -nw <folder-path>`
A new window opens containing only the requested tab (difference from master - fixes issue)

Close the first window then close the second window
Reopen Terminal
Terminal opens with the default tab

 #### Running Commands

The `-e` and `-x` options behave similarly.  If they create the first window then tabs are restored (if settings permit) and then an additional tab at the current working directory is created for each command to run in.  If used with the `-n` option to create an additional window then only tabs running the commands are created.


jeremypw:
Probably needs splitting into two PRs.

jeremypw:
Now only fixes the linked issue.  A separate PR dealing with option handling will be prepared later.

src/Application.vala
```diff
@@ -201,7 +201,11 @@ public class Terminal.Application : Gtk.Application {
         );
 
         var new_window_action = new SimpleAction ("new-window", null);
-        new_window_action.activate.connect (new_window);
+        new_window_action.activate.connect (() => {
+            var win = new MainWindow (this, false); // Do not restore tabs
```
Marukesu, line 205: this will break having the first window be created by the action when we enable DBus activation (the only thing that was stopping us of doing that was the duplicate tab). so the `active_window == null` is necessary here.

src/Application.vala
```diff
@@ -201,7 +201,11 @@ public class Terminal.Application : Gtk.Application {
...
+            win.add_tab_with_working_directory (null);  // Add default tab
+            win.present ();
+        });
```
Marukesu, line 208: please, keep the separate callback, it makes easier to breakpoint it during debug after.

src/Application.vala
```diff
@@ -216,15 +220,15 @@ public class Terminal.Application : Gtk.Application {
     protected override int command_line (ApplicationCommandLine command_line) {
         unowned var options = command_line.get_options_dict ();
         var window = (MainWindow) active_window;
-        bool new_window;
+        bool new_window = false;
```
Marukesu, line 223: no need for the default value here, options.lookup() will always set a value.

src/Application.vala
```diff
@@ -216,15 +220,15 @@ public class Terminal.Application : Gtk.Application {
...
+        options.lookup ("new-window", "b", out new_window);
+        /* Uncertain whether tabs should be restored when app is launched with working directory from commandline.
+         * Currently they are set to restore (subject to the restore-tabs setting).
+         * If it is desired that tabs should never be restored in these circimstances add another check below.
+         */
```
Marukesu, line 228: this commentary doesn't make sense any more. Before, launching with one of `-x`, `-e`, or with a command didn't restored tabs. But now the behaviour is to always restore if `-n` is not used too.

src/Application.vala
```diff
@@ -216,15 +220,15 @@ public class Terminal.Application : Gtk.Application {
...
 
-        if (window == null || options.lookup ("new-window", "b", out new_window) && new_window) {
-            /* Uncertain whether tabs should be restored when app is launched with working directory from commandline.
-             * Currently they are set to restore (subject to the restore-tabs setting).
-             * If it is desired that tabs should never be restored in these circimstances add another check below.
-             */
-            bool restore_tabs = !("commandline" in options || "execute" in options) || window == null;
-            window = new MainWindow (this, restore_tabs);
+        if (window == null || new_window) {
```
Marukesu, line 230: keep the options.lookup() inline here, it's not only getting the value of `new-window` option, but validating that it's a boolean value too.

src/Application.vala
```diff
@@ -216,15 +220,15 @@ public class Terminal.Application : Gtk.Application {
...
+            window = new MainWindow (this, !new_window);
```
Marukesu, line 231: if this is the first window, we aren't restoring. is that expected? i don't see a why `-n` won't restore the last session if it's the first window.

src/Application.vala
```diff
@@ -246,6 +250,11 @@ public class Terminal.Application : Gtk.Application {
             window.add_tab_with_working_directory (working_directory, null, new_tab);
         }
 
+        // Fallback to a default tab if no tabs were opened 
+        if (window.terminals.length () == 0) {
+            window.add_tab_with_working_directory (Environment.get_home_dir ());
+        }
```
Marukesu, line 256: this cannot even happen, because we are always creating a tab above.

src/MainWindow.vala
```diff
@@ -1338,8 +1337,8 @@ namespace Terminal {
             string[] zooms = {};
 
             Application.saved_state.set_double ("zoom", current_terminal.font_scale);
-
-            if (Granite.Services.System.history_is_enabled () &&
+            if (tabs_were_restored &&  // Do not save tabs if window opened without restoring to avoid overwriting
+                Granite.Services.System.history_is_enabled () &&
```
Marukesu, line 1341: I believe the agreement in the files PR was that we always save the last closed window state. Not the first opened one.

src/tests/Application.vala
```diff
@@ -165,18 +165,22 @@ namespace Terminal.Test.Application {
             option ("{'new-window':<false>}", "@a{sv} {}", () => {
                 var n_windows = (int) application.get_windows ().length ();
                 assert_cmpint (n_windows, CompareOperator.EQ, 1);
+                unowned var window = (MainWindow) application.active_window;
+                assert_nonnull (window);
+                var n_tabs = (int) window.terminals.length ();
+                assert_cmpint (n_tabs, CompareOperator.EQ, 1);
```
Marukesu, line 171: this is in the wrong callback no? here we aren't creating a new window, so you are actually checking the restored tabs.

src/Application.vala
jeremypw: I think is is cleaner and more predictable if `-n` always creates a fresh new window.  If the user wants restored tabs then omit the `-n`.

src/Application.vala
jeremypw: It can happen if the `execute` option is used without any command.

src/tests/Application.vala
jeremypw: Yes, I was checking something - not needed now.

src/Application.vala
jeremypw: Done

src/Application.vala
jeremypw: That will not work.  We need a value for `new_window` when `window` is `null`.  If `window` is `null` and the lookup is inline then `new_window` will be undefined because of lazy evaluation of OR expression.

Marukesu: you can use `window != null` in the restore_tabs argument, that has the same meaning of `new_window` being true.

src/Application.vala
Marukesu: keep in mind that `-n` will also mean a hard "drop tabs" when used in the first launch, because there's no way to not restore the tabs in any other case that isn't the first launch.

if you fine with that behaviour, i won't oppose the change.

jeremypw: On reflection I have left restoring tabs to MainWindow and not dependent on options as it is simpler.

src/Application.vala
jeremypw: Not sure what this means :disappointed: as I am not sure how DBus activation works.  Could you give more details or suggested code?

Marukesu:
> Application: Distinguish when a working directory was explicitly requested (-w option) when an instance is already running in order to allow different behaviour for -t option It is an open question whether the -t option should behave the same as the new-tab action (honor "follow-last-tab" setting).

The idea with the rewrite is that working directory isn't a "real" option, so it shouldn't cause different behaviour by itself.

> It also fixes a small issue in master where if Terminal is launched with io.elementary.terminal when an instance is already added then a new unrequested tab is added to the first window.

the behaviour of opening a new tab when invoked in a directory without a tab already open is also intentional. the rationale is that any command line invocation is done with the intention of open a tab or window, so it act as a "weakened" (it won't create a tab, if one in the current directory already exists) version of `-t`.

if we are going to change the behaviour here, i would prefer that it's to drop the weakened `-t` for assuming a real one.

src/Application.vala
```diff
@@ -217,17 +217,18 @@ public class Terminal.Application : Gtk.Application {
...
+            window = new MainWindow (this);
         }
 
-        unowned var working_directory = command_line.get_cwd ();
+        unowned string working_directory;
+        var wd_was_requested = options.lookup ("working-directory", "^&ay", out working_directory);
+        working_directory = command_line.get_cwd ();
+
+        assert (working_directory != "\0");  // Can we be sure of this? If not, need fallback
+
```
Marukesu, line 231: as explained, working-directory isn't meant to be a real option, also assert here isn't right, as it will force quit the primary instance.

src/Application.vala
```diff
@@ -237,15 +238,20 @@ public class Terminal.Application : Gtk.Application {
...
+            window.add_tab_with_working_directory (working_directory, command);
+        } else if (wd_was_requested) {
+            window.add_tab_with_working_directory (working_directory, null);
+        } else if (new_tab) {
+            window.add_tab_with_working_directory (working_directory, null); //FIXME Should we honor "follow-last-tab" setting?
         }
 
+        // Only if no other tabs, add default tab
+        add_default_tab (window, false);
+
```
Marukesu, line 254: this won't respect the current working directory as the others.

src/MainWindow.vala
```diff
@@ -260,28 +258,20 @@ namespace Terminal {
 
             restorable_terminals = new HashTable<string, TerminalWidget> (str_hash, str_equal);
 
-            if (recreate_tabs) {
-                open_tabs ();
+            if (is_first_window) {
```
Marukesu, line 261:
```suggestion
            if (app.get_windows ().length () == 1) {
```

then you can drop both the `window_number` and `is_first_window` properties.

src/MainWindow.vala
```diff
@@ -260,28 +258,20 @@ namespace Terminal {
...
+                open_saved_tabs ();
             }
+
+            // This no longer ensures at least one tab is opened.
+            // The creator of the window (Application.vala or on_tab_moved) must add a tab if necessary
```
Marukesu, line 266: Unnecessary comment.

src/MainWindow.vala
```diff
@@ -1177,9 +1161,13 @@ namespace Terminal {
             // Closing a tab will switch to another, which will trigger check for same names
         }
 
-        private void action_new_tab () {
+        public void action_new_tab () {
```
Marukesu, line 1164: This doesn't need to be public.

src/MainWindow.vala
```diff
@@ -1177,9 +1161,13 @@ namespace Terminal {
...
             if (Application.settings.get_boolean ("follow-last-tab")) {
-                new_tab (current_terminal.get_shell_location ());
+                if (current_terminal != null) {
+                    new_tab (current_terminal.get_shell_location ());
+                } else {
+                    new_tab (Environment.get_current_dir ());
+                }
```
Marukesu, line 1170: it's better to keep any "current directory" concept exclusive to the application class. because of the local/primary instance logic.

src/tests/Application.vala
```diff
@@ -145,7 +145,7 @@ namespace Terminal.Test.Application {
                 unowned var window = (MainWindow) application.active_window;
                 assert_nonnull (window);
                 var n_tabs = (int) window.terminals.length ();
-                assert_cmpint (n_tabs, CompareOperator.EQ, 2);
+                assert_cmpint (n_tabs, CompareOperator.EQ, 1); // No default tab added as well in this case
```
Marukesu, line 148: this test was made to assert that we always create a new tab when requested, so this change is wrong.

src/tests/Application.vala
```diff
@@ -169,14 +169,14 @@ namespace Terminal.Test.Application {
...
-            string[] execute = { "true", "echo test", "echo -e te\\tst", "false" };
+            string[] execute = { "true", "echo test", "echo -e te\\tst", "false" }; // 4 commands
 
             // valid
             option ("{'execute':<[b'%s']>}".printf (string.joinv ("',b'", execute)), "@a{sv} {}", () => {
                 unowned var window = (MainWindow) application.active_window;
                 assert_nonnull (window);
                 var n_tabs = (int) window.terminals.length ();
-                assert_cmpint (n_tabs, CompareOperator.EQ, 5); // include the guaranted extra tab
+                assert_cmpint (n_tabs, CompareOperator.EQ, 5); // Includes initial default tab added when no tabs restored.
```
Marukesu, line 179: now that we aren't creating the extra tab, this should be 4.

src/Application.vala
jeremypw: I'll remove the assert before merge - that is for testing. So far I havent been able to trigger it so probably do not need a fallback?

src/Application.vala
jeremypw: Whether or not it respects the current working directory depends on the "follow-last-tab" setting (that is what the *default* tab does) since it uses the action-new-tab. This is for consistency with other situations where there is no specified tab.

src/MainWindow.vala
jeremypw: That does not work.  The number of windows in the app may change after construction of the window.

src/MainWindow.vala
jeremypw: Debatable but I'll remove it.

src/MainWindow.vala
jeremypw: Oh yes, I was going to call it directly but opted for activate_action instead. Will revert.

src/MainWindow.vala
jeremypw: So what do you suggest?

src/tests/Application.vala
jeremypw: It is necessary for the test to pass.  If the original test is correct then the app is wrong

src/tests/Application.vala
jeremypw: Hmm then it should be failing - will investigate.

jeremypw:
> the behaviour of opening a new tab when invoked in a directory without a tab already open is also intentional. the rationale is that any command line invocation is done with the intention of open a tab or window

Again, I do not understand this. The change (and the whole PR) is about not opening *two* tabs when the user would expect only one.  The change from `master` being discussed here is about when Terminal is invoked without *any* options so the user is not requesting a new tab.  Usually it is because the user has forgotten Terminal is already running so it should behave as if the user had just clicked on the dock icon again - i.e. it should just focus the existing window.

jeremypw:
There is a case for reconsidering whether we need both `-w` and `-t` options or whether they should do the same or different things, bearing in mind we do not want to break existing scripts if we can help it. But I am not sure this PR is the place - I am just trying to make sure Terminal continues to follow the information that the `-h` option gives while fixing the issue of unexpected extra tabs.

jeremypw:
Elementary Terminal is currently different from other terminals in that it has a "follow-last-tab" setting (unexposed) that is by default `false` so that additional tabs are by default opened at the user's home directory *not* the current directory. This complicates matters considerably. Might be worth discussing whether that setting should continue to be supported or whether we should just emulate other major terminals. But that is for another PR.

jeremypw:
> so it act as a "weakened" (it won't create a tab, if one in the current directory already exists) version of -t.

This would make commandline invocation different from invocation from the dock or App Menu which both just focus an existing primary window, so I disagree.

Interestingly `gnome-terminal` *always* opens a new window when a primary window exists  even when invoked with no options. `xterm` does the same.  Following that would probably simplify things but would need design team sign off.

src/tests/Application.vala
jeremypw: So what is happening is that the test environment opens a window without restoring tabs *before* processing the option.  So there is already a default tab in the window and the execute option creates 4 more, one for each command, giving a total of 5.

If you execute multiple commands before any window is opened e.g. with a commandline `io.elementary.terminal -e "true" -e "echo Hi"` then no extra default tab appears.

src/tests/Application.vala
jeremypw: If you run `io.elementary.terminal -t` manually without a Terminal window open you get one tab (as expected). If you run the same command with the primary window open (but no restored tabs) then you get 2 tabs - the original default one and the requested additional one.

I am not exactly sure what happens in the test environment - does a window get created before the options are processed or not?

jeremypw: After some investigation it seems like before processing the commandline in `option ()` there is 1 window with 1 tab. So I would have expected `-t` true to create a new tab like when used manually.  But it doesn't in the test environment for some reason.

jeremypw: OK, so it is race - the callback is executed before the window terminal list is updated in the `on_tab_added` signal handler.  Need to get a more immediate count of true number of tabs in the notebook.  I'll review whether the terminals list is needed or not.

jeremypw:
@Marukesu I'm going to have to ask you to look at that race issue in the testing of adding tabs I am afraid.  I tried a few things but I do not fully understand the code :disappointed:  Even getting the `n_tabs` directly from `notebook` does not work.  The `callback` is running before `MainWindow` even starts to add the tab.  Looks like `iterate_context` is not blocking for long enough for MainWindow to complete everything needed to add a tab.

jeremypw:
Not sure why we're getting the expected tab count in the `--execute` test but not the `new-tab` option test though :shrug:

src/MainWindow.vala
jeremypw: Note: Environment.get_current_dir () is already in use in TerminalWidget

src/tests/Application.vala
jeremypw: Further investigation indicates that *6* tabs are added during this test but the callback occurs before the last one is added. So the same race between the callback and adding a tab that was noted in the `new-tab` test is happening here as well

jeremypw:
So it seems that for each call to `option ()` in the Test environment, the application command_line function gets called *twice*?  Is this intended?  It is not what happens when launching the application on the desktop.  It results in unexpected  numbers of tabs appearing (at least, different from real life).

Marukesu:
yes it does, that is the only way to test the options and actions without having interference of the command line. the first invocation override the default handler, so for the tests, the second invocation appears as if it was the first one.

src/tests/Application.vala
Marukesu: simillary as the new-tab test, there's 1 tab from first launch, and 4 from the commands, the sixty one seems strange.

src/MainWindow.vala
Marukesu: > So what do you suggest?

Only the Application class known the real working directory for the current invocation, so either action-new-tab is updated to take a path, or we don't try  to use the action in the command line context, where is the only place that a concept of working directory exists.

> Note: Environment.get_current_dir () is already in use in TerminalWidget

It actually isn't, the only place that we call activate_shell() without a location is in the case that we call new_tab() with a empty location, what isn't possible.

src/Application.vala
Marukesu: this is already being inconsistent with the others command line options handling, where the current working directory is always respected. the code in the action-new-tab won't respect it either, it will either use the last tab directory or the current working directory of the first invocation.

jeremypw:
I have found a fix for the tests by allowing the test to choose whether or not `command_line` is called with a `null` parameter before the test commandline.  This defaults to "yes" but for the "new-tab" tests it needs to be "no" in order to get the "expected" result (i.e. the same as real life).  I havent fully got to the bottom of it but there does seem to be a timing issue when adding a second tab to the existing window in the new tab test.

jeremypw:
I'll have another look tomorrow - I've been down enough rabbit holes and followed enough red herrings today :stuck_out_tongue_closed_eyes:

PR #740, opened 2023-10-21
Expose audible-bell setting in Settings Menu

### LINKED ISSUE #596: [odin beta] No terminal bell any more
## What Happened
I ran ping -a in the new odin terminal and there was no sound

## Expected Behavior
The terminal bell should be heard to indicate that the server is up (it sends the bell character).
If the bell is disabled by default, a toggle to turn it back on would be nice

## Steps to Reproduce

1. Open a terminal
2. ping -a a server that is up

## Platform Information

* Elementary OS 6 (updated to 20210611)

jeremypw:
Fixes #596 

A possible refinement of this would be to only show this option when the relevant system settings mean that an audible or visual notification will be made on an event alert.

### DIFF
--- a/src/Widgets/SettingsPopover.vala
+++ b/src/Widgets/SettingsPopover.vala
@@ -113,7 +113,7 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
             active = Application.settings.get_boolean ("natural-copy-paste")
         };
 
-        var audible_bell_button = new Granite.SwitchModelButton (_("Audible Bell")) {
+        var audible_bell_button = new Granite.SwitchModelButton (_("Event Alerts")) {
             description = _("Send an event alert for invalid input or multiple possible completions (subject to system sound settings)"),
             active = Application.settings.get_boolean ("audible-bell")
         };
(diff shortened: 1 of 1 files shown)


src/Widgets/SettingsPopover.vala
```diff
@@ -113,6 +113,11 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
             active = Application.settings.get_boolean ("natural-copy-paste")
         };
 
+        var audible_bell_button = new Granite.SwitchModelButton (_("Audible Bell")) {
+            description = _("Send an event alert for invalid input or multiple possible completions (subject to system sound settings)"),
```
danirabbit, line 117: In the description text in System Settings we say "notify" here. Also use the same format we use in dialogs etc to describe System Settings locations

```suggestion
            description = _("Notify for invalid input or multiple possible completions (subject to System Settings → Sound)"),
```

danirabbit (changes requested):
Looks good just some minor copy adjustments

jeremypw:
Thanks for the improvements :smile:

danirabbit (approved):
Looks good :rocket:

PR #742, opened 2023-10-29
Continous saving of current tabs and tab-zooms

ldrahnik:


jeremypw (changes requested):
This is mostly working - thanks!  A couple of issues:

 - Changing the default font with `<Control>scroll` does not change settings because at the moment the Terminal Widget changes the font directly.  You need to change the scroll handler to activate the relevant window action instead.

 - I suggest you split the `save_opened_terminals` function up so that you can change the zooms and uris settings independently to save unnecessary work being done.

ldrahnik:
>Changing the default font with scroll does not change settings because at the moment the Terminal Widget >changes the font directly. You need to change the scroll handler to activate the relevant window action instead. 

https://github.com/elementary/terminal/pull/742/commits/8b126dd31bf1c43c8461938b48245b3967c38e53 

> I suggest you split the save_opened_terminals function up so that you can change the zooms and uris settings independently to save unnecessary work being done. It might lead to not identical length of array with tabs against tab-zooms. Why? Because now there is not added terminal with a null location or empty location.

https://github.com/elementary/terminal/pull/742/commits/24ce4cc781da0c79e634cc0ff484c14c62d95942

src/MainWindow.vala
```diff
@@ -1344,9 +1344,77 @@ namespace Terminal {
             return;
         }
 
+        private void zoom_changed (TerminalWidget terminal_widget) {
+            save_opened_terminal_zoom (terminal_widget);
+        }
```
jeremypw, line 1349: This function appears redundant? It just calls another function with the same signature.

src/MainWindow.vala
```diff
@@ -1344,9 +1344,77 @@ namespace Terminal {
...
+
+        private void cwd_changed () {
+            check_for_tabs_with_same_name ();
```
jeremypw, line 1352: I don't think this call is necessary - the function is called by the `window_title_changed` signal handler anyway.

src/MainWindow.vala
```diff
@@ -639,7 +630,13 @@ namespace Terminal {
...
 
         private void on_tab_reordered (Granite.Widgets.Tab tab, int new_pos) {
+            var terminal_widget = get_term_widget (tab);
+
             current_terminal.grab_focus ();
+
+            terminals.remove (terminal_widget);
+            terminals.insert (terminal_widget, new_pos);
+
             save_opened_terminals ();
```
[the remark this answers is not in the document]
ldrahnik: https://github.com/elementary/terminal/pull/742/commits/21c27f527972f77f0a53f89a5d79b112899bf467

jeremypw (changes requested):
You seem to have repeated a lot of code in `save_open_terminals_with_zooms`? Can you just call the split functions?

Adding another signal to the terminal widget is not the best solution - you can just activate the appropriate action in the scroll handler instead of calling `increment_size` and `decrement_size`.  This can be done with code like
`window.get_simple_action (MainWindow.ACTION_ZOOM_FONT_IN).activate (null);`  (other examples of this pattern are already used in the code)

I have made a couple of other comments inline.

src/MainWindow.vala
ldrahnik: I am sorry, I do not know. Eventually, I would remove call `check_for_tabs_with_same_name` in master branch as this should be focused only on continuous saving in case this branch would be reverted for example in the future.

src/MainWindow.vala
ldrahnik: https://github.com/elementary/terminal/pull/742/commits/0a0d06c9c9f4263deddfbefa450788166df8331e

ldrahnik:
> Adding another signal to the terminal widget is not the best solution - you can just activate the appropriate action in the scroll handler instead of calling increment_size and decrement_size. This can be done with code like
window.get_simple_action (MainWindow.ACTION_ZOOM_FONT_IN).activate (null); (other examples of this pattern are already used in the code)

https://github.com/elementary/terminal/pull/742/commits/0a0d06c9c9f4263deddfbefa450788166df8331e

ldrahnik:
> You seem to have repeated a lot of code in save_open_terminals_with_zooms? Can you just call the split functions?

https://github.com/elementary/terminal/pull/742/commits/2574910eb4aea85e98ff250682aac87b780b85b5

Maybe would be a better solution to make again one function but with bool value `save_zooms` to avoid duplicated cycles though opened terminals.

I think I processed all notes. Thank you for reviewing.

ldrahnik:
https://github.com/elementary/terminal/pull/742/commits/9a9d24a910e40ec79686da710f225aefa7db04f2 

Handle the situation when is opened first terminal, then second one, second one is closed and then is laptop rebooted. Without this code would be saved tabs and tab-zooms from second terminal but should be from first one.

jeremypw:
@ldrahnik I am fine with you merging the saving functions back into one with an extra parameter.

src/MainWindow.vala
```diff
@@ -250,6 +248,7 @@ namespace Terminal {
                 if (focus_timeout == 0) {
                     focus_timeout = Timeout.add (20, () => {
                         focus_timeout = 0;
+                        save_opened_terminals_with_zooms ();
```
jeremypw, line 251: What is the reason for saving the tabs on `focus-in` event?

ldrahnik: Handle the situation when is opened first terminal, then second one, second one is closed and then is laptop rebooted. Without this code would be saved tabs and tab-zooms from second terminal but should be from first one.

ldrahnik: But it works only when is focus passed back to previously opened terminal, do you know what I mean, it works only when the previously opened terminal is in the layer immediately under the closed one.

jeremypw: Yes, the issue of what to do about saving settings in second and subsequent windows is not really decided. One point of view (including me) is that settings only refer to the first opened window (that is what Files implements). Others think that the last *closed* window is the one that should save the settings (which is what Terminal implements).  I think the issue you are trying to fix is also present in master because rebooting forcibly kills the process and the save on close code never runs.  I would not add this line in this PR since master does not save on focus in and there is no agreed "right" answer.

ldrahnik:
Refactored to one function but with 2 bool parameters because I realized I do not know how to do that with only 1 - 1. state only tabs, 2. state only zooms, 3. state both

jeremypw:
It could be done with an enum as a parameter but probably not worth it in this case.

src/MainWindow.vala
ldrahnik: @jeremypw I am afraid that I discovered weird behavior without added resaving `tabs` and `tab-zooms` during focus in event. Because when I close terminal, then I return to previously opened terminal and I for example change cwd in any tab and then is saved broken state, because on cwd are saved only `tabs` and no `tab-zooms`, in other words saving GLib.Settings calls returns back attack like boomerang :D My added saving to focus in event would solve this problem, and in best way I think or this would be solved by not saving calls and call `tabs` or `tab-zooms` all time so when I return back to previously opened terminal and I start somehow interact with him broken state can not occurs. What do you think?

ldrahnik: @jeremypw I have also an idea about adding a new option continuous saving to remember tabs, so change it from bool to select with variants continuous saving, on close and do not save, on close, and could be even by default on the close variant.

jeremypw: @ldrahnik Yes, on reconsideration I think we have to save on focus-in because the settings must always reflect the state of the current window.  I think the issue of whether continuous saving should be optional should be raised separately as it needs signing off by the design team to add another setting.  In general elementary tries to avoid too many settings.  There would need to be two frequently used situations where different settings are required.  Why would users not want to continuous save?

jeremypw: Yes, I think you would struggle to convince the design team that was a good idea unless the app was really draining the battery with continuous saving - in which case it would be better to throttle it (Code autosaving is throttled so that it only happens after one second of inactivity).

jeremypw:
@ldrahnik One small thing I noticed - `save_opened_terminals ()` get called before `current_terminal` is set leading to the critical warning  `gdouble vte_terminal_get_font_scale)VteTerminal*): assertion VTE_IS_TERMINAL(terminal) failed`  (MainWindow lin 1361)

ldrahnik:
> One small thing I noticed - save_opened_terminals () get called before current_terminal is set leading to the critical warning gdouble vte_terminal_get_font_scale)VteTerminal*): assertion VTE_IS_TERMINAL(terminal) failed (MainWindow lin 1361)

https://github.com/elementary/terminal/pull/742/commits/2e059a23f0201c09caa4a23617d9488c4438e7bc

jeremypw (approved):
I think this is good to go now - thanks for your contribution!

jeremypw:
Need to fix CI workflow (or override) before can be merged.

jeremypw (changes requested):
Sorry, missed the fact that one of the unit tests now fails  (CI was failing anyway in other PRs for an unrelated reason).  The error given is 
```
▶ 2/3 ERROR:../src/tests/Application.vala:155:__lambda53_: assertion failed (n_tabs == 1): (2 == 1) ERROR 
```
Not immediately obvious why this PR should affect the number of tabs - I'll investigate.

jeremypw:
@ldrahnik I have pushed a PR to merge into this one that should fix the unit testing issue for now. The unit tests should probably be rewritten/extended to take continuous saving (or not) into account but that is complicated so can be left for another PR.

ldrahnik:
@jeremypw Merged.

jeremypw (approved):

PR #743, opened 2023-10-31
Fix unwanted extra tab 

### LINKED ISSUE #736: When creating a new terminal window with the setting the working directory, there is an additional tab that starts at the working directrory

jeremypw:
Fixes #736 

 - Application only adds an extra tab only when specifically asked
 - MainWindow adds an extra tab when required to ensure at least one tab


jeremypw:
Fixing CI ...

src/Application.vala
```diff
@@ -110,8 +110,6 @@ public class Terminal.Application : Gtk.Application {
             if (working_directory != "\0") {
                 Environment.set_current_dir (working_directory); // will be sent via platform-data
             }
-
-            options.remove ("working-directory");
```
Marukesu, line 114: instead of keeping "working-directory", add "new-tab" to the options, the working directory is sent via platform data.

src/Application.vala
```diff
@@ -230,32 +231,43 @@ public class Terminal.Application : Gtk.Application {
...
 
-        if (window == null || options.lookup ("new-window", "b", out new_window) && new_window) {
+        if (window == null || (options.lookup ("new-window", "b", out new_window) && new_window)) {
             /* Uncertain whether tabs should be restored when app is launched with working directory from commandline.
              * Currently they are set to restore (subject to the restore-tabs setting).
              * If it is desired that tabs should never be restored in these circimstances add another check below.
              */
-            bool restore_tabs = !("commandline" in options || "execute" in options) || window == null;
-            window = new MainWindow (this, restore_tabs);
+            bool restore_tabs = !("commandline" in options || "execute" in options || "working-directory" in options) || window == null;
```
Marukesu, line 239: i believe we still want to restore tabs if a working directory is set.

also, this change is invalidating the commentary above, it should be removed if this change is to stay.

src/Application.vala
```diff
@@ -230,32 +231,43 @@ public class Terminal.Application : Gtk.Application {
...
+            window = new MainWindow (this, restore_tabs, false);
```
Marukesu, line 240: you can lookup for "new-tab" earlier and then do:
```suggestion
            window = new MainWindow (this, restore_tabs, !new_tab);
```

src/Application.vala
```diff
@@ -230,32 +231,43 @@ public class Terminal.Application : Gtk.Application {
...
         }
 
-        unowned var working_directory = command_line.get_cwd ();
         unowned string[] commands;
         unowned string command;
+        unowned string wd;
```
Marukesu, line 245: keep using `command_line.get_cwd()` it's already set to be the directory passed in "working-directory" or the directory of the current invocation.

src/Application.vala
```diff
@@ -230,32 +231,43 @@ public class Terminal.Application : Gtk.Application {
...
-            window.add_tab_with_working_directory (working_directory, command, new_tab);
-        } else {
-            window.add_tab_with_working_directory (working_directory, null, new_tab);
+            window.add_tab_with_working_directory (
+                wd != null ? wd : command_line.get_cwd (),
+                command,
+                new_tab
+            );
+        } else if (wd_option_present) {
+            window.add_tab_with_working_directory (wd != null ? wd : command_line.get_cwd (), null, new_tab);
```
Marukesu, line 268: you won't need this block if you change "woking-directory" to be "new-tab" in `handle_local_options()`.

src/Application.vala
```diff
@@ -230,32 +231,43 @@ public class Terminal.Application : Gtk.Application {
...
+        } else if (new_tab) {
+            window.add_tab_with_working_directory (Environment.get_current_dir (), null, new_tab);
```
Marukesu, line 270: `Environment.get_current_dir()` here (and amost everywhere in the code) is wrong.

the only places we can be sure it's correct is in `main()`, `new Application()`, `construct`, `local_command_line()`, and `handle_local_options()`.

src/MainWindow.vala
```diff
@@ -214,15 +214,15 @@ namespace Terminal {
...
             if (directory == null || directory == "") {
                 if (notebook.tabs.first () == null || command != null || create_new_tab) { //Ensure at least one tab
                     new_tab ("", command);
                 }
 
                 return;
             } else {
-                location = directory;
+                // Substitute "~" etc but do not add the "file://" scheme
+                location = Utils.sanitize_path (directory, Environment.get_current_dir (), false);
```
Marukesu, line 225: this is wrong, `Environment.get_current_dir()` is the directory of the first invocation (in most of cases, $HOME), not of the current invocation.

for example, assuming terminal was already opened via dock or wingpanel, if the user is in `/home/user/projects/terminal/build` and run `io.elementary.terminal -t --working-directory=../src`, this will try to open `/home/src` not `/home/user/projects/terminal/src`.

ps: `-tw ../src` and `--working-directory ../src` will work as expected because they are handled by the shell, so i don't think there's much value on this change.

src/tests/Application.vala
```diff
@@ -199,7 +199,7 @@ namespace Terminal.Test.Application {
                 unowned var window = (MainWindow) application.active_window;
                 assert_nonnull (window);
                 var terminal_directory = window.current_terminal.get_shell_location ();
-                assert_cmpstr (terminal_directory, CompareOperator.EQ, working_directory);
+                assert_cmpstr (terminal_directory, CompareOperator.EQ, Environment.get_current_dir ());
```
Marukesu, line 202: this is wrong too, this test is ensuring that we respect the working directory of the remote invocation, `Environment.get_current_dir()` isn't it.

Marukesu (changes requested):

src/Application.vala
jeremypw: I have simplified and clarified this - Tabs are always restored for first created window (subject to privacy settings);

src/Application.vala
jeremypw: I have removed the third parameter for window construction now

src/MainWindow.vala
jeremypw: OK i'll revert this for now - it can be addressed later if needed.

jeremypw:
@Marukesu  Thanks for the detailed review.  I have tried to address your comments.  I have reduced the scope of this PR to not include path abbreviations.  This will be addressed in a separate PR.  

If the behaviour is now as expected, please feel free to push a commit rearranging the code as you would like.

jeremypw:
There is still a minor issue with relative paths in that it is not obvious to the user why `io.elementary.terminal -tw ../src` works (in a build directory) but not `io.elementary.terminal -w ../src`  Maybe `-w` by itself should imply `-t`?

src/Application.vala
jeremypw: I now add "new-tab" to the options if "-w" option is present.

jeremypw:
I have now pushed a possible fix for the relative path issue.  After this commit both `io.elementary.terminal -w ../src` and `io.elementary.terminal --working-directory=../src` behave as expected.

Marukesu (approved):
LGTM.

PR #749, opened 2023-11-22
MainWindow: Move Actions and Keyboard Handling Related to The View to TerminalWidget

Marukesu:
As part of reducing the size of the MainWindow class, these was relying on the current_terminal property to work, however, both key handling and actions goes from the focused widget to the topmost parent. so we can move them to TerminalWidget class without problems.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -67,6 +67,26 @@ namespace Terminal {
             }
         }
 
+        public const string ACTION_COPY = "term.copy";
+        public const string ACTION_COPY_OUTPUT = "term.copy-output";
+        public const string ACTION_PASTE = "term.paste";
+        public const string ACTION_RELOAD = "term.reload";
+        public const string ACTION_SCROLL_TO_COMMAND = "term.scroll-to-command";
+        public const string ACTION_SELECT_ALL = "term.select-all";
+        public const string ACTION_ZOOM_DEFAULT = "term.zoom::default";
+        public const string ACTION_ZOOM_IN = "term.zoom::in";
+        public const string ACTION_ZOOM_OUT = "term.zoom::out";
+
+        public const string[] ACCELS_COPY = { "<Control><Shift>C", null };
+        public const string[] ACCELS_COPY_OUTPUT = { "<Alt>C", null };
+        public const string[] ACCELS_PASTE = { "<Control><Shift>V", null };
+        public const string[] ACCELS_RELOAD = { "<Control><Shift>R", "<Ctrl>F5", null };
+        public const string[] ACCELS_SCROLL_TO_COMMAND = { "<Alt>Up", null };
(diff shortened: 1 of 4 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -261,22 +335,183 @@ namespace Terminal {
...
+
+                if (!text.validate ()) {
+                    warning ("Dropping invalid UTF-8 paste");
+                    return;
+                }
+
+                unowned var toplevel = (MainWindow) get_toplevel ();
+
+                if (!toplevel.unsafe_ignored && Application.settings.get_boolean ("unsafe-paste-alert")) {
+                    string? warning = null;
```
jeremypw, line 485: Best not use the name of a command as a variable name.  Cannot insert `warning ()` commands into the code when this variable is in scope.

src/Widgets/TerminalWidget.vala
```diff
@@ -261,22 +335,183 @@ namespace Terminal {
...
+                    text._strip ();
```
jeremypw, line 486: This does not work for some reason.  No dialog is shown when `sudo` is included (shown in master).

src/Widgets/TerminalWidget.vala
```diff
@@ -261,22 +335,183 @@ namespace Terminal {
...
+
+                    if ("\n" in text) {
```
jeremypw, line 488: This alters the logic of the test compared to the original causing the warning to not show if there is no newline.  Was this the intention? Probably best to push a separate PR for changes in the logic and keep this one purely for moving code.

src/Widgets/TerminalWidget.vala
```diff
@@ -187,18 +214,17 @@ namespace Terminal {
...
-                resized = true;
-            });
-
+            selection_changed.connect (() => copy_action.set_enabled (get_has_selection ()));
+            size_allocate.connect (() => resized = true);
             contents_changed.connect (check_cwd_changed);
-
             child_exited.connect (on_child_exited);
+            ulong once = 0;
+
```
jeremypw, line 222: I would remove this space.

jeremypw (changes requested):
I have left some inline comments.  Everything else seems OK but I'll test a but more later.

src/Widgets/TerminalWidget.vala
Marukesu: No, it was not a intended change, thanks for catching it.

jeremypw (changes requested):
The "copy last output" and "scroll to last command" actions are not triggered by the keyboard (but the former is triggered by the context menu option).  Something to do with the "key pressed" handler I think.

jeremypw (approved):
Its a pity this change turned out to be less straightforward than anticipated and a somewhat non-obvious workround is required to avoid breaking some actions.  However this all seems to work now.  So I guess it can be merged.

Maybe possible to refactor simplify later?

PR #751, opened 2023-12-03
TerminalWidget: Port Context Menu to GLib.MenuModel and Gtk.PopoverMenu

Marukesu:
In preparation to Gtk4. The changing from Gtk.Menu to Gtk.PopoverMenu happens because Gtk.Menu will inherit the monospace font from the TerminalWidget, sadly, it requires that we lost the accels in the menu for now.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -27,7 +27,9 @@ namespace Terminal {
         private Gtk.Stack title_stack;
         private Gtk.ToggleButton search_button;
         private Dialogs.ColorPreferences? color_preferences_dialog;
-        private Granite.AccelLabel open_in_browser_menuitem_label;
+
+        private GLib.Menu open_in_menu;
+        private GLib.Menu search_menu;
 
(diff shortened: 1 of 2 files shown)


jeremypw:
How near are we to switching to Gtk4? Would it be better to leave this until switch is imminent so degradation of UI is minimised?

Marukesu:
> How near are we to switching to Gtk4?

We don't have the Gtk4 version of VTE in jammy, so it would need to wait for once we are "done" with releases for OS 7.

> Would it be better to leave this until switch is imminent so degradation of UI is minimised?

I'm fine in waiting for it, this part of code won't have much changes, if any, until there.

jeremypw:
Adding blocked label - waiting for Gtk4 version of Vte.

jeremypw:
Looks like the context menu is broken in master at the moment, so it might be better to proceed with this.  Might be possible work around the accelerator issue. I'll look into it.

danirabbit:
Closing since #769 was merged

PR #753, opened 2023-12-09
Fix inoperative header decorations when fullscreened

### LINKED ISSUE #752: The F11 shortcut key causes the maximize/unmaximize button to fail, in termial,
### What Happened?

The F11 shortcut key causes the maximize/unmaximize button to fail.  Double-clicking on the title bar also does not restore. The only way to recover is to press F11 again. Double-clicking on the title bar also does not restore.  The default maximization shortcut is Super+Up. I can't figure out where this F11 shortcut came from.

### Steps to Reproduce

1. Open a Terminal.
2. Press F11.
At this time,  the maximize/unmaximize button is disabled.  Double-clicking on the title bar also does not restore.

### Expected Behavior

At any time, it should be possible to maximize or exit by double-clicking on the title bar. Or  press super+up

### OS Version

7.x (Horus)

### Software Version

### Log Output

```shell
It is not possible to update flatpak in China.
```

### Hardware Info

_No response_

jeremypw:
Fixes #752 

When fullscreened, the default decoration is overridden to show only the Close button (which still works)  and a custom "Exit Fullscreen" button is revealed to replace the inoperative "Maximize/UnMaximise" button (with the same icon but with a Granite tooltip).  This follows the method used by Epiphany (except that Epiphany does not use a tooltip).

If this solution is accepted then it could also be used for e.g. Code which has the same issue.


src/MainWindow.vala
```diff
@@ -30,7 +31,22 @@ namespace Terminal {
...
+                if (value) {
+                    fullscreen ();
+                    header.decoration_layout_set = true;
+                    //TODO Show Toast like Chrome? or additional unfullscreen button?
+                } else {
+                    unfullscreen ();
+                    header.decoration_layout_set = false;
+                }
+            }
+        }
```
Marukesu, line 49: you can make this query the window state, that matches the behaviour of Gtk.Window.fullscreened in Gtk4.

```suggestion
        private bool is_fullscreen {
            get {
                if (!get_realized ()) {
                    return false;
                }
                
                return FULLSCREEN in get_window ().get_state ();
            }

            set {
                if (value) {
                    fullscreen ();
                } else {
                    unfullscreen ();
                }
            }
        }
```

then you can do `bind_property ("is-fullscreen", header, "decoration-layout-set", INVERT_BOOLEAN)`.

src/MainWindow.vala
```diff
@@ -285,6 +301,8 @@ namespace Terminal {
 
             unowned Gtk.StyleContext header_context = header.get_style_context ();
             header_context.add_class ("default-decoration");
+            header.decoration_layout = "";
```
Marukesu, line 304: this could be set at construction, no?

Marukesu (commented):
A litte suggestion with Gtk4 in mind.

Also, a pior art in this case is epiphany, that `pack_end()` a Gtk.Button that trigger the fullscreen action and is only visible when fullscreened.

something like this should be simple to do here too.

src/MainWindow.vala
jeremypw: Yes, that's better - thanks.  I was aware that fullscreen () cannot be guaranteed to actually fullscreen the window.  That is a nice solution if it works.  I am not sure that `bind_property` works with a property that is not backed by a private variable though - `notify` doesn't.

src/MainWindow.vala
jeremypw: Sure.

src/MainWindow.vala
Marukesu: GLib.Binding uses notify underneath, so yeah, i believe you will need to call `notify_property()` in a window_state_event handler for that.

jeremypw: I tried getting this to work but there seems to be some weird timing issues where changing `decoration_layout_set` in `configure-event` has no effect.  As this is all going to have to be rewritten for Gtk4 anyway I do not want to spend more time on it.  Is it OK to go with the current solution?

Marukesu: Fine for me.

src/MainWindow.vala
```diff
@@ -245,6 +260,17 @@ namespace Terminal {
...
+            var unfullscreen_button = new Gtk.Button.from_icon_name ("view-restore-symbolic") {
+                action_name = ACTION_PREFIX + ACTION_FULLSCREEN,
+                margin_start = 12,
+                no_show_all = true,
+                visible = false
+            };
+            unfullscreen_button.tooltip_markup = Granite.markup_accel_tooltip (
+                action_accelerators[ACTION_FULLSCREEN].to_array (),
+                _("Exit FullScreen")
+            );
```
danirabbit, line 272: Make sure to style this the same as other title buttons

```suggestion
            );
            unfullscreen_button.get_style_context ().remove_class ("image-button");
            unfullscreen_button.get_style_context ().add_class ("titlebutton");
```

src/MainWindow.vala
```diff
@@ -245,6 +260,17 @@ namespace Terminal {
         }
 
         private void setup_ui () {
+            var unfullscreen_button = new Gtk.Button.from_icon_name ("view-restore-symbolic") {
+                action_name = ACTION_PREFIX + ACTION_FULLSCREEN,
```
danirabbit, line 264: Title buttons don't receive keyboard focus

```suggestion
                action_name = ACTION_PREFIX + ACTION_FULLSCREEN,
                can_focus = false,
```

danirabbit (changes requested):

Marukesu:
@jeremypw you already implemented my suggestion. a fullscreen button in the menu would be good indeed, but that can be left to another PR.

danirabbit (approved):
Works as expected, nice job!

PR #760, opened 2024-06-20
Replace Granite.DynamicNotebook with Hdy TabView and TabBar

jeremypw:
This adapts the DocumentView in Code which is already based on the Hdy widgets.

This mostly reproduces current behaviour except dropping files onto the tab bar now works.  

Note that the empty part of the tabbar is not accessible, it seems, so drops have to be either on the add button (which creates a new tab) or onto an existing tab (which changes the current directory.  Only local files and folders can be dropped.

Duplicating a tab creates the new tab next to the target tab (like in Browser)

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -356,7 +356,7 @@ namespace Terminal {
             notebook.tab_view.notify["selected-page"].connect (() => {
                 var term = get_term_widget (notebook.tab_view.selected_page);
                 if (term == null) {
-                    critical ("No terminal in selected page");
+                    // Happens on closing window - ignore
                     return;
                 }
 
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -20,7 +20,7 @@ namespace Terminal {
     public class MainWindow : Hdy.Window {
         private Pango.FontDescription term_font;
         private Hdy.HeaderBar header;
-        public TerminalView notebook { get; set construct; }
+        public TerminalView notebook { get; private set construct; }
         private Gtk.Clipboard clipboard;
         private Gtk.Clipboard primary_selection;
         private Terminal.Widgets.SearchToolbar search_toolbar;
@@ -30,7 +30,6 @@ namespace Terminal {
         private Dialogs.ColorPreferences? color_preferences_dialog;
         private Granite.AccelLabel open_in_browser_menuitem_label;
 
-        private HashTable<string, TerminalWidget> restorable_terminals;
         private bool is_fullscreen {
             get {
                 return header.decoration_layout_set;
@@ -64,22 +63,17 @@ namespace Terminal {
         public SimpleActionGroup actions { get; construct; }
         public TerminalWidget? current_terminal {
              get {
-                if (notebook.selected_page != null) {
-                    return get_term_widget (notebook.selected_page);
-                } else {
-                    return null;
-                }
+                return get_term_widget (notebook.selected_page);
              }
         }
 
-        public GLib.List <TerminalWidget> terminals = new GLib.List <TerminalWidget> ();
 
         public const string ACTION_PREFIX = "win.";
         public const string ACTION_CLOSE_TAB = "action-close-tab";
         public const string ACTION_CLOSE_TABS_TO_RIGHT = "action_close_tabs_to_right";
         public const string ACTION_CLOSE_OTHER_TABS = "action_close_other_tabs";
         public const string ACTION_FULLSCREEN = "action-fullscreen";
-        public const string ACTION_NEW_TAB = "action-new-tab";
+        public const string ACTION_NEW_TAB = "action-term_widgetnew-tab";
         public const string ACTION_RESTORE_CLOSED_TAB = "action-restore-closed-tab";
         public const string ACTION_DUPLICATE_TAB = "action-duplicate-tab";
         public const string ACTION_NEXT_TAB = "action-next-tab";
@@ -228,8 +222,6 @@ namespace Terminal {
 
             set_size_request (app.minimum_width, app.minimum_height);
 
-            restorable_terminals = new HashTable<string, TerminalWidget> (str_hash, str_equal);
-
             restore_saved_state ();
             show_all ();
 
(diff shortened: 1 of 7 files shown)


jeremypw:
Outstanding issue is a transparent gap between the tab bar and the header bar through which the desktop shows:


Screenshot from 2024-06-21 10 33 00

This occurs on X and on Wayland,  but does not occur in Files and Code which have very similar widgets.

@danirabbit Is this a styling issue or a difference in coding that I can't see?

jeremypw:
@danirabbit The problem seems to be due to the "terminal-window" styleclass previously applied to the grid containing the header and view.  I have removed that class for now.  I am not sure what it did or how to fix it.

jeremypw:
@danirabbit The problem seems to be due to the styleclass "terminal-window" previously applied to the grid containing the header and view.  I have removed it for now. I don't know what it was doing or how to fix it.

danirabbit (changes requested):
It looks like zoom controls are broken with this branch:

Screenshot from 2024-06-21 13 55 51

danirabbit:
@jeremypw proposed a fix here :) https://github.com/elementary/terminal/pull/761

src/MainWindow.vala
```diff
@@ -816,54 +747,61 @@ namespace Terminal {
                 terminal_widget.run_program (program, location);
             }
 
-            save_opened_terminals (true, true);
+            terminal_widget.child_exited.connect (() => {
+                if (!terminal_widget.killed) {
```
leolost2605, line 751: I think this will cause a memory leak since the lambda keeps a reference on us and the terminal widget so neither of them will be freed :(

See https://gitlab.gnome.org/GNOME/vala/-/issues/1548

leolost2605 (changes requested):

jeremypw:
@leolost2605 Yes, you are right - TerminalWidgets are not being destroyed when tabs are closed.  I had similar issues a long time ago with Files widgets and ended up just disconnecting the signals before closing.  I'll see if I can find a better way here.

leolost2605:
@jeremypw for the terminal_widget you can use the instance that's always passed in signal handlers (so something like `terminal_widget.child_exited.connect ((instance, any_other_variables_that_are_passed_to_the_signal_handler, ...) => {` which might be enough to break the cycle? Atleast if the terminal widgets lifetime doesn't depend on `this`. If it does then the program local variable is still an issue so maybe add a property in the terminal widget for it? I'm not at all in codebase though so I'm not sure whether that makes any sense here :)

leolost2605:
I'm currently trying to write the logics here down maybe for inclusion in docs.vala.dev and for me to better understand it since it seems to be a pretty big problem especially in widget hierarchies so I'll leave it here might be completely wrong though :)

> A word on vala memory management with callbacks.

> 
> With signals:
> If local variables are used inside signals vala will ref the current this (see glib.signal.connect_object, which is used if no local variables are used vs. glib.signal.connect_data_full which is used if local variables are used). This can cause memory leaks because of reference cycles if the signal is not manually disconnected and the lifetime of the object that the signal was connected to depends on the this lifetime. This is for example the case if the this is a widget and we add another widget to it and connect to a signal of that widget. This is also the case if the object where the signal was connected to itself is a local variable and used within the lambda. The latter case can be easily mitigated by using the instance that will always be given as an (optional) first argument to the signal handler instead of the local variable.
> 
> TLDR: If you don't know what you are doing try to avoid using local variables in lambdas.
> 
> For other callbacks individual cases have to be looked at. Here it might be worth looking at the c code and what code vala generates. For example
> ```
> 
> public class Immortal : Object {
>     public GTK.SortListModel sorted_items;
> 
>     construct {
>         var items = new ListStore (typeof(SomeItemClass));
> 	var sorter = new GTK.CustomSorter (sort_func);
>         sorted_items = new GTK.SortListModel (items, sorter);
>     }
> 
>     private int sort_func (Object? obj1, Object? obj2) {
> 	//Sort this stuff
>         return 0;
>     }
> }
> ```
> If you use this class in you GTK App for example to display a list and then the user navigates somewhere else and the list should be destroyed and freed. Then this class will never be freed even if e.g. the widget that holds a reference on this will be freed. This is because the sorter "holds" a reference on this and we hold a reference on it (via the sorted_items model). To break this cycle we can add a method
> ```
>     public void cleanup () {
>         sorted_items = null;
>     }
> ```
> which should be called by the destructor of the class that uses this class (e.g. the listview that displays our items).

jeremypw:
@Marukesu This branch breaks some of the tests (although in reality it behaves OK).  I have been unable to find out what is going wrong.  It appears to be a timing thing for the "execute" test - the number of tabs gets tested *before* the context iteration (?).  Putting debug messages in the code indicates that 5 tabs are indeed added by MainWindow but the test only sees one tab.

I changed the closing down code in order to make the other tests run.

If you have time I would be grateful for your input/investigation/fix :smile:

jeremypw:
@leolost2605  For now I have used disconnection of terminal widget signals before closing in order for them to be destroyed.  If you have a more elegant way please let me know.

Marukesu:
some things:

1. TerminalWidget.program_string is never set, so `--execute` is broken (this is the reason for the test failure).

2. Vte.Terminal.child_exited will be called if the TabView is destroyed with tab opens (happens with the shutdown signal during the tests) and the close_page call inside it will cause a segfault, a extra `tw.tab.child == tw.parent` check should stop it from occurring.

3. The Idle block during the `selected-page` notify is unnecessary and setting the tab tooltip there causes the `action/new-window` test to fail, it can be removed since the tooltip is already set during the tab construction and when the cwd changes.

4. `title = current_terminal.current_working_directory` is a regression as it doesn't respect the user/program defined title.

jeremypw:
@Marukesu Thanks for your helpful comments - I have tried to implement them.  After those changes, the `action-quit` test was still not working (timing out) so I have have made some further changes to `tests/Application.vala` that do allow it to pass.  However, as you have a much deeper knowledge of what is going on these tests I would be grateful for your input.

With regard to the window title, it now tracks the current terminal `window-title` property (read only). Is that what is intended?

jeremypw:
@Marukesu Just noticed that the test pass in two of the builds but not in the development-target build. I have not idea why (It pass locally on OS8)

src/Widgets/TerminalWidget.vala
```diff
@@ -1,4 +1,5 @@
 // -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
+// -*- Mode: vala; indent-tabs-mode: nil; tab-width: 4 -*-
```
Marukesu, line 2: Duplicated line.

src/Widgets/TerminalWidget.vala
```diff
@@ -26,43 +27,40 @@ namespace Terminal {
...
             get {
-                return _tab_label;
+                return tab != null ? tab.title : "";
             }
 
             set {
-                if (value != null) {
-                    _tab_label = value;
-                    tab.label = tab_label;
+                            if (value != null && tab != null) {
```
Marukesu, line 62: Wrong indentation.

Marukesu (approved):
The tests look alright, the last fail is not consistent, and i believe there's no way to get the CI's coredumps to understand what happened there, sadily.

danirabbit (dismissed):
LGTM :tada:

jeremypw:
@Marukesu Is this OK to merge now, do you think?

ryonakano:
@jeremypw Thank you for this! Could you add `src/Widgets/TerminalView.vala` to POTFILES so that we can translate menu items?

jeremypw:
@ryonakano Done! Sorry - I always forget to do that :frowning_face:

Marukesu (approved):

PR #761, opened 2024-06-21
Fix transparent background

danirabbit:
This style class is used to make sure custom colors can use alpha. Otherwise they'll just be overlaying on a solid color background


jeremypw (approved):
Thanks for the fix!

PR #768, opened 2024-08-04
Fix empty context menu regression

jeremypw:
Reinsert `show_all ()`

Also changes non-specific name `menu` to more specific `context_menu` to more easily distinguish from MenuButton popover menu.


danirabbit:
I can confirm the fix for `show_all` but this branch is failing tests. Do tests rely on this being named menu maybe?

jeremypw:
@danirabbit The tests are now passing after merging the latest master although no obvious reason why :shrug: .  None of the changes here should affect the tests and indeed they passed locally.

danirabbit (approved):

PR #769, opened 2024-08-05
Use GLib.MenuModel for context menu

jeremypw:
This is similar to #751 but based on latest master and more closely following current code to reduce diff (a little).  Code cleanup can be done in separate PR.

Also added custom "accel" menu attribute to the items.  This does not seem to be working with the current version - maybe it ill work in Gtk4?

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -194,19 +194,20 @@ namespace Terminal {
             );
             select_all_menuitem.set_attribute_value ("accel", new Variant ("s", TerminalWidget.ACCELS_SELECT_ALL[0]));
 
-            context_menu_model = new Menu ();
-            // "Open in" item must be in position 0 (see update_menu_label ())
-            context_menu_model.append_item (open_in_browser_menuitem);
-
             var terminal_action_section = new Menu ();
             terminal_action_section.append_item (copy_menuitem);
             terminal_action_section.append_item (copy_last_output_menuitem);
             terminal_action_section.append_item (paste_menuitem);
             terminal_action_section.append_item (select_all_menuitem);
-            context_menu_model.append_section ("", terminal_action_section);
+
             var search_section = new Menu ();
             search_section.append_item (search_menuitem);
-            context_menu_model.append_section ("", search_section);
+
+            context_menu_model = new Menu ();
+            // "Open in" item must be in position 0 (see update_menu_label ())
+            context_menu_model.append_item (open_in_browser_menuitem);
+            context_menu_model.append_section (null, terminal_action_section);
+            context_menu_model.append_section (null, search_section);
 
             setup_ui ();
 
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -158,7 +158,7 @@ namespace Terminal {
 
             //Window actions
             open_in_browser_menuitem = new MenuItem (
-               "DUMMY open in name",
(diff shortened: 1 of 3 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -42,12 +42,13 @@ namespace Terminal {
             }
         }
 
-        private Gtk.Menu menu {
+        private Menu context_menu_model {
```
danirabbit, line 45: Hm so I get a warning here about:
```
../src/Widgets/TerminalWidget.vala:45.9-45.39: warning: Terminal.TerminalWidget.context_menu_model hides inherited property `Vte.Terminal.context_menu_model'. Use the `new' keyword if hiding was intentional
   45 |         private Menu context_menu_model {
```

I'm actually not sure what is the purpose of this property? it seems to only be used in a single private function

src/Widgets/TerminalWidget.vala
```diff
@@ -42,12 +42,13 @@ namespace Terminal {
...
             get {
-                return main_window.menu;
+                return main_window.context_menu_model;
             }
         }
 
+        private Gtk.PopoverMenu context_popover;
```
danirabbit, line 51: This should probably not be `Gtk.PopoverMenu` in Gtk3, but instead `Gtk.Menu` and we can change it to `PopoverMenu` in Gtk4

src/Widgets/TerminalWidget.vala
```diff
@@ -460,11 +451,26 @@ namespace Terminal {
...
+
+            // Popup context menu below cursor position
+            context_popover = new Gtk.PopoverMenu () {
+                relative_to = this,
+                pointing_to = rect
+            };
+
+            context_popover.bind_model (context_menu_model, null);
+            context_popover.show_all ();
+            context_popover.popup ();
```
danirabbit, line 471: I think `show_all` and `popup` are redundant here. I'm not sure `show_all` is necessary when using a menu from a model

danirabbit (changes requested):

src/Widgets/TerminalWidget.vala
jeremypw: Now removed.

src/MainWindow.vala
```diff
@@ -154,53 +156,57 @@ namespace Terminal {
...
             primary_selection = Gtk.Clipboard.get (Gdk.Atom.intern ("PRIMARY", false));
 
-            var open_in_browser_menuitem = new Gtk.MenuItem () {
-                action_name = ACTION_PREFIX + ACTION_OPEN_IN_BROWSER
-            };
-            open_in_browser_menuitem_label = new Granite.AccelLabel.from_action_name (
-                "", open_in_browser_menuitem.action_name
+            //Window actions
+            open_in_browser_menuitem = new MenuItem (
+               "DUMMY open in name",
```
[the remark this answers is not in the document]
jeremypw: Fixed

src/Widgets/TerminalWidget.vala
jeremypw: It does not work in Gtk3 without these.  Revisit in Gtk4

danirabbit:
@jeremypw proposed https://github.com/elementary/terminal/pull/770 to merge into this branch. Fixes missing accels

danirabbit (approved):

PR #770, opened 2024-08-08
Use Gtk.Menu

danirabbit:
Needs some CSS here to fix stylesheet bugs, but this will show accelerators. Gotta make the section labels null or we get a weird bug with separator height being forced to match an empty label

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -43,3 +43,12 @@ tabbar .box {
     margin: 0;
     padding-top: 2px;
 }
+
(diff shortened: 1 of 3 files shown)


jeremypw:
I see you have used Gtk.Menu instead of GLib.Menu to get the accel labels back.  This will have to be changed for Gtk4 as Gtk.Menu does not exist there but hopefully by then there will be a way of showing accel labels in GLib.Menu?

jeremypw (approved):

PR #772, opened 2024-08-10
Show zoom level in centre or terminal when changing.

jeremypw:
#626 Based on igordsm/zoom-overlay incorporating @danirabbit 's suggestions.

Granite.OverlayBar is used for simplicity and consistency with Files.


jeremypw:
Should ZoomOverlay widget be merged into MainWindow now thereby losing a Gtk.Bin?

danirabbit:
@jeremypw sorry thought I responded to this. What do you think about using Granite.Overlaybar for this? https://github.com/elementary/terminal/pull/773

jeremypw:
@danirabbit Looks like it would be appropriate - thanks.  Is Granite-7 available on OS7.1 though?

danirabbit:
@jeremypw it is! But granite-7 is just for gtk4 apps. This uses overlay bar from granite 6

jeremypw:
OK, sorry for the confusion.  I think we would still have to use markup to get 24pt text though? Are you OK with that size? I feel it needs to be more prominent than default text size if displayed over text.

jeremypw:
@danirabbit  I changed it to Granite.OverlayBar but then noticed you had already pushed a new PR!  The main difference is that the size and position of the overlaid widget is not controllable so the vision of the OP cannot be realized.  However using the native Granite widget has the advantage of being consistent with the Files appearance when file items are selected.  Also fewer code lines.

danirabbit (approved):
Clean!

PR #774, opened 2024-08-20
upd(TerminalWidget): add doas checking

Elsie19:
Doas is also a program used to gain administrative access so I thought I would add it here just to cover all corners.


danirabbit (approved):
Makes sense to me, thanks for this!

danirabbit (approved):

PR #776, opened 2024-08-31
Fix "Open in Terminal" from Files when Terminal instance already running

### LINKED ISSUE #775: "Open in Terminal" menu option in Files does not work correctly when uri contains spaces
### What Happened?

If the "Open in Terminal" context menu option is used for a location containing spaces it opens a tab at the user Home directory instead.

### Steps to Reproduce

1. Open Files and navigate to a location whose uri contains spaces
2. Use the background context menu option "Open with ... Terminal"
3. Terminal opens with an extra tab at the Home folder

### Expected Behavior

Terminal opens with an extra tab at the requested location

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #775

The exact reason is uncertain but it appears that passing the working directory via the platform-data fails in some cases.  The solution suggested here does not remove the "working-directory" commandline option after setting the CWD and uses that to determine where to open the tab when it is available, falling back to the CWD when it is not.

The sanitization of the uri no longer escapes the uri as this is not needed (and can cause problems)

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -104,17 +104,18 @@ public class Terminal.Application : Gtk.Application {
 
     protected override int handle_local_options (VariantDict options) {
         unowned string working_directory;
+        string sanitized_path;
         unowned string[] args;
 
         if (options.lookup ("working-directory", "^&ay", out working_directory)) {
             if (working_directory != "\0") {
-                Environment.set_current_dir (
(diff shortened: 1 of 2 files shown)


jeremypw:
Closed in favour of #777.  It appears it was the escaping of the uri of the working directory alone that was causing the problem.

PR #777, opened 2024-08-31
Correctly open uris containing spaces

### LINKED ISSUE #775: "Open in Terminal" menu option in Files does not work correctly when uri contains spaces

jeremypw:
Fixes #775


teamcons:
Any ETA for merging this bugfix ? The PR seems stale

src/Widgets/TerminalWidget.vala
```diff
@@ -776,21 +776,27 @@ namespace Terminal {
...
                     File file;
-
                     for (var i = 0; i < uris.length; i++) {
-                         file = File.new_for_uri (uris[i]);
-                         if ((path = file.get_path ()) != null) {
-                             uris[i] = Shell.quote (path) + " ";
+                        // Get sanitized unquoted path (Files sends uris that are escaped and quoted)
+                        // We do not want the `file://` scheme included
+                        // We assume dropped paths are absolute
+                        file = File.new_for_uri (Utils.sanitize_path (Shell.unquote (uris[i]), "", false));
```
ryonakano, line 783: This line causes the following warning on build:

```
../src/Widgets/TerminalWidget.vala:783.71-783.93: warning: unhandled error `GLib.ShellError'
  783 |                         file = File.new_for_uri (Utils.sanitize_path (Shell.unquote (uris[i]), "", false));
      |                                                                       ^~~~~~~~~~~~~~~~~~~~~~~              
```

src/Widgets/TerminalWidget.vala
```diff
@@ -776,21 +776,27 @@ namespace Terminal {
                     var uris = selection_data.get_uris ();
                     string path;
                     File file;
-
                     for (var i = 0; i < uris.length; i++) {
-                         file = File.new_for_uri (uris[i]);
-                         if ((path = file.get_path ()) != null) {
-                             uris[i] = Shell.quote (path) + " ";
+                        // Get sanitized unquoted path (Files sends uris that are escaped and quoted)
```
ryonakano, line 780: > Files sends uris that are escaped and quoted

I'm not sure if this behavior is just about Files. I'm not sure about each implementation, but this branch works as expected if I open Terminal from Code, GNOME Files, and Dolphin, the file manager made for KDE.

(I'm concerning this comment sounds like as if this code only supports URIs passed from Files and does not support them passed from other apps)

src/Widgets/TerminalWidget.vala
jeremypw: ShellError now handled

src/Widgets/TerminalWidget.vala
jeremypw: I have amended the comment accordingly

jeremypw:
@ryonakano Thanks for the review. I have now addressed your comments.

ryonakano (approved):

teamcons:
_**YES**_
Thank you !!!!

PR #778, opened 2024-09-02
Implement Clear and Reset actions with hotkeys

### LINKED ISSUE #181: Support clearing the screen mid process (Reset + Clear)
CTRL-L will only clear the window when the bash shell is ready for a command.

If for instance you have a server process running with many logs, you need to clear the window to get rid of the clutter then currently its not possible unless you exit the process, hit CTRL-L and restart the process.

This is possible in Gnome Terminal with Reset + Clear and in OSX Terminal with CMD + K 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52617346-support-clearing-the-screen-mid-process-reset-clear?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F61246377&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #181 (?)

I am unsure whether the Clear action should also clear the scrollback buffer.  At the moment it does not because the "Reset" action does that.  The native hotkey `<Control>L` does not clear the buffer but typing `clear` into the terminal does.

Is it worth implementing two actions - "Clear Screen Only" and "Clear Screen and Scrollback"?

The XFCE terminal labels the reset action "Clear Scrollback and Reset" - should we follow or is that too verbose?

Should the "Clear Screen" menu option be labelled "Clear Screen Only"?


zeebok (approved):
Works as advertised. To me this resolves what they were asking for, though it does make me curious what was handling the CTRL + L shortcut before this?

jeremypw:
Lets merge and refine with further PRs if required.

PR #780, opened 2024-09-04
Port to Gtk4

jeremypw:
Compiles and runs but pretty broken.

Need to extract what can be done in Gtk3 first.

Notes:
 - CheckButtons do not seem to work with only an action_target (no action name) in Gtk4 so the theme string is now added as data. Smaller diff than creating an action to change the theme.


jeremypw:
CI tests failing both locally and here - some obscure errors about MESA, glx etc.  Not sure how to fix this. Will see if any other Gtk4 projects have working tests.

jeremypw:
@danirabbit I have a problem with styling the theme buttons in the SettingsPopover.  It looks like valid CSS is being loaded into the css_provider but it has no effect.   Maybe you can shed light?

src/Application.vala
```diff
@@ -180,7 +180,7 @@ public class Terminal.Application : Gtk.Application {
 
     protected override void startup () {
         base.startup ();
-        Hdy.init ();
+        Adw.init ();
```
danirabbit, line 183: Need to Granite.init () first to avoid getting Adw styles and make sure we load Granite styles
```suggestion
        Granite.init ();
        Adw.init ();
```

src/Widgets/SettingsPopover.vala
```diff
@@ -13,14 +13,15 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
...
+            //TODO Do we need to do this in Gtk4?
+            // if (value != null) {
+            //     insert_action_group ("term", value.get_action_group ("term"));
+            // }
         }
     }
 
     private const string STYLE_CSS = """
-        .color-button radio {
+        .theme-button check {
```
danirabbit, line 24: the CSS node here is still `radio`
```suggestion
        .theme-button radio {
```

src/Widgets/SettingsPopover.vala
```diff
@@ -157,54 +158,63 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
...
+        button.set_data<string> ("theme", theme);
         css_provider = new Gtk.CssProvider ();
 
-        unowned var style_context = button.get_style_context ();
-        style_context.add_class (Granite.STYLE_CLASS_COLOR_BUTTON);
-        style_context.add_provider (css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        button.add_css_class (Granite.STYLE_CLASS_COLOR_BUTTON);
+        button.add_css_class ("theme-button");
+
+        Gtk.StyleContext.add_provider_for_display (
```
danirabbit, line 174: If you want to add the provider to the display then you also need to give the button a unique css class and make sure to add that class to the selector in the custom CSS as well. Otherwise all buttons are sharing this style.

Maybe better to leave that for a follow up branch since it's just deprecated behavior and won't block a port?

data/Application.css
```diff
@@ -30,11 +30,11 @@ vte-terminal {
...
 
-.color-custom radio {
+.color-custom check {
     -gtk-icon-source: -gtk-icontheme("list-add-symbolic");
 }
 
-.color-custom radio:checked {
+.color-custom check:checked {
     -gtk-icon-source: -gtk-icontheme("check-active-symbolic");
 }
```
danirabbit, line 40: Don't forget to remove the CSS below this line :)

src/Application.vala
```diff
@@ -167,7 +167,7 @@ public class Terminal.Application : Gtk.Application {
                 terminal.tab.icon = process_icon;
             }
 
-            if (!(Gdk.WindowState.FOCUSED in terminal.main_window.get_window ().get_state ())) {
+            if (!(Gtk.StateFlags.FOCUSED in terminal.main_window.get_state_flags ())) {
```
danirabbit, line 170: i think what you want instead here is:
```suggestion
            if (!get_active_window ().is_active) {
```

src/Widgets/SettingsPopover.vala
```diff
@@ -13,14 +13,15 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
...
+            //TODO Do we need to do this in Gtk4?
+            // if (value != null) {
+            //     insert_action_group ("term", value.get_action_group ("term"));
+            // }
         }
     }
 
     private const string STYLE_CSS = """
-        .color-button radio {
+        .theme-button radio {
```
danirabbit, line 24: `color-button` is actually a style class in the system stylesheet, so you might lose some styles from the system by changing this

data/Application.css
jeremypw: Workarounds removed

src/Widgets/SettingsPopover.vala
jeremypw: Reverted

src/Widgets/TerminalView.vala
```diff
@@ -65,12 +64,11 @@ public class Terminal.TerminalView : Gtk.Box {
         };
 
         tab_history_button = new Gtk.MenuButton () {
-            image = new Gtk.Image.from_icon_name ("document-open-recent-symbolic", Gtk.IconSize.MENU),
-            tooltip_text = _("Closed Tabs"),
-            use_popover = false,
```
danirabbit, line 70: To fix this you can add `tab_history_button.popover.has_arrow = false;` right after `tab_history_button.menu_model = new Menu ();` on line 110

data/Application.css
```diff
@@ -30,21 +30,11 @@ vte-terminal {
     padding: 0 9px;
 }
 
-.color-custom radio {
+.color-custom check {
     -gtk-icon-source: -gtk-icontheme("list-add-symbolic");
 }
 
-.color-custom radio:checked {
+.color-custom check:checked {
```
danirabbit, line 37: These should also still be radio

jeremypw: Done

src/Widgets/TerminalView.vala
jeremypw: Done

src/MainWindow.vala
```diff
@@ -434,11 +406,9 @@ namespace Terminal {
                     color_preferences_dialog = new Dialogs.ColorPreferences (this);
                 }
 
-                color_preferences_dialog.destroy.connect (() => color_preferences_dialog = null);
                 color_preferences_dialog.present ();
             });
 
-            bind_property ("title", header, "title", GLib.BindingFlags.SYNC_CREATE);
```
danirabbit, line 441: Do we not need this still?

src/MainWindow.vala
```diff
@@ -825,24 +743,22 @@ namespace Terminal {
...
             GLib.Icon? icon,
             TerminalWidget term,
             int pos
         ) {
-            var sw = new Gtk.ScrolledWindow (null, null) {
+            var sw = new Gtk.ScrolledWindow () {
                 vadjustment = term.get_vadjustment (),
                 child = term
             };
-
```
danirabbit, line 838: White space is good :)

src/MainWindow.vala
```diff
@@ -877,30 +793,50 @@ namespace Terminal {
...
+                        })
+                    );
+
+                    return Gdk.EVENT_STOP;
                 }
             }
 
-            foreach (var t in tabs_to_terminate) {
-                t.term_ps ();
+            terminate_all ();
```
danirabbit, line 822: Won't this just run immediately without waiting for confirmation? This seems like it should be in an early return instead

src/MainWindow.vala
```diff
@@ -877,30 +793,50 @@ namespace Terminal {
             }
         }
 
+        private bool close_immediately = false;
         //TODO Make TerminalWidget.confirm_kill_fg_process asynchronous and terminate all in callback
         public bool on_delete_event () {
+            if (close_immediately) {
```
danirabbit, line 799: I don't think I understand the addition of this

src/Utils.vala
```diff
@@ -136,7 +136,9 @@ namespace Terminal.Utils {
      */
     public bool is_safe_paste (string text, out string[]? msg_array) {
         string[] msgs = {};
-        if ("\n" in text || "&" in text || "|" in text || ";" in text ) {
+        var newline_index = text.index_of ("\n"); // First occurrence of new line
+        bool embedded_newline = newline_index >= 0 && newline_index < text.length - 1;
```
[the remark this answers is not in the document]
jeremypw: I've now pushed a PR that should remove this diff when merged.

src/Widgets/SettingsPopover.vala
```diff
@@ -13,16 +13,13 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
...
-                insert_action_group ("term", value.get_action_group ("term"));
-            }
         }
     }
 
     private const string ACTION_GROUP_NAME = "settings";
 
     private BindingGroup terminal_binding;
-    private Gtk.Box theme_buttons;
+    public Gtk.Box theme_buttons { get; private set; }
```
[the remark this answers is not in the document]
jeremypw: This is still a working part - without it and the associated widgets the theme buttons do not appear.  I noticed however that the theme buttons currently do not show the appropiate color so that needs fixing.  Need to check what the recommended Gtk4 solution was.

src/MainWindow.vala
```diff
@@ -18,27 +18,7 @@ namespace Terminal {
...
-                    header.decoration_layout = "close:";
-                    unfullscreen_button.visible = true;
-                    fullscreen ();
-                } else {
-                    header.decoration_layout = null;
-                    unfullscreen_button.visible = false;
-                    unfullscreen ();
-                }
-            }
-        }
```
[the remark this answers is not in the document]
jeremypw: It does not actually determine the state of the window and in Gtk4 there is now a function `is_fullscreen ()` to accurately determine the window state so it is not needed.  The code is incorporated into `action_fullscreen ()`.

src/MainWindow.vala
```diff
@@ -285,69 +268,63 @@ namespace Terminal {
...
                 hhomogeneous = false,
                 hexpand = false
             };
-
             search_toolbar.width_request = 300;
-            title_stack.add (title_label);
-            title_stack.add (search_toolbar);
+            title_stack.add_child (title_label);
+            title_stack.add_child (search_toolbar);
             // Must show children before visible_child can be set
```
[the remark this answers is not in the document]
jeremypw: Done

src/MainWindow.vala
jeremypw: Adw.HeaderBar does not have a "title" property - it is replaced by a "title-widget" property and we use that for the binding.

src/MainWindow.vala
jeremypw: This is intended to run *after* the user has confirmed that all processes should be terminated. I did not think a separate confirmation for each tab desirable (although I guess with extra code it could be accomodated).  I'll check that the behaviour is as intended.

src/meson.build
```diff
@@ -1,10 +1,6 @@
 terminal_deps = [
-    glib_dep,
-    gobject_dep,
-    gio_dep,
-    gtk_dep,
```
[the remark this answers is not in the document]
jeremypw: Reverted

meson.build
```diff
@@ -54,7 +52,7 @@ endif
 add_project_arguments(
     '-DGETTEXT_PACKAGE="@0@"'.format(meson.project_name()),
     '-DPCRE2_CODE_UNIT_WIDTH=0',
-    '-DHANDY_USE_UNSTABLE_API',
+    # '-DHANDY_USE_UNSTABLE_API',
```
danirabbit, line 55: This should be removed since we're not using handy anymore :)

src/Dialogs/ColorPreferencesDialog.vala
```diff
@@ -81,33 +81,30 @@ public class Terminal.Dialogs.ColorPreferences : Granite.Dialog {
...
+        light_cyan_button = new Gtk.ColorDialogButton (color_dialog);
+        white_button = new Gtk.ColorDialogButton (color_dialog);
+        background_button = new Gtk.ColorDialogButton (color_dialog);
+        foreground_button = new Gtk.ColorDialogButton (color_dialog);
+        cursor_button = new Gtk.ColorDialogButton (color_dialog);
 
         var contrast_top_label = new Gtk.Label (""); // Text will be set on showing
         var contrast_bottom_label = new Gtk.Label (""); // Text will be set on showing
-        var contrast_image = new Gtk.Image.from_icon_name ("process-completed", Gtk.IconSize.LARGE_TOOLBAR);
+        var contrast_image = new Gtk.Image.from_icon_name ("process-completed");
```
danirabbit, line 107: Probably needs `pixel size = 24`

src/Widgets/SettingsPopover.vala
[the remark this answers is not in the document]
danirabbit: I meant I don't think this needs to be publicly gettable

src/MainWindow.vala
```diff
@@ -434,11 +406,9 @@ namespace Terminal {
                     color_preferences_dialog = new Dialogs.ColorPreferences (this);
                 }
 
-                color_preferences_dialog.destroy.connect (() => color_preferences_dialog = null);
```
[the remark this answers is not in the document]
danirabbit: This seems to cause a bug where opening the color preferences dialog twice makes it impossible to close it

danirabbit (changes requested):
You've introduced a lot of `todo` and `fixme` and `note` comments here that are just things like "Make sure this works in GTK4". So I think those all need to be resolved/removed before merging

src/MainWindow.vala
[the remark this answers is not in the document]
jeremypw: OK, I'll investigate that bug.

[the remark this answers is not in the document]
jeremypw: Bug fixed - the dialog was being destroyed elsewhere - now only hides.

src/Widgets/TerminalWidget.vala
```diff
@@ -468,66 +490,79 @@ namespace Terminal {
...
-                );
-
-                if (dialog.run () == Gtk.ResponseType.ACCEPT) {
-                    dialog.destroy ();
-                    kill_fg ();
-                } else {
-                    dialog.destroy ();
-                    return false;
-                }
+                var dialog = new Gtk.AlertDialog (primary_text) {
```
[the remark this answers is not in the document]
danirabbit: Yeah I would really prefer we don't use the Gtk Dialogs they're designed to mimic Adwaita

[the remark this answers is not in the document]
jeremypw: Done in recent commits

danirabbit:
@jeremypw can you do any of these test changes in separate branches? The diff is now over 1,100 lines :(

lenemter:
Looks like color dialog styles got broken a bit:

# Gtk3
Screenshot from 2025-10-19 21 59 33

# Gtk4
Screenshot from 2025-10-19 21 59 03

src/Widgets/TerminalView.vala
```diff
@@ -48,16 +48,15 @@ public class Terminal.TerminalView : Gtk.Box {
         vexpand = true;
 
         var app_instance = (Gtk.Application) GLib.Application.get_default ();
-
-        tab_view = new Hdy.TabView () {
+        tab_view = new Adw.TabView () {
```
lenemter:
```suggestion

        tab_view = new Adw.TabView () {
```

src/Widgets/TerminalWidget.vala
```diff
@@ -784,8 +883,9 @@ namespace Terminal {
             }
         }
 
-        private string? get_link (Gdk.Event event) {
-            return this.match_check_event (event, null);
+        private string? get_link (double? x, double? y) {
+            int tag = 0;
+            return check_match_at (x, y, out tag);
```
lenemter:
```suggestion
            return check_match_at (x, y, null);
```

src/Widgets/TerminalWidget.vala
```diff
@@ -809,11 +909,11 @@ namespace Terminal {
             }
         }
 
-        protected override void increase_font_size () {
+        public new void increase_font_size () {
             font_scale += 0.1;
         }
 
-        protected override void decrease_font_size () {
+        public new void decrease_font_size () {
```
lenemter:
```suggestion
        public override void decrease_font_size () {
```

jeremypw:
Ah, the leading margin is a little larger in Gtk4. And the trailing margin and margin between grid and button.

src/Widgets/TerminalWidget.vala
jeremypw: This didn't used to work in Vala but I guess later versions have been fixed - thanks.

src/Widgets/TerminalView.vala
```diff
@@ -81,28 +79,42 @@ public class Terminal.TerminalView : Gtk.Box {
...
+        button_target.notify["value"].connect (() => {
+            var val = button_target.get_value ();
+            var uris = Uri.list_extract_uris (val.dup_string ());
+            foreach (string s in uris) {
+                if (!Utils.valid_local_uri (s, null)) {
+                    button_target.reject ();
+                    break;
+                }
+            }
+        });
```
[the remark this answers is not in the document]
jeremypw: Not sure why but sometimes the target value holds `null` instead of a string when entering or leaving the button so I have inserted a check for that.  If you drag a regular file then a tab for the parent is inserted.

lenemter (changes requested):

src/Widgets/TerminalWidget.vala
```diff
@@ -596,9 +659,11 @@ namespace Terminal {
         }
 
         private void update_current_working_directory (string cwd) {
-            current_working_directory = cwd;
-            tab.tooltip = current_working_directory;
-            cwd_changed ();
+            if (tab is Adw.TabPage) { // May not be the case if closing tab
```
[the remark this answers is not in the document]
jeremypw: OK, fair enough.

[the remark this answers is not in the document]
jeremypw: Now fixed

danirabbit:
Yup! I realized we don't have a packaging update prepared, so I will do that and then merge and rename the development branch etc

PR #781, opened 2024-09-05
Gtk4Prep: Use CheckButtons for theme buttons

jeremypw:
RadioButton does not exist in Gtk4
CheckButton group is not readable in Gtk4

Some extra code is required in Gtk3 to make the CheckButtons act as a group, which can be lost in Gtk4.

### DIFF
--- a/data/Application.css
+++ b/data/Application.css
@@ -30,11 +30,11 @@ vte-terminal {
     padding: 0 9px;
 }
 
-.color-custom radio {
+.color-custom check {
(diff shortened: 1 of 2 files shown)


danirabbit:
This seems like actually creating a larger diff away from a gtk4 port than making a smaller diff since gtkcheckbutton is pretty much a drop in replacement when they're already using the `group` property

jeremypw:
Yes, your probably right - it was the wrong thing to choose

PR #783, opened 2024-09-05
Gtk4Prep - Handle restoring and syncing window state in Application

jeremypw:
- Gtk4 Does not have configure-event
 - Follow pattern of Icon Browser (Gtk4) with regards to settings 
 - Restore saved state function can just be deleted on porting
 - Replaced by settings binding to window properties.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -288,12 +288,35 @@ public class Terminal.Application : Gtk.Application {
             window.add_tab_with_working_directory (working_directory, null, new_tab);
         }
 
+        //TODO In Gtk4 we can just bind the settings to first window properties
+        // instead of this function
+        restore_saved_state (window, is_first_window);
+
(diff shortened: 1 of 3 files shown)


src/Application.vala
```diff
@@ -296,6 +300,25 @@ public class Terminal.Application : Gtk.Application {
...
+            window.resize (
+                Application.saved_state.get_int ("window-width"),
+                Application.saved_state.get_int ("window-height")
+            );
+        }
+
+        if (is_first_window) {
+            window.size_allocate.connect ((alloc) => {
+                Application.saved_state.set_int ("window-width", alloc.width);
+                Application.saved_state.set_int ("window-height", alloc.height);
```
danirabbit, line 316: Gotta make sure to skip these if window is maximized otherwise you get weird sizes

danirabbit (changes requested):
One comment otherwise looks good

PR #784, opened 2024-09-05
Gtk4Prep - Do not subclass Gtk.Label

jeremypw:


danirabbit (approved):
Very straightforward :+1:

PR #785, opened 2024-09-05
SettingsPopover: use provider_for_screen for theme buttons

danirabbit:
GTK4 prep. Adding a provider to a widget is deprecated. Also, in Gtk4 adding a provider to a context doesn't add it to the children of the context, so this was broken

### DIFF
--- a/src/Widgets/SettingsPopover.vala
+++ b/src/Widgets/SettingsPopover.vala
@@ -166,7 +166,7 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
         };
 
         button.get_style_context ().add_class (Granite.STYLE_CLASS_COLOR_BUTTON);
-        button.get_style_context ().add_class (theme_to_css_class (theme));
+        button.get_style_context ().add_class (theme);
 
         css_provider = new Gtk.CssProvider ();
 
@@ -201,7 +201,7 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
         var foreground = theme_palette[Themes.PALETTE_SIZE - 2].to_string ();
 
         try {
-            css_provider.load_from_data (STYLE_CSS.printf (theme_to_css_class (theme), background, foreground));
+            css_provider.load_from_data (STYLE_CSS.printf (theme, background, foreground));
         } catch (Error e) {
             critical ("Unable to style color button: %s", e.message);
         }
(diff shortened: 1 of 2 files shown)


src/Widgets/SettingsPopover.vala
```diff
@@ -211,4 +211,18 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
...
         return true;
     }
+
+    // Valid css selectors must not start with a number or contain delimiters besides `-`
+    private static string theme_to_css_class (string theme) {
+        var theme_palette = Themes.get_rgba_palette (theme);
+        var background = theme_palette[Themes.PALETTE_SIZE - 3];
+        var foreground = theme_palette[Themes.PALETTE_SIZE - 2];
+
+        var css_class = "theme-%.0lf%.0lf%.0lf%.0lf-%.0lf%.0lf%.0lf%.0lf".printf (
```
jeremypw, line 221: Do we need so complicated css class?  See what I did in the Gtk4 branch - one css class for each theme name.  Colors will be overwritten if they change?

src/Widgets/SettingsPopover.vala
```diff
@@ -20,11 +20,9 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
     }
 
     private const string STYLE_CSS = """
-        .color-button radio {
+        .color-button.%s radio {
             background-color: %s;
             color: %s;
-            padding: 10px;
```
jeremypw, line 26: When I did the same in the Gtk4 branch, the buttons became small so is this necessary here?

danirabbit: I just moved this to the application.css since it never changes

src/Widgets/SettingsPopover.vala
danirabbit: Oh yeah theme name probably works fine. I definitely overthought it 😅

src/Widgets/SettingsPopover.vala
jeremypw: Oh I see - I missed that.

danirabbit:
Fixed! Dunno why I thought `theme` was an object and not a string here but this seems to work perfectly

jeremypw (approved):

PR #786, opened 2024-09-06
Gtk4 prep - Move zoom actions to MainWindow

jeremypw:
Having the zoom actions in TerminalWidget does not work in Gtk4 as it is not an ancestor of SettingsPopover which needs to use them.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -250,9 +250,9 @@ public class Terminal.Application : Gtk.Application {
         set_accels_for_action (TerminalWidget.ACTION_RELOAD, TerminalWidget.ACCELS_RELOAD);
         set_accels_for_action (TerminalWidget.ACTION_SCROLL_TO_COMMAND, TerminalWidget.ACCELS_SCROLL_TO_COMMAND);
         set_accels_for_action (TerminalWidget.ACTION_SELECT_ALL, TerminalWidget.ACCELS_SELECT_ALL);
-        set_accels_for_action (TerminalWidget.ACTION_ZOOM_DEFAULT, TerminalWidget.ACCELS_ZOOM_DEFAULT);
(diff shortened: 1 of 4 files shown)


Marukesu:
For Gtk4, the idea was to use `Gtk.Widget.add_binding_signal()` to handle the shortcuts and remove the action.

In the SettingsPopover, you can use a GLib.SignalGroup to connect the buttons to the `{decrease,increase}_font_size` signals in the Vte.Terminal.

jeremypw:
@Marukesu This is mainly to decrease the diff between the Gtk3 and Gtk4 versions by using code that works in both.  After the initial port then the Gtk4 version can be improved to use unique features of Gtk4.  However, it may not be worth having the intermediate step in this case?

jeremypw:
@danirabbit Do you recommend going straight to using `add_binding_signal` in the Gtk4 port or have this intermediate step?

danirabbit:
@jeremypw I'm not actually familiar with `add_binding_signal` or `GLib.SignalGroup` so I'm not sure I can give a meaningful opinion on what would be the easier path forward, sorry! So far what's in the main port branch isn't terribly difficult to review so either would probably be fine?

jeremypw:
@jeremypw Me neither but it looks like you can move a set of signal bindings to one instance of a class to another instance in one go. Could be useful.

danirabbit:
Since neither of us are familiar with the new method, let's move forward with this way and then we can revisit it in the future

danirabbit (changes requested):
For some reason accelerators aren't working correctly here. Keypad accels work, but Ctrl + and Ctrl - do not

jeremypw:
@danirabbit Just realized there is a much easier way to do this!  I forgot that SettingsPopover already has a reference to the current terminal via its terminal_binding_group so we can just call its functions directly. I'll close this and propose a better PR.

PR #787, opened 2024-09-08
Separate button controllers - Gtk4 prep

jeremypw:
Extracted from #780  as far as possible.

As the handlers do not share significant code it is clearer to have them separate.

We will not need to keep additional references to the controllers in Gtk4


danirabbit (approved):
Seems fine to me :+1:

PR #789, opened 2024-09-11
Gtk4 Prep - SettingsPopover - call terminal zoom functions directly

jeremypw:
In Gtk4, the terminal zoom actions will be lost (replaced with `add_binding_signal ()` to provide keyboard shortcuts for zooming).  We do not need to use them in SettingsPopover as we already have a reference to the current terminal.


danirabbit (approved):
That works

PR #794, opened 2024-10-13
Check for unsafe  drop 

### LINKED ISSUE #793: Drag-drop command bypasses sudo warning
### What Happened?

If a command containing `sudo` is pasted into a terminal a warning appears.  However this does not occur if the same text is dragged and dropped onto the terminal.

### Steps to Reproduce

1. Create a sudo command in text editor
2. Drag and drop it onto a terminal, including the crlf.
3. No warning appears - if the terminal already has admin privilege it will execute the command.

### Expected Behavior

The same warning should appear as for pasting the text.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #793 

In order to use the same code to check for unsafe input, both paste and drop end up feeding the text to the terminal child.  This actually saves some lines.


teamcons:
The issue was marked critical, but the PR seems out of date - Any ETA for merging ?

jeremypw:
I marked it critical because the issue has theoretical security implications.  I guess folks are focused on getting OS8 released atm.

ryonakano (dismissed):
The clipboard or dragged text are no longer pasted if we once uncheck "Show paste protection warnings" in the UnsafePasteDialog.

jeremypw:
@ryonakano Oops! Thanks for spotting that.  Now fixed.  It occurred to me that if unsafe paste alert is turned off in the dialog, there is no way of turning it back on through the UI.  Am I right?  If so that needs fixing but in another PR.

jeremypw:
I noticed that this PR was pasting the stripped text into the terminal whereas in master, the text was refetched from the clipboard. Now the stripped text is only used for the dialog message as before.

teamcons:
> @ryonakano Oops! Thanks for spotting that. Now fixed. It occurred to me that if unsafe paste alert is turned off in the dialog, there is no way of turning it back on through the UI. Am I right? If so that needs fixing but in another PR.

following your comment i did a PR for that.
Pantheon-tweaks always offered that option, im not sure why there was no toggle-back in first instance.

src/Widgets/TerminalWidget.vala
```diff
@@ -519,46 +519,41 @@ namespace Terminal {
...
+        private void validated_feed (string? text) {
+            if (text != null && text.validate ()) {
+                var stripped_text = text.strip ();
                 unowned var toplevel = (MainWindow) get_toplevel ();
+                if (!toplevel.unsafe_ignored &&
+                    Application.settings.get_boolean ("unsafe-paste-alert")) {
 
-                if (!toplevel.unsafe_ignored && Application.settings.get_boolean ("unsafe-paste-alert")) {
                     string? warn_text = null;
-                    text._strip ();
```
ryonakano, line 535: I think we should call `var stripped_text = text.strip ();` here instead of L529 for lesser scope.

src/Widgets/TerminalWidget.vala
```diff
@@ -519,46 +519,41 @@ namespace Terminal {
...
-                               remember_command_start_position ();
-                               base.paste_clipboard ();
+                                feed_child (text.data);
                             }
-
-                            dialog.destroy ();
                         });
 
                         dialog.present ();
                         return;
```
ryonakano, line 551: This return should be no longer necessary due to changes in this PR.

ryonakano (commented):
> The clipboard or dragged text are no longer pasted if we once uncheck "Show paste protection warnings" in the UnsafePasteDialog.

Confirmed this is fixed, thank you!

Leaving another two tiny comments.

src/Widgets/TerminalWidget.vala
jeremypw: Removed var altogether

src/Widgets/TerminalWidget.vala
jeremypw: Removed

ryonakano (approved):
Code looks good and fixes the linked issue. Thank you!

PR #795, opened 2024-10-16
Update cyclic search button when toggled

jeremypw:
Updates the icon name and tooltip text according to the state of the toggleaction.

This makes it easier to see whether the cyclic search is enabled or not.

### DIFF
--- a/src/Widgets/SearchToolbar.vala
+++ b/src/Widgets/SearchToolbar.vala
@@ -49,13 +49,23 @@ namespace Terminal.Widgets {
                 _("Next result")
             );
 
-            cycle_button = new Gtk.ToggleButton ();
-            cycle_button.image = new Gtk.Image.from_icon_name (
-                "media-playlist-repeat-symbolic", Gtk.IconSize.SMALL_TOOLBAR
-            );
-            cycle_button.sensitive = false;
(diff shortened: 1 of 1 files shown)


alainm23 (approved):

src/Widgets/SearchToolbar.vala
```diff
@@ -49,13 +49,23 @@ namespace Terminal.Widgets {
...
-            cycle_button.set_can_focus (false);
-            cycle_button.tooltip_text = _("Cyclic search");
+            cycle_button = new Gtk.ToggleButton () {
+                active = false,
+                can_focus = false,
+                image = new Gtk.Image ()
+            };
+            cycle_button.toggled.connect (() => {
+                if (cycle_button.active) {
+                    cycle_button.tooltip_text = _("Cyclic search is enabled");
```
danirabbit, line 59: Tooltip text for image buttons should describe the action that will happen when the button is pressed

danirabbit (changes requested):

src/Widgets/SearchToolbar.vala
jeremypw: Done

danirabbit (approved):
Let's go :rocket:

PR #802, opened 2024-11-08
TerminalWidget: Fix non administrative commands not being pasted

ryonakano:
Fix a regression from #794 that non administrative commands like `apt list --installed` no longer pasted.


jeremypw:
#804 Should fix this?

ryonakano:
@jeremypw Yes, but I want to get merged this PR first to separate bug fix and refactoring. #804 is just based on this branch to avoid conflict when this PR get merged into master.

jeremypw (approved):
That fixed it.

PR #803, opened 2024-11-08
Drop dead feature that disables unsafe paste warning while Terminal opens

ryonakano:
This PR removes the dead feature that disables unsafe paste warning while Terminal opens.

---

As far as reading the code of UnsafePasteDialog.vala, this dialog shouldn't appear again while the current Terminal window is alive once if the user clicks "Paste Anyway". However, this isn't working at the moment because the clicked event of the "Paste Anyway" button is treated as a response of UnsafePasteDialog and the `on_ignore ()` method is not triggered at all.

On the other hand, however, I also feels the current "unintentional" behavior that this dialog always appears when pasting administrative commands makes sense too. We can now toggle unsafe paste warning from preferences menu,
so this dead feature is useless. Therefore, I just dropped the dead feature instead of correctly implementing it in favor of current "unintentional" behavior.


jeremypw:
I think now we have the toggle in the App menu to control the unsafe warning we do not need this additional check in the dialog.

ryonakano:
@jeremypw Thank you for your opinion. I agree with you and perform changes to drop this dead feature.

jeremypw (approved):
Does what it says on the tin.  

I am not sure we even need the CheckButton either but that can be left for further consideration.

PR #804, opened 2024-11-08
TerminalWidget: Refactor validated_feed to avoid nested conditions

ryonakano:
The current code of `validated_feed ()` is difficult to recognize routes for each condition, so make it more readable by avoiding nested conditions.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -525,35 +525,44 @@ namespace Terminal {
 
         // Check pasted and dropped text before feeding to child;
         private void validated_feed (string? text) {
-            if (text != null && text.validate ()) {
-                unowned var toplevel = (MainWindow) get_toplevel ();
-                if (!toplevel.unsafe_ignored &&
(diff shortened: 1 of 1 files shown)


src/Widgets/TerminalWidget.vala
```diff
@@ -525,33 +525,44 @@ namespace Terminal {
...
+
+            string? warn_text = null;
+            if ("\n" in text) {
+                warn_text = _("The pasted text may contain multiple commands");
+            } else if ("sudo" in text || "doas" in text) {
+                warn_text = _("The pasted text may be trying to gain administrative access");
+            }
+
+            // No user interaction for safe commands
+            if (warn_text != null) {
```
jeremypw, line 551: Should be `warn_text == null` ?

ryonakano: Ah silly mistake…… :dizzy_face: I'll fix after #802 and #803 is get merged that should conflict with this PR. Thank you for pointing out this.

jeremypw (approved):
Code looks good and works as expected, thanks.

PR #809, opened 2024-11-21
TerminalWidget: Use the non-deprecated spawn API

tintou:


jeremypw:
If run on OS7.1 this seems to cause the whole session to crash when the terminal window is closed.  It runs OK on OS8 though.

jeremypw:
The problem is caused by an incorrect signature for `TerminalSpawnAsyncCallback` in the version of vte-2.91.vapi in OS7.1.  If a corrected version is put in the the `vapi` folder of the source code then the issue is fixed.

src/Widgets/TerminalWidget.vala
```diff
@@ -665,29 +665,58 @@ namespace Terminal {
...
+                (terminal, pid, error) => {
+                    if (pid != -1) {
+                        this.child_pid = pid;
+                    } else {
+                        warning (error.message);
+                    }
+                }
+            );
 
             check_cwd_changed ();
```
jeremypw, line 686: Shouldn't this be inside the callback?

src/Widgets/TerminalWidget.vala
```diff
@@ -665,29 +665,58 @@ namespace Terminal {
...
+                Vte.PtyFlags.DEFAULT,
+                dir,
+                { shell },
+                envv,
+                SpawnFlags.SEARCH_PATH,
+                null,
+                -1,
+                null,
+                (terminal, pid, error) => {
+                    if (pid != -1) {
```
jeremypw, line 678: Better to check `error == null` since the else clause dereferences error?

PR #813, opened 2024-11-29
Application: withdraw stale notifications

### LINKED ISSUE #812: Notifications are not withdrawn
### What Happened?

I have old notifications from Terminal - Process ended with errors, process ended
Most were addressed (terminal focused)
A lot of them are from one or two sessions ago. So long obsolete

### Steps to Reproduce

receive notifications from terminal
never withdrawn

### Expected Behavior

terminal should withdraw addressed notifications

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

danirabbit:
Fixes #812

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -204,17 +204,11 @@ public class Terminal.Application : Gtk.Application {
                 send_notification ("process-finished-%s".printf (id), notification);
 
                 terminal.main_window.notify["current-terminal"].connect (() => {
-                    if (terminal.main_window.current_terminal == terminal) {
-                        terminal.tab.icon = null;
-                        withdraw_notification ("process-finished-%s".printf (id));
-                    }
+                    withdraw_notification_for_terminal (terminal, id);
                 });
 
                 terminal.main_window.focus_in_event.connect (() => {
-                    if (terminal.main_window.current_terminal == terminal) {
-                        terminal.tab.icon = null;
-                        withdraw_notification ("process-finished-%s".printf (id));
-                    }
+                    withdraw_notification_for_terminal (terminal, id);
 
                     return Gdk.EVENT_PROPAGATE;
                 });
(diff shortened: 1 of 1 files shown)


lenemter:
@danirabbit we don't support withdrawing notifications yet, do we? https://github.com/elementary/wingpanel-indicator-notifications/issues/204

danirabbit:
@lenemter [we do](https://github.com/elementary/wingpanel-indicator-notifications/commit/3d12554509ea25db9de756ead9ab0a63b7991cbf), but it doesn't work across sessions yet. So if you restart in between it won't get cleared because the notification IDs I think aren't restored

jeremypw:
Shouldn't we remove the "process-completed" tab icon as well when tab is focussed?

jeremypw:
@danirabbit  Works within same session.  Not sure if you want to fix the icon issue, which is kind of related, in this PR or later.  Do you want to wait cross-session ability too?

src/Application.vala
```diff
@@ -202,12 +202,38 @@ public class Terminal.Application : Gtk.Application {
...
+    private void withdraw_notification_for_terminal (TerminalWidget terminal, string id, ulong tab_change_handler, ulong focus_in_handler) {
+        if (terminal.main_window.current_terminal != terminal) {
+            return;
+        }
+
+        terminal.tab.icon = null;
+        withdraw_notification ("process-finished-%s".printf (id));
+
+        terminal.main_window.disconnect (tab_change_handler);
+        terminal.main_window.disconnect (focus_in_handler);
```
jeremypw, line 234: Adding 
```
        if (terminal == terminal.main_window.current_terminal) {
            terminal.tab.icon = null;
        }
```
fixes the icon issue.

danirabbit: I already have that on line #230?

jeremypw: Sorry, I must have got confused about which branch I was running at some point :disappointed: Your PR does indeed fix that issue!

jeremypw (approved):
I am going to approve this as within session functionality is still an improvement

PR #816, opened 2024-12-06
Add tab to new window created with <Shift><Ctrl>N

### LINKED ISSUE #814: New window created with `<Ctrl><Shift>N' has zero tabs and is transparent
### What Happened?

Pressed `<Ctrl><Shift>N`.  A (small) new window appeared with the background showing through where the tab should be.

### Steps to Reproduce

1. Open Terminal
2. Press '<Ctrl><Shift>N`

### Expected Behavior

The new window should contain one default tab

### OS Version

7.x (Horus)

### Session Type

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #814 

 - New tab added in Application `new-window` action handler
 - New window follows size of current window
 - New tab follows current working directory of current window

There is scope for improving consistency and DRYing with code handling new windows created by dragging tabs, from the dock and from the commandline but that is left for now.

Code potentially changing the window size by adding a tab was removed as it did not seem appropriate. The window size is already controlled by settings, the user and the window manager.

### DIFF
--- a/data/io.elementary.terminal.appdata.xml.in
+++ b/data/io.elementary.terminal.appdata.xml.in
@@ -73,6 +73,7 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/terminal/issues/814">New window created with Ctrl Shift N has zero tabs and is transparent</issue>
         <issue url="https://github.com/elementary/terminal/issues/812">Notifications are not withdrawn</issue>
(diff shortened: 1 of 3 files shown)


danirabbit (approved):

danirabbit:
Nice fix!

PR #822, opened 2024-12-14
Revert "Gtk4Prep - Handle restoring and syncing window state in App

### LINKED ISSUE #821: The new terminal (v6.3.0) is opening with ever increasing geometry
### What Happened?

The elementary terminal application (v6.2.0) used to open with the fixed size in the middle of the screen. The new terminal released two days ago (v6.3.0) keeps opening with ever increasing window size which amounts to a pretty poor user experience.

### Steps to Reproduce

1. Open the elementary terminal from the application launcher
2. Close it either by clicking the close button or pressing Ctrl-D
3. Open it again, watch it open with a slightly larger size
4. Keep repeating steps 2 and 3 and eventually a new terminal window occupies the entire screen like a Christmas monster

### Expected Behavior

It was fine the way it was before. Be nice if it remembered how big it was but it doesn't have to keep increasing in size with every summon.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

```shell
Using `G_MESSAGES_DEBUG=all io.elementary.terminal -n` did not produce any logs :-(
```

### Hardware Info

_No response_

jeremypw:
This reverts commit 32ff88ecea950d172e65c8cfd7cb86b22800712f.

Fixes #821

No elegant fix was found that worked with Gtk3/X  on OS7.1 so the proposed changes in this commit will have to wait until the Gtk4 port I guess.  The draft Gtk4 port in preparation does not suffer from this issue.

For simplicity, the new window action now creates a window the same size as the window currently in use rather getting from the settings.

### DIFF
--- a/data/io.elementary.terminal.gschema.xml
+++ b/data/io.elementary.terminal.gschema.xml
@@ -17,20 +17,15 @@
   </enum>
 
   <schema path="/io/elementary/terminal/saved-state/" id="io.elementary.terminal.saved-state">
-    <key name="window-height" type="i">
-      <default>700</default>
-      <summary>Most recent window height</summary>
-      <description>Most recent window height</description>
+    <key name="window-size" type="(ii)">
+      <default>(-1, -1)</default>
+      <summary>Most recent window size</summary>
+      <description>Most recent window size (width, height)</description>
     </key>
-    <key name="window-width" type="i">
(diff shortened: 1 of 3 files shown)


jeremypw:
Investigating CI failure.

alainm23 (approved):

PR #823, opened 2024-12-16
Fix clear and reset actions when ongoing foreground process

### LINKED ISSUE #819: Reset action uses Ctrl+K as hotkey
### What Happened?

I am a very heavy [nano](https://nano-editor.org/) user (yeah...) and when testing [6.3.0](https://github.com/elementary/terminal/releases/tag/6.3.0) I found myself when doing <kbd>Ctrl</kbd> + <kbd>K</kbd> in nano it no longer cuts the line but instead reset the session (<kbd>Ctrl</kbd> + <kbd>K</kbd> is the hotkey in nano to cut a line), I believe this is introduced in https://github.com/elementary/terminal/pull/778.

It will be nice if there is a way to opt-out that (use another key, or remove the hotkey, or make it configurable, or anything...), though for now I guess i will just patch my own installation, assuming it is hardcoded... :joy:

https://github.com/elementary/terminal/blob/b113a966e5f9b506576224645e566ec32a98a79f/src/Widgets/TerminalWidget.vala#L78

https://github.com/elementary/terminal/blob/b113a966e5f9b506576224645e566ec32a98a79f/src/Widgets/TerminalWidget.vala#L450-L453

### Steps to Reproduce

1. Open terminal and run `nano some_file`
2. Type random stuff
3. <kbd>Ctrl</kbd> + <kbd>K</kbd>

### Expected Behavior

1. The hotkey is handled by nano instead of the terminal :joy:

### OS Version

NixOS

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

### LINKED ISSUE #820: The new terminal (v6.3.0) is destroying current process on Ctrl-L
### What Happened?

The elementary terminal (v6.2.0) used to work fine and pressing Ctrl-L in it would clear the screen as expected of a terminal. The new terminal released two days ago (v6.3.0) still clears the screen but kills the current process in doing so. So if I'm logged in with SSH to a remote machine, pressing Ctrl-L kill the SSH session and drops me in the / path on my local system. 

### Steps to Reproduce

1. Open elementary terminal
2. Type `ls` or any command. Heck, even switch to root with `sudo su -`. Or SSH into some remote machine
3. Press Ctrl-L, watch the session die, watch the screen clear, watch the current working directory switch to `/`

### Expected Behavior

Pressing Ctrl-L should clear the screen only and it should NOT kill the current process or change the current working directory.

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

```shell
Using `G_MESSAGES_DEBUG=all io.elementary.terminal -n` does not produce any logs :-(
```

### Hardware Info

_No response_

jeremypw:
Fixes #819 
Fixes #820 

 - [x] Confirm kill foreground process before reset action
 - [x] Ignore clear screen action if there is a foreground process 
 - [x] Change hotkeys for clear and reset actions so do not clash with `nano`
 - [x] DRY related code  

Because we cannot guarantee the state of the terminal if the foreground process is killed (see #824 ) the clear action is ignored when a foreground process is running. Resetting the terminal is OK as the terminal state is reset to something sensible.

There is little difference between the "Reload" action and the "Reset" action. The former starts a new shell process whereas the latter resets the existing one.  Not sure if we need both. Only the latter is shown in the context menu atm.

### DIFF
--- a/src/Dialogs/ForegroundProcessDialog.vala
+++ b/src/Dialogs/ForegroundProcessDialog.vala
@@ -20,42 +20,21 @@
 public class Terminal.ForegroundProcessDialog : Granite.MessageDialog {
     public string button_label { get; construct; }
 
-    public ForegroundProcessDialog (MainWindow parent) {
+    public ForegroundProcessDialog (
+        MainWindow parent,
+        string primary_text,
+        string button_label
+    ) {
         Object (
             transient_for: parent,
-            primary_text: _("Are you sure you want to close this tab?"),
-            secondary_text:
-                _("There is an active process on this tab.") + " " +
-                _("If you close it, the process will end."),
-            buttons: Gtk.ButtonsType.CANCEL,
-            button_label: _("Close Tab")
-        );
-    }
-
-    public ForegroundProcessDialog.before_close (MainWindow parent) {
-        Object (
-            transient_for: parent,
-            primary_text: _("Are you sure you want to quit Terminal?"),
-            secondary_text:
-                _("There is an active process on this terminal.") + " " +
-                _("If you quit Terminal, the process will end."),
-            buttons: Gtk.ButtonsType.CANCEL,
-            button_label: _("Quit Terminal")
-        );
-    }
-
-    public ForegroundProcessDialog.before_tab_reload (MainWindow parent) {
-        Object (
-            transient_for: parent,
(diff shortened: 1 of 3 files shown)


stsdc:
Works for me.

stsdc (approved):

PR #829, opened 2024-12-19
Apply minimum dimensions on restore

### LINKED ISSUE #762: Small window on first run under Wayland
### What Happened?

The window is really small if the app is first run under wayland

### Steps to Reproduce

1. Have a clean install of OS 8
2. Start the wayland session instead of X11
3. open terminal
4. See screenshot:

Screenshot from 2024-06-28 13 43 07@2x

### Expected Behavior

My guess is that there's some code here that's trying to get the size of the display and we don't have access to that under Wayland maybe? So there probably needs to be a better fallback

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #762 

Simple fix using existing minimum dimension constants - does not consider font size or scaling.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -508,6 +508,9 @@ namespace Terminal {
                 default_height = geometry.height * 3 / 4;
             }
 
+            default_width = int.max (Application.MINIMUM_WIDTH, default_width);
+            default_height = int.max (Application.MINIMUM_HEIGHT, default_height);
+
             var window_state = Terminal.Application.saved_state.get_enum ("window-state");
             if (window_state == MainWindow.MAXIMIZED) {
                 maximize ();


leolost2605:
I think a better fix would be to use the get_monitor_at_window instead of the primary one which is
- supported on Wayland
- usually the primary one anyways on first launch
- and if not it makes more sense to adjust the size according to the monitor it's actually on and not to the primary one

alainm23 (dismissed):

jeremypw:
@leolost2605  I'll try it. I was assuming it was the `get_geometry ()` that was the problem tho` - might be as well to have an ultimate hard-coded fallback still.

jeremypw:
The problem with using `get_monitor_at_window ()`  and `get_primary_monitor ()` is that it does not work unless the window is realized (which it isn't at the moment).  However if you restore the window size after realizing then setting the default dimensions does not work. In Gtk3 you can use `resize ()` but that does not exist in Gtk4 so I do not want to introduce it here.  I think therefore we will have to abandon the idea of resizing the terminal window according to the monitor size as a fallback

leolost2605:
> The problem with using get_monitor_at_window () and get_primary_monitor () is that it does not work unless the window is realized (which it isn't at the moment). However if you restore the window size after realizing then setting the default dimensions does not work. In Gtk3 you can use resize () but that does not exist in Gtk4 so I do not want to introduce it here. I think therefore we will have to abandon the idea of resizing the terminal window according to the monitor size as a fallback

Where in GTK3 you can do resize (), in GTK4 setting default size after realizing will work so it can be done :)

(I don't have an opinion here just want to say that it's possible :))

jeremypw:
For no obvious reason the zoom overlay appears (blank) on startup in this branch whereas it does not in `master` - investigating.

jeremypw:
Zoom_overlay problem due to a change in timing/order of events. Now fixed.

zeebok:
When I tested on OS8/X I now have a small terminal window

Issue #832, opened 2024-12-22
Setting to hide tab title when there is only one tab no more work


BorisNA:
### What Happened?

After the latest update (version 6.3.0+r2614+pkg89~ubuntu7.1) terminal appears to ignore old gsetting

`gsettings set io.elementary.terminal.settings tab-bar-behavior 'Hide When Single Tab'`

'gsettings' still lists this option as a valid one, but the terminal opens with the tab title shown, even for the one tab.

'Never Show Tabs' does not work either

### Steps to Reproduce

1. Updated to 6.3.0+r2614+pkg89~ubuntu7.1
2. Opened a terminal
3. Re-checked correct gsettings value (Hide When Single Tab)
4. Tab's header is there 

![Screenshot from 2024-12-23 00 13 14](https://github.com/user-attachments/assets/03d161d1-d078-47b5-bfac-a5e02bc43c36)

### Expected Behavior

Tab headers should be hidden

In fact, they take too much space on old small displays (think about old x220 and like).

### OS Version

7.x (Horus)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

ryonakano:
Seems like this feature was removed in #760 and the gsettings key is just a leftover

ryonakano:
I think we can re-implement (at least) "Hide When Single Tab" by using [Hdy.TabBar.autohide](https://valadoc.org/libhandy-1/Hdy.TabBar.autohide.html) which is fixed to false in [the current code](https://github.com/elementary/terminal/blob/01b68dc6701f26611c43041c46bc779b6f32042e/src/Widgets/TerminalView.vala#L75), but I'm not sure if developers interested in re-implementing this because this feature has been a hidden settings which means most users couldn't and wouldn't configure it unless they know gsettings command or unofficial tools like Pantheon Tweaks.

jeremypw:
Its the designer's call but I do not think the benefit of these settings (if any) is worth the added visual noise.

BorisNA:
There would be no 'visual noise' becasue these settings are controlled through gsettings, that is command line, without confusing other users.

As for tabs - yes, it's designers call, I never understood rationale for tabs in terminal emulator in window DE (unlike screen or tmux in real terminals, but it a completely different story.

BorisNA:
Proposed PR should completely bring 6.1.2/6.2.0 behavior back if maintainers approve it.

[...]

PR #833, opened 2024-12-23
Bring back autohide feature from #832

BorisNA:
Bring back autohide feature from #832

Terminal v 6.2.0 had an option to hide tab bar completely or when there is only one tab is present. All the options could not be re-implemented, but we can re-intorduce hide-when-single

- keep old  gschema values for existing user settings and guides compatibility
- local changes and borrow the enum from Granite

### DIFF
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -10,6 +10,12 @@ public class Terminal.TerminalView : Gtk.Box {
         URI_LIST
     }
 
+    public enum TabBarBehavior {
+        ALWAYS = 0,
+        SINGLE = 1,
+        NEVER = 2
+    }
+
     public signal void new_tab_requested ();
     public signal void tab_duplicated (Hdy.TabPage page);
 
@@ -71,8 +77,9 @@ public class Terminal.TerminalView : Gtk.Box {
             use_popover = false,
         };
 
+        var tab_bar_behavior = (TabBarBehavior)Application.settings.get_enum ("tab-bar-behavior");
         tab_bar = new Hdy.TabBar () {
-            autohide = false,
+            autohide = (tab_bar_behavior == TabBarBehavior.SINGLE), // This is <value value="1" nick="Hide When Single Tab"/>
             expand_tabs = false,
             inverted = true,
             start_action_widget = new_tab_button,


BorisNA:
I have tested it from the "horus" branch on my 7.1 installation - appears to work good.

Hope it will be added to the next update for 7.1

ryonakano:
Does this work for "Always", which means the tabs should be always invisible even when there are more than two tabs?

ryonakano:
What about using the visible property inherited from Gtk.Widget?

https://valadoc.org/gtk+-3.0/Gtk.Widget.visible.html

jeremypw:
I am not against bringing this option back as there was, as far as I recall, any conscious decision to remove it. However, if it is worth having I would like to see it exposed in the UI.  Needs input from the @danirabbit.

BorisNA:
> What about using the visible property inherited from Gtk.Widget?

This may work, thanks (I have almost no prior experience with Gtk or UI widgets at all). I will definitely look into it

What really bothers me is that option "always hide" in this implementation will be mega-confusing. Say, I have enabled always hiding the tab bar and later spawned a tab with command-line key or shortcut. Would that mean that I'll get an invisible tab in the background, or the current tab I'm working with will magically disappear "behind" a new one? And no way to switch or even know that there is something there.

Looks like the logic for "always hide" should be something like "hide the tab bar _and_ substitute opening new tab with opening new window".


> I am not against bringing this option back

I'm truly sorry if I misunderstood you and my reply was harsh.

When I work on the big monitor, Ok, I can live with a single tab in the tab bar, it just doesn't worth arguing. But for smaller screens it really hurts, I cannot have for example two terminals side by side (vertically) and so on :(  And it worked before, on the same OS version.

jeremypw:
> What really bothers me is that option "always hide" in this implementation will be mega-confusing. 

I agree.  I can understand hiding the tabbar when there is one tab if you need the screen space, but if there are multiple tabs you need some way of knowing where you are.  One possibility is some kind of tab-switcher in the headerbar but there would be no space for names.

BorisNA:
> One possibility is some kind of tab-switcher in the headerbar

Drop-down tab switcher with some custom button in the header bar indeed is a possibility, but this will introduce nonstandard behavior. I believe it will confuse users too and be against main elementary idea (as I get it - simplicity and consistence).

Maybe we really can change the 'new tab' behavior with the 'new window'? That is, all accelerators, command line options and so on will just spawn a new window? Each window will have just one tab and have the tabbar hidden.

It wil lrequire some painstaking code fixes, since new tab is spawn in different places here and there, but at least will not introduce new concepts for user.

jeremypw:
> I believe it will confuse users too and be against main elementary idea (as I get it - simplicity and consistence).

The simplest and most consistent way is to always show tabs :wink:

BorisNA:
Ehrm, I'd disagree. The simplest way is to not use tabs at all and leave window switching for whom it belongs - that is window manager.

BorisNA:
I tested the old version (6.1.2/6.2.0) behavior: it really didn't show any bar when the "always hide" option is used, but the user could change tabs with the keyboard shortcuts.

So, I replicated this behavior by not adding `tab_bar` widget to the parent box - it appears to work in the same way: you can spawn a new tab, change and close tabs. Unfortunately, this couldn't be done with `hide()` or `set_visible` - may be Hdy.TabBox uses is for its `autohide` feature.

For now, it works 100% as the previous version and I propose to add it to the codebase as a regression fix.

I'm open to discuss new logic I proposed (switching tabs behavior to window behavior for this "Never Show" setting) if the maintainers will approve it.

jeremypw:
I can confirm that this works as expected from the code.  If someone can confirm this reproduces exactly the behaviour of v6.2.0 and before, I will not block merging even though I dislike many aspects of this behaviour, particularly "Never show tabs" e.g. the options shown in the tab context menu are lost and changes to the settings are not reflected in the UI unless the app is closed and reopened.  Also need go-ahead from @danirabbit .

src/Widgets/TerminalView.vala
```diff
@@ -96,7 +103,11 @@ public class Terminal.TerminalView : Gtk.Box {
         tab_bar.extra_drag_dest_targets = new Gtk.TargetList ({uris});
         tab_bar.extra_drag_data_received.connect (on_extra_drag_data_received);
 
-        add (tab_bar);
+        if( tab_bar_behavior != TabBarBehavior.NEVER ) {
+            // Looks like we can't use set_visible since its value should be overriden in TabBar
+            // so the only solution is not to add the widget to the box itself
+            add (tab_bar);
+        }
```
ryonakano, line 110: I think we can use Gtk.Revealer instead of changing the process here.

@BorisNA I've created a PR against your forked repository as a suggested changes, I'd like you to take a look at https://github.com/BorisNA/elementary-terminal/pull/1.

BorisNA: Thanks, it looks great. Works as expected (I even didn't know that such a widget exists)
Accepted it, waiting for maintainers.

danirabbit:
I'm not staunchly opposed to an option that only shows the tab bar when they are multiple tabs, but a never option seems kind of broken. If you don't ever want to see the tab bar, then it seems like that would already be covered by just not opening new tabs. And in the case that a new tab is somehow opened automatically, you need to be able to get to it and could just close it.

Perhaps we could revisit a tabbarless option in GTK4 when we have the tab overview available so that tab switching is still accessible. But I'm personally -1 on shipping an option that is a footgun

jeremypw:
Marking as blocked until the "Never" option is removed.

PR #836, opened 2024-12-31
Cycle tabs with Ctrl+Tab

### LINKED ISSUE #834: Tab cycling is not cyclical
### What Happened?

Tab cycling is not cyclical anymore

### Steps to Reproduce

1. open terminal
2. create new tab
3. cycle tab
4. tab does not cycle to the first one

### Expected Behavior

tab does cycle to the first one

### OS Version

8.x (Circe)

### Session Type

### Software Version

### Log Output

```shell
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:44.450: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:45.542: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:45.542: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:45.542: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:46.073: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:46.073: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:46.073: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:47.080: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:47.080: change_fast
(io.elementary.terminal:338911): dconf-DEBUG: 21:31:47.080: change_fast
```

### Hardware Info

Quad-Core 11th Gen Intel® Core(TM) i7-1165G7 @ 2.80GHz
Intel® TigerLake-LP GT2 [Iris Xe Graphics]

jeremypw:
Fixes #834 

Rather than use the native Hdy.TabView select_next_page ()  and select_previous_page () functions, which do not cycle, a custom function is used.

### DIFF
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -13,7 +13,7 @@ public class Terminal.TerminalView : Gtk.Box {
     public signal void new_tab_requested ();
     public signal void tab_duplicated (Hdy.TabPage page);
 
-    public uint n_pages {
+    public int n_pages {
         get {
             return tab_view.n_pages;
         }
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -165,6 +165,18 @@ public class Terminal.TerminalView : Gtk.Box {
         tab_view.selected_page = target;
     }
 
+    public void cycle_tabs (Hdy.NavigationDirection direction) {
+        var pos = (int) tab_view.get_page_position (selected_page);
+        pos = direction == FORWARD ? pos + 1 : pos - 1;
+        if (pos == n_pages) {
+            pos = 0;
+        } else if (pos < 0) {
+            pos = (int) n_pages - 1;
+        }
+
+        selected_page = tab_view.get_nth_page (pos);
+    }
+
     public void transfer_tab_to_new_window () {
         var target = tab_menu_target ?? tab_view.selected_page;
 
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -166,7 +166,7 @@ public class Terminal.TerminalView : Gtk.Box {
     }
 
     public void cycle_tabs (Hdy.NavigationDirection direction) {
-        var pos = (int) tab_view.get_page_position (selected_page);
+        var pos = tab_view.get_page_position (selected_page);
         pos = direction == FORWARD ? pos + 1 : pos - 1;
         if (pos == n_pages) {
             pos = 0;
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -168,11 +168,7 @@ public class Terminal.TerminalView : Gtk.Box {
     public void cycle_tabs (Hdy.NavigationDirection direction) {
         var pos = tab_view.get_page_position (selected_page);
         pos = direction == FORWARD ? pos + 1 : pos - 1;
-        if (pos == n_pages) {
-            pos = 0;
-        } else if (pos < 0) {
-            pos = (int) n_pages - 1;
-        }
+        pos = (pos + n_pages) % n_pages;
 
         selected_page = tab_view.get_nth_page (pos);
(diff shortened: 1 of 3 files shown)


src/Widgets/TerminalView.vala
```diff
@@ -165,6 +165,18 @@ public class Terminal.TerminalView : Gtk.Box {
         tab_view.selected_page = target;
     }
 
+    public void cycle_tabs (Hdy.NavigationDirection direction) {
+        var pos = (int) tab_view.get_page_position (selected_page);
```
ryonakano, line 169: Is there any reason you explicitly cast to int here? The return type of `Hdy.TabView.get_page_position ()` seems to be int: https://valadoc.org/libhandy-1/Hdy.TabView.get_page_position.html

src/Widgets/TerminalView.vala
```diff
@@ -165,6 +165,18 @@ public class Terminal.TerminalView : Gtk.Box {
...
+        pos = direction == FORWARD ? pos + 1 : pos - 1;
+        if (pos == n_pages) {
+            pos = 0;
+        } else if (pos < 0) {
+            pos = (int) n_pages - 1;
+        }
```
ryonakano, line 175: Maybe we can calculate this instead of conditions like this:

```vala
pos %= n_pages;
```

src/Widgets/TerminalView.vala
jeremypw: You're right - its not needed. Removed

src/Widgets/TerminalView.vala
jeremypw: Nice idea but it doesn't work for negative values of `pos` (-1 % n_pages = -1).  You can use `pos = (pos + (int) n_pages ) % ((int) n_pages)`   but at least the current solution is clear what is going on.

ryonakano: > You can use `pos = (pos + (int) n_pages ) % ((int) n_pages)`

Sorry, this is what I wanted to suggest. :bow: Plus `n_pages` is actually typeof int because its value come from `Hdy.TabView.n_pages`, so we can make it simpler by removing casts:

```suggestion
        pos = (pos + tab_view.n_pages) % tab_view.n_pages;
```

> but at least the current solution is clear what is going on.

Yes, I'm also a fan of readable codes. However, I don't think choosing calculation instead of the condition here decreases readability of the code because this formula is simple. Also, I think it's often used elsewhere e.g. when calculating next index in ring buffers. Conditions sometimes causes bugs so I think we should avoid it when it's possible and avoiding it doesn't lose readability.

jeremypw: OK, I am persuaded. I noticed that the TerminalView.n_pages does not need to be a `uint` anyway. I presume there was some reason in the past that was done but making it an `int` does not cause any problems and removes the need for several other casts.

ryonakano (commented):
If we remove the unnecessary cast against `n_pages` here, we can remove the following places too ~~, excepting the cast against the result of `focus.clamp ()`~~ :

```
[ryo@b760m ~/work/tmp/terminal (jeremypw/cycle-tabs =)]$ git grep -n n_pages | grep -F '(int)'
src/MainWindow.vala:485:                        notebook.selected_page = notebook.tab_view.get_nth_page ((int)notebook.n_pages - 1);
src/MainWindow.vala:718:                var tab = notebook.tab_view.get_nth_page ((int)(focus.clamp (0, notebook.n_pages - 1)));
src/MainWindow.vala:727:            int pos = (int)notebook.n_pages
[ryo@b760m ~/work/tmp/terminal (jeremypw/cycle-tabs =)]$
```

jeremypw:
Nice use of `git grep` - I was not familiar with that.

ryonakano (approved):
LGTM and works as expected. Thank you for your fix!

PR #841, opened 2025-01-03
Center the title, set min searchentry width

### LINKED ISSUE #840: Title in titlebar not centered
### What Happened?

If it's a bug it's for sure minimal, but the title was nicely centered in previous versions of the application, including the picture used in elementary.io and now it's not. Since it's still centered in the other applications that have similar titlebars, it stands out as inconsistent. It seems to have become centered in relation to the space between the buttons, instead of centered in the titlebar.
Captura de tela de 2025-01-03 05 40 37

### Steps to Reproduce

1. Open the Terminal application

### Expected Behavior

Centered title

### OS Version

7.x (Horus)

### Session Type

### Software Version

### Log Output

_No response_

### Hardware Info

_No response_

jeremypw:
Fixes #840, with side-effects

The cause of the issue is that the custom title is set to horizontally expand so the headerbar cannot center it on itself.

There is no simple fix to this issue while keeping the same appearance of the search widget as far as I can find. As a compromise, the search widget now has a fixed width which should be sufficient for most search terms and is centered in the headerbar.

BEFORE
Screenshot from 2025-01-03 12 14 33

Screenshot from 2025-01-03 12 14 42

AFTER
Screenshot from 2025-01-03 12 13 54

Screenshot from 2025-01-03 12 14 04

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -332,8 +332,12 @@ namespace Terminal {
             title_label.get_style_context ().add_class (Gtk.STYLE_CLASS_TITLE);
 
             title_stack = new Gtk.Stack () {
-                transition_type = Gtk.StackTransitionType.SLIDE_UP_DOWN
+                transition_type = Gtk.StackTransitionType.SLIDE_UP_DOWN,
+                hhomogeneous = false,
+                hexpand = false
             };
+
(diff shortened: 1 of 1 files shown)


zeebok (approved):
300 might be a little small but is a good enough starting point to fix the issue

PR #847, opened 2025-01-06
Remove tab process-finished icon on select

### LINKED ISSUE #831: Process exit status icon remains forever
### What Happened?

After running a process in the terminal, the green/red icon that represents the exit status of the process sometimes remains even after you activate the tab. I think it's limited to long-running processes, and since I always have more than one tab open it might be limited to such use-cases.

### Steps to Reproduce

1. Start a long-running process, such as a gcc compilation job
2. Activate another window (or maybe another terminal tab), and let the process finish
3. Once the process has finished, a green/red icon (depending on the exit status of the process) will show up next to the directory name in the tab
4. Activate the tab to look at the process output
5. The icon remains

### Expected Behavior

The icon goes away once the tab has been activated - at least that's how it was in 7.x.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

_No response_

### Hardware Info

8 × AMD Ryzen 7 4800U with Radeon Graphics 
AMD® Renoir [Radeon RX Vega 6 (Ryzen 4000/5000 Mobile Series)]

jeremypw:
Fixes #831

The "process-finished" icon is only added if the tabpage is not selected when the process finishes so we only need remove it when the tabpage is selected.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -14,6 +14,8 @@ public class Terminal.Application : Gtk.Application {
     public static GLib.Settings settings;
     public static GLib.Settings settings_sys;
 
+    public const string PROCESS_COMPLETED_ICON_NAME = "process-completed-symbolic";
+
     public bool is_testing { get; set construct; }
 
     private static Themes themes;
@@ -186,7 +188,7 @@ public class Terminal.Application : Gtk.Application {
             }
 
             var process_string = _("Process completed");
-            var process_icon = new ThemedIcon ("process-completed-symbolic");
+            var process_icon = new ThemedIcon (PROCESS_COMPLETED_ICON_NAME);
             if (exit_status != 0) {
                 process_string = _("Process exited with errors");
                 process_icon = new ThemedIcon ("process-error-symbolic");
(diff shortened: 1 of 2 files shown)


jeremypw:
Not obvious why CI fails for this in unstable whereas `master` passes but it seems to be related to #893

jeremypw:
@lenemter Hmm, you are right - that procedure produces different behaviour to the one I used.  I'll investigate.

src/Application.vala
lenemter, line 1: I don't think we should introduce constants here - they don't improve clarity and are only used once

src/MainWindow.vala
```diff
@@ -394,9 +394,14 @@ namespace Terminal {
                     return;
                 }
 
-
                 title = term.window_title != "" ? term.window_title
                                                 : term.current_working_directory;
+                if (term.tab == null) {
```
lenemter, line 399:
```suggestion
                                                : term.current_working_directory;

                if (term.tab == null) {
```

lenemter (changes requested):

src/Application.vala
jeremypw: Removed

lenemter (approved):

PR #849, opened 2025-01-06
Show tab numbers with shortcut

### LINKED ISSUE #655: Show the number of tabs.
### Problem

When switching 6 more multiple tabs in a terminal using ALT + tab-number, it is pain to count its number repeatedly.

### Proposal

It will be nice that each tab shows its tab number for fast switching.

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #655

We can't just show tab numbers immediatley when `<Alt>` is pressed as this key is also used for e.g. `<Alt>BackSpace>`. The solution adopted is to use a "long-press" on the `<Alt>` key.  A previous solution was use the unused `<Alt>+0` combination to show the tab numbers immediately. This is retained for now but can be removed if considered superfluous.

The tab numbers are removed when a key is released, another key is pressed or focus lost.

It would be better to used an icon to show the numbers as that would be independent of the label and there is a suitable tabpage property,  but there is no quick way to create an icon from a number (there are no builtin number icons available and we cannot use an emblem). 

So the tab number is prefixed to the tab label for now.

Screenshot from 2025-01-06 20 27 58


zeebok:
Would it be possible to show tool tips for all of the tabs with the shortcut? Also, you say we can't use `<Alt>` because there is the `<Alt>Backspace` shortcut but couldn't it enter a state and show something when holding `<Alt>` for some amount of time?

jeremypw:
@zeebok 
> couldn't it enter a state and show something when holding <Alt> for some amount of time?

I have produced an implementation of this behaviour.

> Would it be possible to show tool tips for all of the tabs with the shortcut? 

Do you mean show each tab's shortcut window permanently until tab numbers removed? Not sure this is possible - maybe use the `show_help` signal (in Gtk3). I'll investigate.

jeremypw:
Cannot use the "show_help" signal for tab tooltips because tabpages are not widgets but sealed objects.

zeebok:
@jeremypw 

> Do you mean show each tab's shortcut window permanently until tab numbers removed? Not sure this is possible - maybe use the `show_help` signal (in Gtk3). I'll investigate.

Mostly trying to think of a way to display the emblems without having to modify the tab widgets directly. I thought maybe having the shortcut emblem in the tooltip and having it display while holding <Alt> might be a good way to only show the shortcuts in a temporary fashion. If it doesn't, it doesn't work.

jeremypw:
> @jeremypw
> 
> > Do you mean show each tab's shortcut window permanently until tab numbers removed? Not sure this is possible - maybe use the `show_help` signal (in Gtk3). I'll investigate.
> 
> Mostly trying to think of a way to display the emblems without having to modify the tab widgets directly. I thought maybe having the shortcut emblem in the tooltip and having it display while holding might be a good way to only show the shortcuts in a temporary fashion. If it doesn't, it doesn't work.

I see. The problem is the tabs are not widgets (or at least not accessible ones) so we can only do what the API allows :disappointed:

jeremypw:
Looks like the badges in the dock are styled labels not icons.  Gtk3 has `NumerableIcon` which would be suitable but it is now deprecated unfortiunately.

zeebok (approved):

zeebok (dismissed):

jeremypw:
@zeebok I have dismissed your review to ensure this is not prematurely merged - but it is appreciated as always.

@danirabbit I am doubtful whether the appearance of this PR is up to elementary standards - do you have any better solution? Or is the issue sufficiently worthwhile to fix anyway?

jeremypw:
Easier to create new Gtk4 branch

PR #863, opened 2025-02-27
MainWindow: GTK4 prep

danirabbit:
* Bump license header
* No more `decoration_layout_set` in GTK4
* Explicit expands
* Child property
* More accurate cast, no more `Gtk.Bin`
* ScrolledWindow doesn't take args in GTK4, so use vadjustment property
* Minimize use of Gtk.Clipboard functions that are going away
* Remove `set_visual`. This is I think really old stuff and it's not clear it even does anything for us in GTK3


jeremypw (approved):
This looks good and works as expected in Horus.  There are problems with OS8 in a VM but I think this is appears to be associated with the automatic creation of new workspace on fullscreening.  When that is turned off the problem goes away.

PR #864, opened 2025-03-29
Utils: add a warning about commands that skip confirmation

danirabbit:


jeremypw (approved):
Code looks good and works as expected. 

Maybe consider adding "--force" to check?

PR #866, opened 2025-04-01
Return and show all paste warnings

### LINKED ISSUE #865: Show all detected unsafe paste warnings
### Problem

At the moment unsafe paste detection only shows one type of warning, even if a given command contains multiple triggers

### Proposal

if ie "sudo apt update && sudo apt install -y fuse" is pasted, we should warn about admin privileges, multiple commands, and that it skips confirmations and not just whatever returns first.

This means the `out` value for unsafe past detection should be a string array and we should not return early but make sure we test for everything before returning

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #865


src/Utils.vala
```diff
@@ -129,15 +129,14 @@ namespace Terminal.Utils {
      *
      * @return true if safe, false if unsafe.
      */
-    public bool is_safe_paste (string text, out string msg) {
+    public bool is_safe_paste (string text, out string[]? msg_array) {
```
danirabbit, line 132: Does this need to be nullable? I don't think this ever goes to a null check. Would returning an empty array be fine?

jeremypw: Just avoids unnecessarily initializing an array which is not used but an empty array could be used if you prefer.

danirabbit (approved):

PR #877, opened 2025-07-16
Fix dropping files with '#' in name

### LINKED ISSUE #876: Dropping a file with '#' in the name causes truncation
### What Happened?

Drag/dropped a file from Files to Terminal named "### TEST ###.txt".  The uri pasted in to shell was truncated at the first '#'

### Steps to Reproduce

1. Create a file containing '#' in the name
2. Drag it from a filemanager into Terminal
3. Observe truncated path

### Expected Behavior

The path should be complete and quoted.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #876 

Dropped uris were being sanitized by creating a file from them and then obtaining the path but this seems unnecessary and was causing truncation.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -848,8 +848,7 @@ namespace Terminal {
                         }
                         // Sanitize the path as we do not want the `file://` scheme included
                         // and we assume dropped paths are absolute.
-                        file = File.new_for_uri (Utils.sanitize_path (unquoted_uri, "", false));
-                        path = file.get_path ();
+                        path = Utils.sanitize_path (unquoted_uri, "", false);
                         if (path != null) {
                             uris[i] = Shell.quote (path) + " ";
                         } else {


Marukesu:
> If the external application is sending files URIs, then he should be escaping those '#' in the filename as it denotes the start of the fragment part of the URI.

That being said, I don't think we should be removing the scheme as the code implies it does. We don't know the intent of those URIs. Even in the 'file://' case, we don't know if they are local or not.

jeremypw:
@Marukesu Could you raise a separate issue if there is a demonstrable downside to removing the "file://" scheme? This PR does not affect that behaviour.

As far as I can tell Terminal only uses dropped files as files if they are dropped on the "Add Tab" button.  Dropping on the Terminal widget just converts to text.  I considered removing URILIST target type but although that works you then get paste warnings.

jeremypw:
Closing in favor of #894 for now.  This PR also fixes dropping with Files but that should be fixed in Files.

PR #884, opened 2025-07-31
ZoomOverlay: Modernize/gtk4 preparation

jeremypw:
Any more prep needed?

Is there any problem changing license to LGPL for consistency with other files? Or should it be left as GPL?


jeremypw:
For some reason using `Object ()` style creation when sub-classing Granite OverlayBar does not work as expected (even though compiles and runs). The construct property `overlay` is deprecated in `granite` but not in `granite-7` for some reason.

danirabbit (changes requested):

PR #887, opened 2025-08-01
Application: Gtk4 preparation

jeremypw:
Some fairly substantive changes around `process-finished` notifications.

TODO (maybe).

Changes need to be made around restoring window size which entail changing the settings schema. These could be done before or during the port and are left for another PR.

@danirabbit When is the correct time to make settings schema changes?

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -1,6 +1,6 @@
 /*
  * SPDX-License-Identifier: LGPL-3.0-or-later
- * SPDX-FileCopyrightText: 2011-25 elementary, Inc. (https://elementary.io)
+ * SPDX-FileCopyrightText: 2011-2025 elementary, Inc. (https://elementary.io)
  */
 
 public class Terminal.Application : Gtk.Application {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -46,16 +46,8 @@ public class Terminal.Application : Gtk.Application {
                 }
             }
 
-            // This is a hack to avoid using Gdk-Xii dependency.  Using present_with_time ()
-            // with the current event time does not work either on X11 or Wayland perhaps
-            // because the triggering event did not occur on the Terminal window?
-            // Using set_keep_above () at least works on X11 but not on Wayland
-            //TODO It may well be possible to use present () on Gtk4 so this needs revisiting
-            window_to_present.set_keep_above (true);
-            window_to_present.present ();
-            window_to_present.grab_focus ();
             Idle.add (() => {
-                window_to_present.set_keep_above (false);
+                window_to_present.present ();
                 return Source.REMOVE;
             });
         });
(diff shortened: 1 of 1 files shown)


src/Application.vala
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright 2011-2023 elementary, Inc. (https://elementary.io)
- * SPDX-License-Identifier: LGPL-3.0-only
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ * SPDX-FileCopyrightText: 2011-25 elementary, Inc. (https://elementary.io)
```
danirabbit, line 3: I don't think this is a valid date format
```suggestion
 * SPDX-FileCopyrightText: 2011-2025 elementary, Inc. (https://elementary.io)
```

src/Application.vala
```diff
@@ -46,16 +46,8 @@ public class Terminal.Application : Gtk.Application {
...
 
-            // This is a hack to avoid using Gdk-Xii dependency.  Using present_with_time ()
-            // with the current event time does not work either on X11 or Wayland perhaps
-            // because the triggering event did not occur on the Terminal window?
-            // Using set_keep_above () at least works on X11 but not on Wayland
-            //TODO It may well be possible to use present () on Gtk4 so this needs revisiting
-            window_to_present.set_keep_above (true);
-            window_to_present.present ();
-            window_to_present.grab_focus ();
             Idle.add (() => {
```
danirabbit, line 49: That's a wild hack, I'm glad we're removing it 😅 Does this still need to be wrapped in an idle?

src/Application.vala
```diff
@@ -196,7 +188,7 @@ public class Terminal.Application : Gtk.Application {
                 terminal.tab.icon = process_icon;
             }
 
-            if (!(Gdk.WindowState.FOCUSED in terminal.main_window.get_window ().get_state ())) {
+            if (!(get_active_window ().is_active)) {
```
danirabbit, line 191: I didn't know this was already available in GTK3, very cool!

danirabbit (changes requested):
Just a tiny change and a question.

I think it makes sense to touch settings in a separate branch since that's not really related to this change

src/Application.vala
jeremypw: It seems to work OK on OS8 without the idle.  In `master` it doesn't work anyway on OS7 (and nor does this PR)

jeremypw:
> I think it makes sense to touch settings in a separate branch since that's not really related to this change

OK, so it can be done in Gtk3 before the port?

danirabbit:
> OK, so it can be done in Gtk3 before the port?

@jeremypw I think there's typically some finickyness about it that will need to be touched anyways, but it wouldn't hurt to get it a bit closer so there's less to do

PR #888, opened 2025-08-01
Mainwindow: Gtk4 preparation

jeremypw:
The substantive changes in this PR are around opening a uri or file path in the appropriate app. `Gtk.show_uri_on_window` does not exist in Gtk4 so an alternative approach is used that works both on Gtk3 and Gtk4.  Should probably move to Gtk.UriLauncher after porting but I could not get that to work last time I tried.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -905,13 +905,22 @@ namespace Terminal {
 
         private void action_open_in_browser () requires (current_terminal != null) {
             var uri = get_current_selection_link_or_pwd ();
-            string? to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location ());
-            if (to_open != null) {
-                try {
-                    Gtk.show_uri_on_window (null, to_open, Gtk.get_current_event_time ());
-                } catch (GLib.Error error) {
-                    warning ("Could not show %s - %s", to_open, error.message);
-                }
+            var to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location (), true);
+            AppInfo? appinfo = get_default_app_for_uri (to_open);
+            if (appinfo != null) {
+                //FIXME Gtk.UriLauncher (Gtk4) fails for some reason ( in a VM at least )
+                List<string> uris = null;
+                uris.append (to_open);
+                appinfo.launch_uris_async.begin (uris, null, null, (obj, res) => {
+                    try {
+                        appinfo.launch_uris_async.end (res);
+                    } catch (Error e) {
+                        warning ("Launcher failed with error %s", e.message);
+                        //TODO Handle launch failure - message box?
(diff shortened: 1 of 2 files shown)


src/MainWindow.vala
```diff
@@ -903,13 +906,22 @@ namespace Terminal {
...
             var uri = get_current_selection_link_or_pwd ();
-            string? to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location ());
-            if (to_open != null) {
-                try {
-                    Gtk.show_uri_on_window (null, to_open, Gtk.get_current_event_time ());
-                } catch (GLib.Error error) {
-                    warning ("Could not show %s - %s", to_open, error.message);
-                }
+            var to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location (), true);
+            AppInfo? appinfo = get_default_app_for_uri (to_open);
```
danirabbit, line 910: AppInfo has a static method for launching from the default. We don't need to get the default separately ahead of time https://valadoc.org/gio-2.0/GLib.AppInfo.launch_default_for_uri_async.html

src/MainWindow.vala
```diff
@@ -903,13 +906,22 @@ namespace Terminal {
...
+                appinfo.launch_uris_async.begin (uris, null, null, (obj, res) => {
+                    try {
+                        appinfo.launch_uris_async.end (res);
+                    } catch (Error e) {
+                        warning ("Launcher failed with error %s", e.message);
+                        //TODO Handle launch failure - message box?
+                    }
+                });
+            } else {
+                //TODO Show AppChooserDialog?
```
danirabbit, line 924: This is automatically handled by the portal

danirabbit (changes requested):

src/MainWindow.vala
```diff
@@ -903,14 +906,15 @@ namespace Terminal {
 
         private void action_open_in_browser () requires (current_terminal != null) {
             var uri = get_current_selection_link_or_pwd ();
-            string? to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location ());
-            if (to_open != null) {
+            var to_open = Utils.sanitize_path (uri, current_terminal.get_shell_location (), true);
+            AppInfo.launch_default_for_uri_async.begin (to_open, null, null, (obj, res) => {
```
danirabbit, line 910: I think you need to add the default launch context for this to send focus to the browser propertly
```suggestion
            var context = Gdk.Display.get_default ().get_app_launch_context ();
            AppInfo.launch_default_for_uri_async.begin (to_open, context, null, (obj, res) => {
```

danirabbit (changes requested):

danirabbit (approved):

PR #893, opened 2025-08-08
Check for possibly null primary monitor (e.g. in CI)

jeremypw:


danirabbit:
Tbh my personal opinion is that setting the window size based on the monitor size is trying to be too clever and probably doesn't make any sense on the extreme ends. Like at a certain small monitor point you just want it maximized and at a certain large monitor point you probably don't want it that big.

So I'm +1 on removing it if you are

lenemter (approved):

jeremypw:
Merged in order to get some other PRs to pass CI.  Will produce a separate PR to remove the `get_geometry ()` code altogether.

PR #894, opened 2025-08-11
Don't sanitize URI on drop

lenemter:
Fixes https://github.com/elementary/terminal/issues/876
Closes https://github.com/elementary/terminal/pull/877

URIs **must** escape special characters and Utils.sanitize_path undoes this. Furthermore, there's no other reasons to use  Utils.sanitize_path here -- dropped paths must be absolute. So in conclusion if a file manager does everything correctly, we shouldn't need this. I tested this branch using Dolphin.

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -835,7 +835,6 @@ namespace Terminal {
                 case DropTargets.URILIST:
                     var uris = selection_data.get_uris ();
                     string path;
-                    File file;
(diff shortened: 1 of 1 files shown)


jeremypw:
We need to fix DnD in Files before this can be merged - otherwise truncation still occurs when testing with Files.

jeremypw (approved):
Fixes the issue for Dolphin and Nautilus.  Needs to be fixed in Files itself. Thanks!

jeremypw:
Not sure why CI fails - passes OK locally.

PR #896, opened 2025-08-14
No longer set window size from monitor as fallback

jeremypw:
Further to #893.


danirabbit (approved):
This looks  good but for whatever reason tests are failing

Issue #899, opened 2025-08-15
Search button is focused instead of the prompt on startup


ryonakano:
### What Happened?

Terminal does no longer focus on the prompt on startup and instead it focus on the search button:

<img width="908" height="659" alt="Image" src="https://github.com/user-attachments/assets/7160fa9d-04d3-48ca-8b37-3576e1603bee" />

I think it's to type a command that users would like to do first when they launch Terminal, but the current behavior requires extra step to manually focus on the prompt, which is a bit frustrating for me.

### Steps to Reproduce

1. Open Terminal
2. See the focus is not in the prompt but in the search button

### Expected Behavior

The prompt is focused on startup.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Classic Session (X11, This is the default)

### Software Version

Latest release (I have run all updates)

### Log Output

```shell

```

PR #903, opened 2025-08-18
Focus tab when opening new window

lenemter:
Fixes https://github.com/elementary/terminal/issues/899
Closes https://github.com/elementary/terminal/pull/901

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -389,29 +389,27 @@ namespace Terminal {
                 current_terminal = term;
                 zoom_overlay.hide_zoom_level ();
 
-                // Happens on closing window - ignore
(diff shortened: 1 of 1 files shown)


lenemter:
@jeremypw sure.

In #901 you say "This issue seems to be cause by a "feature" of the Hdy.TabView whereby the notify["selected-page"] signal is not emitted when the first tab is programmatically selected on start up." which is not true. The signal is emmited, but the issue is that the handler function exits early when creating new window (this is written in the comment). So this branch fixes the issue by moving `term.grab_focus` so that that the term always gets focus regardless of whether the window was just created or a tab was switched.

jeremypw (approved):

PR #904, opened 2025-08-18
Do not use xvfb for Github CI

jeremypw:
As we move away from X and to Gtk4/Wayland (under which xvfb does not work) and Github CI increasingly fails Application tests inconsistently,  we should remove dependence of Github CI on xvfb  (which provides an X framebuffer when running headless).  

Note that the Application tests still run *locally* so are retained.  They are skipped when there is no display available.

### DIFF
--- a/src/tests/Application.vala
+++ b/src/tests/Application.vala
@@ -102,7 +102,6 @@ namespace Terminal.Test.Application {
 
         GLib.Test.init (ref args);
 
-
         /* MainWindow always create a tab in home, make it the current dir
          * so that we don't create a extra tab during tests
          */


tintou:
Note that a recommended alternative is to use weston: `weston --backend=headless --socket=wayland-1` then `export WAYLAND_DISPLAY=wayland-1`

src/tests/Application.vala
```diff
@@ -96,18 +96,38 @@ namespace Terminal.Test.Application {
...
          * so that we don't create a extra tab during tests
          */
         Environment.set_current_dir (Environment.get_home_dir ());
 
+        GLib.Test.add_func ("/application/test-display", () => {
+            Terminal.Test.Application.has_display = true;
+            Gdk.set_allowed_backends ("x11");
+            var x11_display = Gdk.Display.open (null);
+            if (x11_display == null) {
+                Gdk.set_allowed_backends ("wayland");
```
tintou, line 116: why testing for each back-end? We don't do anything with the fact that it is X/Wayland

jeremypw: Perhaps I am misunderstanding something - I am trying to test whether there is *either* an X11 *or* a Wayland display.  If there is neither then tests will not be run.

tintou: I don't think that this is necessary, doing `Gdk.Display.open (null) != null` should be enough I think

jeremypw:
Converting to draft as tests will probably be fixed in the Gtk4 port.

Issue #729, opened 2023-05-31
Semi-translatent background becmoes opaque when Terminal is tiled or maximized


1280px:
### What Happened?

I use semi-transparent bg to save screen space, but when terminal is tiled/maximized (but not in fullscreen mode!), the background always becomes opaque. This should **not** be the case, as it essentially makes semi-transparent background useless, not allowing to see the text and windows behind the terminal.

The issue seems to happen because of the way Gala treats H/V maximized windows, however I wonder if there is any workaround to make Terminal an exception.

### Steps to Reproduce

1. Open Terminal
2. Set the theme to "Custom"
3. In Custom theme settings, set background colour to any semi-transparent value
4. Maximize/tile the Terminal window:

![image](https://github.com/elementary/terminal/assets/71165491/624336e0-b591-4faa-b377-b187906441e6)

### Expected Behavior

![image](https://github.com/elementary/terminal/assets/71165491/e7344bb0-6829-438d-a3b7-f5d556dfbbed)

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

1280px:
Found a related issue, someone made a CSS workaround for it too: https://github.com/elementary/terminal/issues/303 (it only mentions maximizing, however tiling is also affected).

However, it was closed because:

>Going opaque on maximize is intentional because it fixes some UI issues but also because if the Terminal is maximized, it's likely users are focusing on that content.

Now, with custom themes being supported and default ones being fully opaque anyway, I wonder is it worth reconsidering it? I thought the issue was somewhere in how Gala handles maximized windows, but it turned out to be much simpler.

1280px:
I am confused, is this issue out of scope or just not planned? I can propose a fix and more reasoning.

PR #911, opened 2025-08-27
Fix unexpected opacity change on maximize

jeremypw:


danirabbit:
This is intentional and fixes a reported issue

https://github.com/elementary/terminal/pull/239

https://github.com/elementary/terminal/pull/152

danirabbit:
> @danirabbit Oh, OK - I wasn't sure why this was done. I'd better mark #729 as "Design Conflict" then.  Are you sure the issue cannot be fixed elsewhere?

I think ideally Gala itself would black out the space behind maximized/tiled windows. This is kind of a workaround. But that would end with the same result of it being pointless to have transparency behind a maximized/tiled window

jeremypw:
> I think ideally Gala itself would black out the space behind maximized/tiled windows.

Indeed it would be better - or losing rounded corners when tiled would have less impact. To be honest I cannot see the original bug when tiled on OS8 now anyway - is it still there?

By only patching Terminal it means third-party terminals like Black Box continue to be transparent when tiled/maximized.  I guess only terminals are likely to use transparency in this way?

PR #913, opened 2025-09-03
Fix tests in Gtk4

jeremypw:
Trying different things.

Update:
 - Everything now passes locally except the quit action, which can fail
 - If quit action test omitted passes locally with `ninja test`
 - Have to add `dbus-x11` and `at-spi2-core` to CI container (used to be installed by default?)
 - On GitHub CI passes on stable (:tada: ) but fails on unstable and development branch.
 
Problem is : 
 ```# DEBUG: Settings.vala:172: cannot use the portal, using the AccountsService: 

GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: 
The name org.freedesktop.portal.Desktop was not provided by any .service files

not ok /application/action/new-window -

 FATAL-CRITICAL: Settings.vala:114: Could not connect: No such file or directory

```

Seems to be a problem connecting to settings using the unstable version of `Granite`.


jeremypw:
@danirabbit Any idea why the tests are passing on `stable` but not on the other two targets? Appears to be due to recent changes in Granite around Settings but exactly what the problem is I have no idea.

PR #921, opened 2025-09-21
Fix Clear Screen: Move cursor to line start before feeding clear command

### LINKED ISSUE #920: Clear screen does not work properly if there is pending input
### What Happened?

Entered part of a command without pressing `Enter`.  Selected "Clear Screen" from context menu.  The `clear` command was appended to the existing input and so did not work.

### Steps to Reproduce

1. Open a tab
2. Enter something random without pressing `Enter`
3. Select "Clear Screen" from the context menu
4. Observe error

### Expected Behavior

Clear any existing input before feeding in the `clear` instruction

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #920 

Also fixes related functions using feed_child

### DIFF
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -553,7 +553,7 @@ namespace Terminal {
             //Clear pending input
             Posix.kill (child_pid, Posix.Signal.INT); //Equivalent to pressing Ctrl-C
 
-            //Wait for signal handler to finish before feedng in the command
+            //Wait for signal handler to finish before feeding in the command
             Timeout.add (10, ()=> {
                 feed_child (command.data);
                 return Source.REMOVE;
@@ -570,10 +570,6 @@ namespace Terminal {
             var command = "cd '%s';clear\n".printf (path);
             feed_command (command);
         }
-
-        // public void clear_pending_input () {
-        //     Posix.kill (child_pid, Posix.Signal.INT);
-        // }
 
         protected override void paste_clipboard () {
             clipboard.request_text ((clipboard, text) => {


jeremypw:
Converting to draft as does not completely fix issue.

src/Widgets/TerminalWidget.vala
```diff
@@ -547,12 +549,21 @@ namespace Terminal {
...
+                feed_child ("\rreset\n".data);
             }
         }
 
+        public void clear_pending_input () {
+            // This is hacky but no obvious way to feed in escape sequences to clear the line
+            // Assume any pending input is less than 1000 chars.
+            string backspaces = string.nfill (1000, '\b');
+            feed_child (backspaces.data);
+        }
```
lenemter: Can't you just pass Ctrl+C signal to clear current line instead of this?

jeremypw: Finally worked out how to do this with a signal as you suggested!  Thanks.

src/MainWindow.vala
```diff
@@ -963,14 +963,14 @@ namespace Terminal {
...
+            // We need to clear any pending input else it can get unexpectedly executed
+            term.clear_pending_input ();
 
             // Change to requested directory
             var command = "cd '" + path + "'\n";
             term.feed_child (command.data);
 
-            // Clear screen
+            // Clear screen to hide fed in data
             term.feed_child ("clear\n".data);
```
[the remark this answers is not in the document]
jeremypw: Good point! I have moved this code into TerminalWidget so all the feed_child related functions are there and can be DRYed

PR #927, opened 2025-10-04
Auto-hide single tab option

### LINKED ISSUE #832: Setting to hide tab title when there is only one tab no more work

jeremypw:
Fixes #832 

Based on #833 - completed but removed "Never" option

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -422,18 +422,6 @@ namespace Terminal {
                 term.tab.icon = null; // Assume only process icons are set
             });
 
-            notebook.tab_bar.notify["autohide"].connect (() => {
-                new_tab_revealer.reveal_child =
-                    notebook.tab_bar.autohide &&
-                    notebook.tab_view.n_pages == 1;
-            });
-
-            notebook.tab_view.notify["n-pages"].connect (() => {
-                new_tab_revealer.reveal_child =
-                    notebook.tab_bar.autohide &&
-                    notebook.tab_view.n_pages == 1;
-            });
-
             notebook.tab_bar.bind_property (
                 "tabs-revealed", new_tab_revealer, "reveal-child", SYNC_CREATE | INVERT_BOOLEAN
             );


jeremypw:
I thought about that but @danirabbit mentioned introducing a "never" option
in the future.
Jeremy Wootten

src/Widgets/SettingsPopover.vala
```diff
@@ -156,6 +167,9 @@ public sealed class Terminal.SettingsPopover : Gtk.Popover {
         Application.settings.changed.connect ((s, n) => {
             if (n == "background" || n == "foreground") {
                 custom_button.update_theme_provider ();
+            } else if (n == "tab-bar-behavior") {
+                var autohide = Application.settings.get_enum ("tab-bar-behavior") == 1;
+                auto_hide_button.active = autohide;
```
lenemter:
```suggestion
                auto_hide_button.active = Application.settings.get_enum ("tab-bar-behavior") == 1;
```

src/MainWindow.vala
```diff
@@ -410,6 +421,18 @@ namespace Terminal {
...
+                    notebook.tab_bar.autohide &&
+                    notebook.tab_view.n_pages == 1;
+            });
+
+            notebook.tab_view.notify["n-pages"].connect (() => {
+                new_tab_revealer.reveal_child =
+                    notebook.tab_bar.autohide &&
+                    notebook.tab_view.n_pages == 1;
+            });
+
```
lenemter: It shouldn't use `tab_bar.autohide`, it should rely on gsettings instead.

jeremypw:
@lenemter Thanks again for your excellent review on Terminal PRs :heart:  One suggestion here I am not sure about but will implement if you feel strongly.

src/MainWindow.vala
lenemter: "n-pages" is already exposed by TerminalView. You should use it instead of making tab_view public. But for this to work you need to explicitly notify when n-pages changes using `notify_property ("n-pages")`

jeremypw: TabView is already directly accessed by MainWindow in many other places so making the suggested change would involve refactoring not directly related to the issue.   I agree that in principle it would be more "modular" to make both TabView and TabBar private members of TerminalView but would you be happy to defer that to a separate code improvement PR?  Probably best left until after the Gtk4 port now in order avoid delaying that?

lenemter: I'm fine with that

lenemter (approved):
LGTM!

PR #930, opened 2025-10-23
Fix some problems running commands from the commandline.

### LINKED ISSUE #311: Does elementary terminal properly respect -e flag?
would like to automate showing of a nice weather report you can launch in terminal:
`curl wttr.in/bydgoszcz`
screenshot from 2018-10-17 14 46 16 2x
with follwoing script launched from the system menu:
```
#!/bin/sh
exec io.elementary.terminal -e "curl wttr.in/bydgoszcz"
```
But this way the output gets a little "incomplete", just like if the command got executed too fast (notice where the user@machine line went here):
screenshot from 2018-10-17 14 46 45 2x
**So, is there a way to delay the command after -e flag?
Or maybe totally different approach will automate this to show the output properly?**

Tried also this:
```
#!/bin/sh
WTTR=$(curl -s wttr.in/bydgoszcz) && \
exec io.elementary.terminal -e "echo $WTTR"
```
but it produced this:
screenshot from 2018-10-18 22 11 53 2x
When doublequoted - `-e "echo "$WTTR""` - the output is single word - weather
When tried `-e 'echo "$WTTR" '` - output is $WTTR
When tried `-e echo "$WTTR"` - the output is empty line

After trying this:
```
#!/bin/sh
exec io.elementary.terminal -e "sleep1s & curl wttr.in/bydgoszcz"
```
The output is a single dot...
screenshot from 2018-10-18 19 58 55 2x
...and once the terminal gets closed in this case, the whole user session gets restarted! :astonished:
I've run out of ideas :face_with_head_bandage:

### LINKED ISSUE #929: Tab closes after command launched from commandline exits
### What Happened?

With some commands e.g. "sleep 3" or "conda activate <environment>", if launched from the commandline with the `-e` or `-x` option, the tab they were running in closes after the foreground process ends so the output cannot be seen or, in the case of `conda` the required environment used.

According to the code comments this is not the intended behaviour. The tab should remain open.

### Steps to Reproduce

1. Run `io.elementary.terminal -e "sleep 3"` from the commandline.
2. Observe tab closes

### Expected Behavior

The tab stays open

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #929 
Fixes #311  (afaict)

Instead of spawning the requested program directly with Vte (which automatically creates a shell to run it in), we create the shell as usual and then feed the command in as if typed at the commandline.  This leads to more predictable results and the shell does not exit with its child.

It is also now possible to run a sequence of commands in the same tab without using a script by separating them with semi-colons and enclosing in quotes:
`io.elementary.terminal -e "command1;command2;command3"`

This PR arises from the discussion at https://github.com/elementary/terminal/discussions/926


jeremypw:
Closing in favour of #1008

Issue #931, opened 2025-11-03
Add an option to disable right-click context menu


lobre:
### Problem

The right-click context menu in Terminal (which includes actions like Copy, Paste, and Show in Files) is useful in many situations.

However, some terminal applications rely on right-click input for their own interactions. In these cases, the context menu becomes intrusive. For example, in Kakoune, right-clicking is used to extend text selections. Currently, when doing so in elementary Terminal, the context menu still appears on top of the terminal, even though the click event correctly reaches the application underneath.

This dual behavior makes it inconvenient to use such applications with mouse input, since the menu partially obstructs the interface and must be dismissed manually.

There doesn’t seem to be a reliable way for Terminal to automatically detect when a running program expects right-click input, so a user-side control would help.

### Proposal

Add a preference or setting to disable the right-click context menu.

Possible implementations:

- A toggle in Preferences (e.g. “Show context menu on right-click”).
- Alternatively, allow the menu to be triggered via an icon in the sidebar or Shift + Right-Click when disabled.

PR #933, opened 2025-11-06
Disable secondary-click actions while a foreground process is running

lobre:
This is a try to tackle https://github.com/elementary/terminal/issues/931.

This is my first PR and I am not familiar with the codebase, so I would like to be advised if this is the correct way of doing that!

Thanks a lot in advance


jeremypw (approved):
Thanks for tackling Terminal development!  This seems to be the most straightforward solution and I cannot see any drawbacks so lets merge it!

The user still has the option to reload the tab (and kill the process) via the tab label context menu if required.

PR #940, opened 2025-11-18
Fix some warnings

lenemter:
Vala warnings are easy to fix and I fixed most of them

C warnings are hard to fix and most of the time it's the issue in vapi files or libgee (?) I fixed a few simple warnings with marking some methods as static and avoiding duplicating strings


danirabbit (approved):
Looks good! Nice cleanups

PR #942, opened 2025-11-18
Fix search shortcuts

lenemter:
Changes:
1. Change key controller propagation phase to CAPTURE, TARGET is too late :(
2. Check if search bar is focused using "focus_widget.is_ancestor (search_toolbar)" instead of directly checking if searchbar is focused. Gtk.SearchBar is a widget that has Gtk.Text as a child, so this text widget is getting the focus, not the searchbar itself.
3. Correctly implement search action, it was missing change_state handler. Activate function now just toggles the state.
4. Don't rely on search_button's state to determine search action state.


jeremypw (approved):
LGTM. No regressions found

Issue #714, opened 2023-03-27
Allow to use Ctrl+Del shortcut


aitor-gomila:
### Problem

Usually in VSCode and other terminal emulators you're able to remove the last piece of code with Ctrl+Supr.

This is muscle memory for me

### Proposal

For example, given the command `git clone https://`, it would remove all the text from the currently selected position until it finds a symbol or space. One key press would lead to `git clone https`, another would lead to `git clone`, etc.

### Prior Art (Optional)

Will post some art soon, can't access github on gnome web right now. Also, willing to contribute this if someone points out exactly how to do it :)

jeremypw:
@aitor-gomila Should the title of this issue be "Allow use of `Ctrl + Super` shortcut for delete word left"?  `Ctrl + Del` already works for delete word right.  As `Alt + BackSpace` already works for delete word left it may be possible to capture `Ctrl + Super` (if it is not already captured by the window manager) and feed `Alt + BackSpace` into the terminal instead.

stan-janssen:
@jeremypw I think @aitor-gomila uses "Supr" as in the Spanish key for `delete`.

aitor-gomila:
Yes, I meant backspace (is that how it's called?) thanks for clarifying

jeremypw:
OK, so we could map `Ctrl + BackSpace` to `Alt + BackSpace` then.  Making shortcuts discoverable is a bigger issue, affecting all apps. They can be put in menus and tooltips but in a case like this it is more problematic as there is no actionable widget with the same effect.  Some ideas were discussed here https://github.com/elementary/shortcut-overlay/issues/53 but there are technical difficulties that are yet to be resolved.

PR #949, opened 2025-11-20
Allow Ctrl+Backspace for deleting words

lenemter:
Fixes https://github.com/elementary/terminal/issues/714

This probably interferes with some CLI applications. Does this need a toggle?


lenemter:
@jeremypw Done

PR #953, opened 2025-11-22
Move alt+[1-8] shortcuts to TerminalView

lenemter:
Fixes https://github.com/elementary/terminal/issues/955

### DIFF
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -273,21 +273,18 @@ public class Terminal.TerminalView : Granite.Bin {
             case Gdk.Key.@6:
             case Gdk.Key.@7:
             case Gdk.Key.@8:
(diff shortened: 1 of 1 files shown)


src/Widgets/TerminalView.vala
```diff
@@ -257,6 +263,44 @@ public class Terminal.TerminalView : Granite.Bin {
...
+        switch (keyval) {
+            case Gdk.Key.@1: //alt+[1-8]
+            case Gdk.Key.@2:
+            case Gdk.Key.@3:
+            case Gdk.Key.@4:
+            case Gdk.Key.@5:
+            case Gdk.Key.@6:
+            case Gdk.Key.@7:
+            case Gdk.Key.@8:
+                if (ALT_MASK in modifiers &&
```
jeremypw: We could probably determine tab_index first and then test `n_pages > tab_index`? Not sure its worth having two `if` clauses.

PR #958, opened 2025-11-25
Offset appmenu popover to keep in window

jeremypw:
I noticed when preparing new screenshots that the AppMenu button popover no longer stays within the window. 
I believe this is due to differences between Gtk3 and Gtk4.  This PR maintains the previous appearance by offsetting the popover.

Please advise if there is a more elegant method.

BEFORE:
Screenshot from 2025-11-25 15 11 05
AFTER:
Screenshot from 2025-11-25 15 09 57
GTK3
Screenshot from 2025-11-25 15 15 30

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -300,7 +300,7 @@ namespace Terminal {
 
             menu_button.popover.show.connect (() => {
                 var p = menu_button.popover;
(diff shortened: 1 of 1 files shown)


danirabbit:
I believe you can just use `halign'

PR #961, opened 2025-11-26
Gtk4: Update screenshots to Gtk4 version

jeremypw:
Replace existing screenshots with similar Gtk4 versions.  One screenshot has been renamed and shows additional features.

The screenshots were taken at 200% scaling .

This should be merged immediately before v8.0.0 release commit.

PR #970, opened 2025-12-29
Release version 8.0.0

jeremypw:
Includes #961 

Waiting for merging of final milestone PRs:

 - [x] #999 
 - [x] #1000


danirabbit:
I feel like maybe this should bake a little longer. Especially with the GTK4 port and it seems like there's a number of other refactors landing and velocity seems pretty high. Unless there's like an important bug fix that needs to happen

jeremypw:
@danirabbit OK, I'll hold for now.

jeremypw:
Putting back on hold as there are three more issues/regressions that should be fixed before release of Gtk4 version.

danirabbit (approved):

PR #979, opened 2026-01-27
Cleanup clipboard and process-finished action

lenemter:
See commit messages for intent of these changes


jeremypw (commented):
Elementary code style only requires whitespace *after* an outer clause ending `}` or `)` so I think some of the whitespace added here is unnecessary.  Apart from that the changes make sense

danirabbit (approved):

PR #994, opened 2026-02-17
Delay process end notification

### LINKED ISSUE #947: Process completed notification while terminal is focused
### What Happened?

When I run the [pass](https://www.passwordstore.org/) command to copy a password to the clipboard, a "process completed" notification is triggered, even though **the Terminal app is in focus**. This program uses the _wl-clipboard_ package to interact with the clipboard on Wayland, and starts a background process to clear it after a few seconds.

Observations:
- The terminal is in focus the whole time, although **it does seem to blink** for a moment — maybe from losing and regaining focus?
- The notification is never retracted, even if I <kbd>alt</kbd> + <kbd>tab</kbd> away and back to it.
- It doesn't seem to happen if I **move the Terminal window to an empty workspace** and attempt to reproduce it there.

io.elementary.terminal 7.2.0

https://github.com/user-attachments/assets/6ee6940a-845f-4959-b385-bb443fbd9355

### Steps to Reproduce

1. Start a Terminal window and run the `pass -c` command to copy a password to the clipboard. A notification should appear when it's finished.

### Expected Behavior

A "process completed" notification should only be sent if the Terminal isn't in focus at the time. This is probably a bug somewhere.

I cannot reproduce this on GNOME or KDE, using terminal apps such as Konsole, GNOME Terminal, GNOME Console or Ptyxis.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #947 

Do not send notification in case of temporary focus out.

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -187,26 +187,30 @@ public class Terminal.Application : Gtk.Application {
                 terminal.tab_state = NONE;
             }
 
-            if (!(get_active_window ().is_active)) {
-                var notification = new Notification (notification_title);
-                notification.set_body (process);
-                notification.set_icon (tab_state.to_icon ());
-                notification.set_default_action_and_target_value ("app.process-finished", new Variant.string (id));
-                send_notification ("process-finished-%s".printf (id), notification);
+            Timeout.add (200, () => {
+                if (!(get_active_window ().is_active)) {
+                    var notification = new Notification (notification_title);
+                    notification.set_body (process);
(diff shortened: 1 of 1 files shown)


danirabbit (approved):
I can't confirm the OP's issue, but if you can reproduce it I believe you that this fixes it :) I'm not a massive fan of magic number timeouts, but it seems fine ✔️

PR #999, opened 2026-02-19
Provide fallback for title; set title in idle

### LINKED ISSUE #992: Window title can be blank on new tab until tab switch
### What Happened?

This is a follow-up to https://github.com/elementary/terminal/issues/989#issuecomment-3913207059.

On the Gtk4 version, when opening a new tab (or sometimes on startup), the window/tab title can initially be blank. It usually fixes itself after switching to another tab and coming back.

Looking at the behavior, and from what I see in the code, the title seems to be set mainly in two places:

1. When the selected tab changes (selected-page handler), it sets the title to `window_title` or `current_working_directory`.
2. When the terminal emits `window_title_changed`, it sets the title to `current_terminal.window_title`.

With a new tab:

- At first, both `window_title` and `current_working_directory` are often still empty.
- The selected-page handler runs and sets the title, which ends up being empty.
- Later, the cwd becomes known asynchronously, but that event does not directly refresh the window title.
- When switching tabs and coming back, selected-page fires again, and now cwd is filled, so the title finally appears.

So this looks like an event timing issue.

### Steps to Reproduce

1. Open elementary Terminal (Gtk4 version from daily PPA).
2. Open a new tab.
3. Observe the tab/window title immediately.
4. Switch to another tab and then back.

In some cases, the title is blank at first and only appears after switching tabs.

### Expected Behavior

The window title should never be blank.

If both `window_title` and `current_working_directory` are empty at the time the tab becomes selected, there should be:

- either a fallback value like "Terminal", or
- a refresh triggered when cwd becomes available.

That would avoid the brief empty state and make the behavior feel more consistent.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #1002: Window title does not refresh after cd until tab focus changes
### What Happened?

When I change directory in a tab (for example, `cd /tmp`), the tab label updates, but the window title still shows the old directory.

The window title only refreshes after another UI event, like switching to another tab and coming back.

From what we discussed, it seems related to the split between:

- CWD tracking is used for tab label updates, and
- VTE `window-title-changed` updates used for the window title.

From a quick code read, `cwd_changed` updates tab labels but does not directly call `set_title()` in `MainWindow` (`src/MainWindow.vala:773`), while tab switches do call `set_title()` (`src/MainWindow.vala:293`), which matches the behavior above.

Related context: originally discussed in #992, but opening this separate issue as requested.

Please check the following comment on that other issue, on which there is a screencast where we see the problem.

https://github.com/elementary/terminal/issues/992#issuecomment-3927476377

### Steps to Reproduce

1. Open elementary Terminal.
2. In a tab, note the current window title path.
3. Run `cd /tmp` (or any other directory).
4. Observe that the tab label updates, but the window title does not.
5. Switch to another tab and back.
6. Observe that the window title now updates to the correct directory.

### Expected Behavior

The window title should update immediately when the shell working directory changes (same moment the tab label updates), without needing a tab switch or other manual refresh.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #992 
Fixes #1002 

Due to a suspected race condition in the "selected-page" handler, the window title is set later (in a pre-existing Idle).  Also a fallback for the title is now provided so it should never be blank.

It appears the `check_for_tabs_with_same_name` function used to update the window title but that is no longer the case.  The window title is now explicitly updated in the cwd change handler.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -288,12 +288,9 @@ namespace Terminal {
                     return;
                 }
 
-                title = term.window_title != "" ? term.window_title
-                                                : term.current_working_directory;
-
-
                 // Need to wait for default handler to run before focusing
                 Idle.add (() => {
+                    set_title ();
                     term.grab_focus ();
                     return Source.REMOVE;
                 });
@@ -463,8 +460,8 @@ namespace Terminal {
         private void connect_terminal_signals (TerminalWidget terminal_widget) {
             terminal_widget.child_exited.connect (on_terminal_child_exited);
             terminal_widget.cwd_changed.connect (on_terminal_cwd_changed);
-            terminal_widget.foreground_process_changed.connect (on_terminal_program_changed);
-            terminal_widget.window_title_changed.connect (on_terminal_window_title_changed);
+            terminal_widget.foreground_process_changed.connect (set_title);
+            terminal_widget.window_title_changed.connect (set_title);
         }
 
         private void on_terminal_child_exited (Vte.Terminal term, int status) {
@@ -491,8 +488,14 @@ namespace Terminal {
             }
         }
 
-        private void on_terminal_window_title_changed () {
-            title = current_terminal.window_title;
+        private void set_title () {
+            if (current_terminal.window_title != "") {
+                title = current_terminal.window_title;
+            } else if (current_terminal.current_working_directory != "") {
+                title = current_terminal.current_working_directory;
+            } else {
+                title = TerminalWidget.DEFAULT_LABEL; // Do we need a different fallback window title?
+            }
         }
 
         private bool close_immediately = false;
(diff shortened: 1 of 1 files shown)


src/MainWindow.vala
```diff
@@ -480,8 +477,14 @@ namespace Terminal {
             }
         }
 
-        private void on_terminal_window_title_changed () {
-            title = current_terminal.window_title;
+        private void set_title () {
```
danirabbit: `set_title` is already a function of Gtk.Window so this probably needs a different name like `set_title_label`

danirabbit (changes requested):

src/MainWindow.vala
jeremypw: Thanks for noticing that. Changed as suggested.

danirabbit (approved):

PR #1000, opened 2026-02-22
Fix excessive scroll speed in Wayland

### LINKED ISSUE #993: Touchpad scrolling feels overly sensitive since Gtk4 migration
### What Happened?

Since the Gtk4 migration, touchpad scrolling in the terminal buffer feels much more sensitive than before.

Very small movements on the touchpad cause large jumps in scrollback, which makes precise scrolling difficult. I do not remember this behavior in the Gtk3 version.

Mouse wheel scrolling feels mostly normal. The issue appears specific to smooth touchpad scrolling.

Looking at the code, Terminal embeds `Vte.Terminal` inside a `Gtk.ScrolledWindow`. Gtk4 changed scroll semantics (smooth/pixel-based input and different propagation behavior), so I suspected the issue might be related to how VTE handles scroll units and fallback scrolling in this setup.

By default, VTE properties are:

- `enable_fallback_scrolling = true`
- `scroll_unit_is_pixels = false`

I experimented locally and ended up modifying both to:

```
enable_fallback_scrolling = false;
scroll_unit_is_pixels = true;
```

With these two changes, touchpad scrolling feels consistent and proportional again, much closer to the Gtk3 behavior.

I don’t fully understand the exact interaction between Gtk4 smooth scrolling and these VTE settings, so I’m not claiming this is the definitive fix. But at least, this restores "normal scroll feel" on my system.

### Steps to Reproduce

1. Use the Gtk4 (daily) version of elementary Terminal.
2. Open a terminal with significant scrollback.
3. Scroll using a touchpad.
4. Observe that very small movements cause large jumps in the buffer.

### Expected Behavior

Touchpad scrolling should feel proportional and precise, similar to the previous Gtk3 version.

Small touchpad movements should result in small scroll deltas without amplified jumps.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #993 

Thanks to @lobre 

These properties reduced scroll speed to a level consistent with other apps.

### DIFF
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -146,6 +146,8 @@ public class Terminal.TerminalView : Granite.Bin {
 
         var terminal_widget = new TerminalWidget () {
             scrollback_lines = Application.settings.get_int ("scrollback-lines"),
+            enable_fallback_scrolling = false,
+            scroll_unit_is_pixels = true,
             /* Make the terminal occupy the whole GUI */
             hexpand = true,
             vexpand = true
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -146,6 +146,7 @@ public class Terminal.TerminalView : Granite.Bin {
 
         var terminal_widget = new TerminalWidget () {
             scrollback_lines = Application.settings.get_int ("scrollback-lines"),
+            /* Ensure usable scroll speed on Wayland - see https://github.com/elementary/terminal/issues/993 */
             enable_fallback_scrolling = false,
             scroll_unit_is_pixels = true,
             /* Make the terminal occupy the whole GUI */


jeremypw:
I think this should be merged before release of the Gtk4 port.  Previously I was using OS8.1 in a VM running on an X host so did not see this problem.  However, I am now running OS8.1 natively and this issue makes touchpad scrolling on the view almost unusable although it does not affect scrolling while hovering the scroll-bar for some reason.

src/Widgets/TerminalView.vala
```diff
@@ -146,6 +146,8 @@ public class Terminal.TerminalView : Granite.Bin {
 
         var terminal_widget = new TerminalWidget () {
             scrollback_lines = Application.settings.get_int ("scrollback-lines"),
+            enable_fallback_scrolling = false,
+            scroll_unit_is_pixels = true,
```
danirabbit, line 151: Can you add a comment with a link to the issue so it's easier to remember why we're doing this?

jeremypw: Commented as suggested - thanks.

danirabbit (approved):

PR #1003, opened 2026-02-23
Do not show loading spinner for all foreground processes

### LINKED ISSUE #990: Spinner behavior seems tied to terminal title updates, not to long-running foreground tasks
### What Happened?

While testing spinner behavior, I noticed it appears for some long-lived interactive apps and not for others. After debugging, it looks like the spinner is triggered when the terminal title changes while a foreground process is active, rather than being tied directly to "this command is long-running".

This makes the spinner show up in some uncommon cases (apps that update the terminal title), and not in other obvious long-running cases like `sleep 10 && echo hello`.

### Steps to Reproduce

I struggled to find a reproducer, and ended up doing it with `neovim`.

1. Open elementary Terminal.
2. Run: `nvim -u NONE -i NONE +'set title titlestring=spinner-demo titleold=Terminal | redraw!'`
3. Observe the tab spinner while Neovim is open.
4. Quit Neovim with `:q` and observe spinner behavior stops.

### Expected Behavior

I expected spinner behavior to map more clearly to long-running/unfinished command activity. Right now, it seems more like a heuristic tied to title-change events during foreground activity, which is hard to reason about and hard to reproduce unless the app also sets the terminal title. Is this the intended design? If yes, what class of applications/workflows was this spinner behavior originally meant to represent?

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

### LINKED ISSUE #991: Spinner is shown as a static icon when "Reduced Motion" is enabled
### What Happened?

This is a follow-up to https://github.com/elementary/terminal/issues/989#issuecomment-3913207059.

When "Reduce Motion" is enabled in Desktop Settings, the spinner shown in the tab when a foreground process is running does not animate, which is expected. However, it is still displayed as a spinner icon, just frozen.

Visually, this looks a bit odd because it appears like a broken or stuck spinner rather than an intentional static indicator.

### Steps to Reproduce

So far, the reproducer I have for the spinner is using `neovim`, please see [this issue](https://github.com/elementary/terminal/issues/990) for more context.
 
1. Enable "Reduce Motion" in Desktop Settings.
2. Open elementary Terminal (Gtk4 version).
3. Run: `nvim -u NONE -i NONE +'set title titlestring=spinner-demo titleold=Terminal | redraw!'`.
4. Observe the frozen tab icon.
5. Quit `neovim` with `:q`.

### Expected Behavior

If Reduced Motion is enabled, it might be better to either:

- Not show the spinner at all, or
- Replace it with a non-animated static icon that clearly indicates an active foreground process.

Right now, it looks like a spinner that failed to animate, which can be slightly confusing.

### OS Version

8.x (Circe)

### OS Architecture

amd64 (on most hardwares)

### Session Type

### Software Version

### Log Output

```shell

```

### Hardware Info

_No response_

jeremypw:
Fixes #990
Fixes #991

Does not show an animated spinner for any foreground processes.

The "WORKING" tab state and related code is retained pending an alternative way of visualizing the presence of a foreground process and/or the progress of an ongoing process.

### DIFF
--- a/data/terminal.metainfo.xml.in
+++ b/data/terminal.metainfo.xml.in
@@ -71,7 +71,6 @@
         </ul>
       </description>
       <issues>
-        <issue url="https://github.com/elementary/terminal/issues/317">Show symbol while terminal is working</issue>
         <issue url="https://github.com/elementary/terminal/issues/469">Ctrl + Click to open URLs</issue>
         <issue url="https://github.com/elementary/terminal/issues/510">Natural Copy/Paste. Ctrl+c clears the highlighted word. Ctrl+x doesn't.</issue>
         <issue url="https://github.com/elementary/terminal/issues/526">Changing font size when terminal has lots of output is slow.</issue>
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -168,7 +168,6 @@ public class Terminal.TerminalView : Granite.Bin {
 
         terminal_widget.notify["tab-state"].connect (() => {
             tab.icon = terminal_widget.tab_state.to_icon ();
-            tab.loading = terminal_widget.tab_state == WORKING;
         });
 
         // Set correct label now to avoid race when spawning shell
--- a/src/Widgets/TerminalWidget.vala
+++ b/src/Widgets/TerminalWidget.vala
@@ -13,6 +13,7 @@ namespace Terminal {
 
             public GLib.Icon? to_icon () {
                 switch (this) {
+                    //TODO Should we have an icon for the presence of a foreground process?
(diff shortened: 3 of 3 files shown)


jeremypw:
This should probably be merged before release of the Gtk4 port.

ryonakano (commented):
We should also remove the following issue tag from the metainfo file:

```xml
        <issue url="https://github.com/elementary/terminal/issues/317">Show symbol while terminal is working</issue>
```

ryonakano (approved):

PR #1009, opened 2026-05-13
Show tab numbers with shortcut

### LINKED ISSUE #655: Show the number of tabs.

jeremypw:
Fixes https://github.com/elementary/terminal/issues/655

We can't just show tab numbers immediatley when <Alt> is pressed as this key is also used for e.g. <Alt>BackSpace>. The solution adopted is to use a "long-press" on the <Alt> key. A previous solution was to use the unused <Alt>+0 combination to show the tab numbers immediately. This is retained for now but can be removed if considered superfluous.

The tab numbers are removed when a key is released, another key is pressed or focus lost.

It would be better to used an icon to show the numbers as that would be independent of the label and there is a suitable tabpage property, but there is no quick way to create an icon from a number (Gtk.NumerableIcon is discontinued in Gtk4) and we cannot use an overlay as TabPage is sealed.

So the tab number is prefixed to the tab label for now.

image

### DIFF
--- a/src/Widgets/TerminalView.vala
+++ b/src/Widgets/TerminalView.vala
@@ -282,6 +282,12 @@ public class Terminal.TerminalView : Granite.Bin {
 
     private bool key_pressed (uint keyval, uint keycode, Gdk.ModifierType modifiers) {
         switch (keyval) {
+            //NOTE Keys @1 to @9 in combination with Alt_L do not natively change tab - they
+            // cause `bash` to enter "digit argument" mode. Whether this behaviour is overridden is determined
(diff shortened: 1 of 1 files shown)


zeebok (approved):
Looks good! I use fish instead of bash so Alt+<number> switches tabs for me and it is delightful. Thanks for this!

PR #1010, opened 2026-05-21
Add RequestAttention DBus method for tabs needing input

carniz:
## Summary

- Adds a `TabState.ATTENTION` and a new DBus method `io.elementary.terminal.RequestAttention(terminal_id, message)` that lets a long-running program in a tab (notably AI agents) flag the tab when it needs user input.
- Reuses the existing process-completion plumbing: an icon (`dialog-question-symbolic`) is set on the tab, `Adw.TabPage.needs_attention` is toggled for the libadwaita accent-dot indicator, and a freedesktop notification is sent when the window/tab is not focused. State and notification clear automatically when the user focuses the tab.
- Documented in README under a new "Attention Requests" section.

## Why

When you have several long-running agent sessions in separate tabs, it's easy to miss when one of them is waiting for confirmation. This gives them a one-line, no-config way to ping you.

## Screencast
https://github.com/user-attachments/assets/347c360c-503e-4573-8e07-91dd6e620dca

## Test plan

- [ ] Build with `meson setup build && ninja -C build`
- [ ] Run `./build/src/io.elementary.terminal --gapplication-app-id=io.elementary.terminal.dev --new-window` and open two tabs
- [ ] In tab 1: `dbus-send --type=method_call --session --dest=io.elementary.terminal.dev /io/elementary/terminal/dev io.elementary.terminal.RequestAttention string:"$PANTHEON_TERMINAL_ID" string:"hello from agent"` — switch to tab 2 first; verify tab 1 shows the accent dot + question icon and a desktop notification appears
- [ ] Click tab 1: indicator clears, notification is withdrawn
- [ ] Repeat with the window unfocused (focus another app) — verify the desktop notification still appears

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -191,12 +191,22 @@ public class Terminal.Application : Gtk.Application {
             ulong focus_in_handler = 0;
 
             tab_change_handler = window.notify["current-terminal"].connect ((obj, pspec) => {
-                withdraw_attention_for_terminal ((MainWindow) obj, terminal, id, tab_change_handler, focus_in_handler);
+                withdraw_notification_for_terminal (
+                    (MainWindow) obj, terminal,
+                    "attention-requested-%s".printf (id),
+                    { TerminalWidget.TabState.ATTENTION },
+                    tab_change_handler, focus_in_handler
+                );
             });
 
             focus_in_handler = window.notify["is-active"].connect ((obj, pspec) => {
                 if (((MainWindow) obj).is_active) {
-                    withdraw_attention_for_terminal ((MainWindow) obj, terminal, id, tab_change_handler, focus_in_handler);
+                    withdraw_notification_for_terminal (
+                        (MainWindow) obj, terminal,
+                        "attention-requested-%s".printf (id),
+                        { TerminalWidget.TabState.ATTENTION },
+                        tab_change_handler, focus_in_handler
+                    );
                 }
             });
         });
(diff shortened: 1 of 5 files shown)


src/Application.vala
jeremypw: I guess we should check that the state == COMPLETED here?

src/Application.vala
```diff
@@ -228,6 +275,20 @@ public class Terminal.Application : Gtk.Application {
         window.disconnect (focus_in_handler);
     }
 
+    private void withdraw_attention_for_terminal (MainWindow window, TerminalWidget terminal, string id, ulong tab_change_handler, ulong focus_in_handler) {
```
jeremypw: Maybe the code here could be DRYed - it is very similar to the previous function. Could they be combined?

jeremypw (commented):
This works as expected and follows the pattern of existing code, thanks. I was just wondering whether some DRYing of the code is possible?

src/Application.vala
carniz: I'm going to address both your comments in the next commit, but yes, only wiping a specific state (or states) is probably a better idea.

src/Application.vala
```diff
@@ -154,6 +154,63 @@ public class Terminal.Application : Gtk.Application {
         var dbus = new DBus ();
         dbus_id = connection.register_object (object_path, dbus);
 
+        dbus.attention_requested.connect ((id, message) => {
+            TerminalWidget terminal = null;
+
```
jeremypw: Unnecessary whitespace

src/Application.vala
```diff
@@ -154,6 +154,63 @@ public class Terminal.Application : Gtk.Application {
...
+            if (window.is_active && terminal == window.current_terminal) {
+                return;
+            }
+
+            var notification_title = _("Terminal needs your attention");
+            var notification = new Notification (notification_title);
+            if (message.length > 0) {
+                notification.set_body (message);
+            }
```
jeremypw, line 184: Code style requires a blank line after an `if` clause

jeremypw (changes requested):
Works, but I noticed a slight difference between the behaviour of the `processed-finished` and `needs-attention` handlers.  The `needs-attention` handler sends a notification whenever the source tab is not focused even if the window is focused whereas the `processed-finished` handler only sends a notification if the window is not focused.  Is this intentional? There are other minor differences like no timeout but I assume these are not needed here.

I commented on another couple of code-style issues.

src/Application.vala
carniz: Hmm, I noticed that we run vala-lint in the ci workflow, but unless I'm missing something the linter (by default / without a vala-lint.conf) mostly generates a couple of line length warnings. Maybe it's time to add a configuration file for the linter that would catch things like this early on?

carniz:
Good point, I'll mimic the `processed-finished` handler so that it becomes more consistent: only send a notification if the window is unfocused (since the icon should be enough if the window already is focused). 

And yeah, I don't think I want any timeout for `needs-attention` since (at least) agents wait forever for your reply.

But I'll address the code-style issues as well.

jeremypw (approved):
Good to go now, thanks!

